PR #7, opened 2022-12-20
moving grand product computation to prover and adding corresponding test

ledwards2225:
-  Add function for computing the grand product polynomial z_perm into HonkProver.
- Add test that checks equivalence for z_perm computed in 2 ways: 1) by the actual function 'compute_grand_product_polynomial' that the prover will use, and 2) in a simple, un-optimized version local to the test that is much easier to read and verify (e.g. it uses naive inversion instead of batch inversion). This has explanatory benefit and also makes it easy to ensure that future optimizations do not change the result.


Rumata888 (approved):
Apart from a few typos, everything seems alright

PR #23, opened 2023-01-04
Gemini

adr1anh:
Resolves https://github.com/AztecProtocol/aztec2-internal/issues/1828

### DIFF
--- a/cpp/src/aztec/honk/pcs/commitment_key.hpp
+++ b/cpp/src/aztec/honk/pcs/commitment_key.hpp
@@ -1,0 +1,196 @@
+#pragma once
+
+/**
+ * @brief Provides interfaces for different 'CommitmentKey' classes.
+ *
+ */
+
+#include "polynomials/polynomial_arithmetic.hpp"
+#include <polynomials/polynomial.hpp>
+#include <srs/reference_string/file_reference_string.hpp>
+#include <ecc/curves/bn254/scalar_multiplication/scalar_multiplication.hpp>
+#include <ecc/curves/bn254/pairing.hpp>
+
+#include <string_view>
+#include <memory>
+
+namespace honk::pcs {
+
+namespace kzg {
+
+/**
+ * @brief CommitmentKey object over a pairing group 𝔾₁, using a structured reference string (SRS).
+ * The SRS is given as a list of 𝔾₁ points
+ *  { [xʲ]₁ }ⱼ where 'x' is unknown.
+ *
+ * @todo This class should take ownership of the SRS, and handle reading the file from disk.
+ */
+class CommitmentKey {
+    using Fr = typename barretenberg::g1::Fr;
+    // C is a "raw commitment" resulting to be fed to the transcript.
+    using C = typename barretenberg::g1::affine_element;
+    // Commitment represent's a homomorphically computed group element.
+    using Commitment = barretenberg::g1::element;
+
+    using Polynomial = barretenberg::Polynomial<Fr>;
+
+  public:
+    CommitmentKey() = delete;
+
+    /**
+     * @brief Construct a new Kate Commitment Key object from existing SRS
+     *
+     * @param n
+     * @param path
+     *
+     * @todo change path to string_view
+     */
+    CommitmentKey(const size_t n, std::string_view path)
+        : pippenger_runtime_state(n)
+        , srs(n, std::string(path))
+    {}
+
+    /**
+     * @brief Uses the ProverSRS to create a commitment to p(X)
+     *
+     * @param polynomial a univariate polynomial p(X) = ∑ᵢ aᵢ⋅Xⁱ ()
+     * @return Commitment computed as C = [p(x)] = ∑ᵢ aᵢ⋅[xⁱ]₁
+     */
+    C commit(std::span<const Fr> polynomial)
+    {
+        const size_t degree = polynomial.size();
+        ASSERT(degree <= srs.get_size());
+        return barretenberg::scalar_multiplication::pippenger_unsafe(
+            const_cast<Fr*>(polynomial.data()), srs.get_monomials(), degree, pippenger_runtime_state);
+    };
+
+  private:
+    barretenberg::scalar_multiplication::pippenger_runtime_state pippenger_runtime_state;
+    waffle::FileReferenceString srs;
+};
+
+class VerificationKey {
+    using Fr = typename barretenberg::g1::Fr;
+    using C = typename barretenberg::g1::affine_element;
+
+    using Commitment = barretenberg::g1::element;
+    using Polynomial = barretenberg::Polynomial<Fr>;
+
+  public:
+    VerificationKey() = delete;
+
+    /**
+     * @brief Construct a new Kate Commitment Key object from existing SRS
+     *
+     *
+     * @param verifier_srs verifier G2 point
+     */
+    VerificationKey(std::string_view path)
+        : verifier_srs(std::string(path))
+    {}
+
+    /**
+     * @brief verifies a pairing equation over 2 points using the verifier SRS
+     *
+     * @param p0 = P₀
+     * @param p1 = P₁
+     * @return e(P₀,[1]₁)e(P₁,[x]₂) ≡ [1]ₜ
+     */
+    bool pairing_check(const Commitment& p0, const Commitment& p1)
+    {
+        C pairing_points[2]{ p0, p1 };
+        // The final pairing check of step 12.
+        // TODO: try to template parametrise the pairing + fq12 output :/
+        barretenberg::fq12 result = barretenberg::pairing::reduced_ate_pairing_batch_precomputed(
+            pairing_points, verifier_srs.get_precomputed_g2_lines(), 2);
+
+        return (result == barretenberg::fq12::one());
+    }
+
+  private:
+    waffle::VerifierFileReferenceString verifier_srs;
+};
+
+struct Params {
+    using Fr = typename barretenberg::g1::Fr;
+    using C = typename barretenberg::g1::affine_element;
+
+    using Commitment = barretenberg::g1::element;
+    using Polynomial = barretenberg::Polynomial<Fr>;
+
+    using CK = CommitmentKey;
+    using VK = VerificationKey;
+};
+
+} // namespace kzg
+
+namespace fake {
+
+// Define a common trapdoor for both keys
+namespace {
+template <typename G> constexpr typename G::Fr trapdoor(5);
+}
+
+/**
+ * @brief Simulates a KZG CommitmentKey, but where we know the secret trapdoor
+ * which allows us to commit to polynomials using a single group multiplication.
+ *
+ * @tparam G the commitment group
+ */
+template <typename G> class CommitmentKey {
+    using Fr = typename G::Fr;
+    using C = typename G::affine_element;
+
+    using Commitment = typename G::element;
+    using Polynomial = barretenberg::Polynomial<Fr>;
+
+  public:
+    /**
+     * @brief efficiently create a KZG commitment to p(X) using the trapdoor 'secret'
+     * Uses only 1 group scalar multiplication, and 1 polynomial evaluation
+     *
+     *
+     * @param polynomial a univariate polynomial p(X)
+     * @return Commitment computed as C = p(secret)•[1]_1 .
+     */
+    C commit(std::span<const Fr> polynomial)
+    {
+        const Fr eval_secret = barretenberg::polynomial_arithmetic::evaluate(polynomial, trapdoor<G>);
+        return C::one() * eval_secret;
+    };
+};
+
+template <typename G> class VerificationKey {
+    using Fr = typename G::Fr;
+    using C = typename G::affine_element;
+
+    using Commitment = typename G::element;
+    using Polynomial = barretenberg::Polynomial<Fr>;
+
+  public:
+    /**
+     * @brief verifies a pairing equation over 2 points using the trapdoor
+     *
+     * @param p0 = P₀
+     * @param p1 = P₁
+     * @return P₀ - x⋅P₁ ≡ [1]
+     */
+    bool pairing_check(const Commitment& p0, const Commitment& p1)
+    {
+        Commitment result = p0 + p1 * trapdoor<G>;
+        return result.is_point_at_infinity();
+    }
+};
+
+template <typename G> struct Params {
+    using Fr = typename G::Fr;
+    using C = typename G::affine_element;
+
+    using Commitment = typename G::element;
+    using Polynomial = barretenberg::Polynomial<Fr>;
+
+    using CK = CommitmentKey<G>;
+    using VK = VerificationKey<G>;
+};
+} // namespace fake
+} // namespace honk::pcs
--- a/cpp/src/aztec/honk/pcs/gemini/gemini.hpp
+++ b/cpp/src/aztec/honk/pcs/gemini/gemini.hpp
@@ -1,0 +1,509 @@
+#pragma once
+
+#include "../claim.hpp"
+#include "polynomials/polynomial.hpp"
+
+#include <common/assert.hpp>
+#include <vector>
+
+/**
+ * @brief Protocol for opening several multi-linear polynomials at the same point.
+ *
+ *
+ * m = number of variables
+ * n = 2ᵐ
+ * u = (u₀,...,uₘ₋₁)
+ * f₀, …, fₖ₋₁ = multilinear polynomials,
+ * g₀, …, gₕ₋₁ = shifted multilinear polynomial,
+ *  Each gⱼ is the left-shift of some f↺ᵢ, and gⱼ points to the same memory location as fᵢ.
+ * v₀, …, vₖ₋₁, v↺₀, …, v↺ₕ₋₁ = multilinear evalutions s.t. fⱼ(u) = vⱼ, and gⱼ(u) = f↺ⱼ(u) = v↺ⱼ
+ *
+ * We use a challenge ρ to create a random linear combination of all fⱼ,
+ * and actually define A₀ = F + G↺, where
+ *   F  = ∑ⱼ ρʲ fⱼ
+ *   G  = ∑ⱼ ρᵏ⁺ʲ gⱼ,
+ *   G↺ = is the shift of G
+ * where fⱼ is normal, and gⱼ is shifted.
+ * The evaluations are also batched, and
+ *   v  = ∑ ρʲ⋅vⱼ + ∑ ρᵏ⁺ʲ⋅v↺ⱼ = F(u) + G↺(u)
+ *
+ * The prover then creates the folded polynomials A₀, ..., Aₘ₋₁,
+ * commits and opens them at different points, as univariates.
+ *
+ * We open A₀ as univariate at r and -r.
+ * Since A₀ = F + G↺, but the verifier only has commitments to the gⱼs,
+ * we need to partially evaluate A₀ at both evaluation points.
+ * As univariate, we have
+ *  A₀(X) = F(X) + G↺(X) = F(X) + G(X)/X
+ * So we define
+ *  - A₀₊(X) = F(X) + G(X)/r
+ *  - A₀₋(X) = F(X) − G(X)/r
+ * So that A₀₊(r) = A₀(r) and A₀₋(-r) = A₀(-r).
+ * The verifier is able to computed the simulated commitments to A₀₊(X) and A₀₋(X)
+ * since they are linear-combinations of the commitments [fⱼ] and [gⱼ].
+ */
+namespace honk::pcs::gemini {
+
+/**
+ * @brief A Gemini proof contains the m-1 commitments to the
+ * folded univariates, and corresponding evaluations
+ * at -r, -r², …, r^{2ᵐ⁻¹}.
+ *
+ * The evaluations allow the verifier to reconstruct the evaluation of A₀(r).
+ *
+ * @tparam Params CommitmentScheme parameters
+ */
+template <typename Params> struct Proof {
+    /** @brief Commitments to folded polynomials (size = m-2)
+     *
+     * [ C₁, …,  Cₘ₋₁], where Cₗ = commit(Aₗ(X)) of size 2ᵐ⁻ˡ
+     */
+    std::vector<typename Params::Commitment> commitments;
+
+    /**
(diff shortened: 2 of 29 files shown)


adr1anh:
# `ecc/group` and `ecc/field`

- Added convenience static methods `affine_element::one()`, `element::one()`, `element::zero()`.
- Added overload to `batch_invert` to accept `std::span`


# `honk /oracle` and `honk/transcript` 

Imported from `cryptonomicon`, though most of the code is commented out. We pass a `Oracle<Transcript>` to different protocols, as a way of modeling an interactive protocol. The interface for an oracle is 

```c++
void consume(args ...);
Fr generate_challenge();
```

For now, we use a dummy Oracle that generates `Fr(2), Fr(3), ...`. 

Note that the protocols implemented in this PR return the necessary proof data, so an adaptor does not strictly require the transcript to store the proof elements. 

# `polynomial`

Quite a lot of changes here... The starting point for these changes was to add a template argument for the field. This was initially done with a "find-replace". Unfortunately, it wasn't possible to actually instantiate `Polynomial` with another field since many methods call functions in `polynomial_arithmetic` where `fr` is hard-coded. 

Templatization of `polynomial_arithmetic` was done in a hack-y way, due to the fact that the implementation resides in a `.cpp` file. We are required to manually instantiate all the functions for both fields which leads to horrible boiler plate. The main use case for this is to enable testing with `grumpkin::g1`. 

Some parts of the interface were changed to make `Polynomial` conform to `std::ranges::contiguous_iterator`. This makes the class behave more like `std::vector`, allowing for range-based for loops, and conversion to `std::span`. 

Other convenience methods were added that are required for the multivariate PCS.

Also removed the `REPRESENTATION` parameter.

# `honk/pcs`

This PR includes the 3 protocols required to open multi-linear commitments. 

The stack is broadly 
- Gemini: reduce the opening of multilinear polynomials to opening several univariates at different points.
- Shplonk: Efficiently batch all opening from Gemini, resulting in the parties only having to open a single univariate in a single point.
- KZG: Open single univariate polynomial in a single point.

Each protocol accepts an input claim (e.g. commitments, evaluations and opening points), and a witness (polynomials for prover) or proof (for the verifier). The prover returns the new claim, witness, proof triple, while the verifier only returns the next claim. 

It's not strictly necessary for the prover to return the claim, but it helps ensure everything is correct (the prover in this case is just doing the verification as well). All claims can be verified directly without having to finalize the proof (we can check that a univariate opening is correct using the corresponding witness polynomial, by recomputing the commitment and evaluation from the witness and comparing against the values in the claim). This helps us test the modules individually. 

We introduce a new concept of `CommitmentKey` and `VerificationKey` (for a PCS), that are closely related to the concept of SRS. This construction makes it possible to define arbitrary commitment schemes. For now, we implement two:
- `KZGCommitmentKey` is constructed from an existing SRS path, and contains all `G1` points necessary to commit to polynomials. 
- `FakeCommitmentKey` is used for efficient debugging. It resembles a `KZGCommitmentKey` but where the toxic waste is a constant. It allows us to commit to polynomials using a single group multiplication.  It was defined both as a way of testing the "genericity" of the implementation, and being able to more easily debug the protocols.

We also define `VerificationKey` for the commitment scheme, used by the verifier in final step of the opening protocol. For now, it is constructed from the SRS file (or the constant toxic waste), and is only used to perform the pairing check. 
The final IPA verifier will use `IPACommitmentKey = IPAVerificationKey`. 

 A commitment scheme is parametrized by a `honk::pcs::SCHEME::Params<G>`, where we can optionally provide the elliptic curve group. It defines as aliases the types `CommitmentKey` and `VerificationKey` and other group-related types. This allows us to efficiently parametrize various components of the proof system depending on the commitment scheme used. 

## `gemini`

Somewhat optimized version of Gemini that only requires the opening of `log(n)-1` univariate polynomials, each in a single different point. 

## `shplonk` 

We provide two versions, namely `single_shplonk` and `multi_shplonk`. Both protocol are used to batch many univariate opening, but the former is limited to only one evaluation per protocol. It is also the version used by Gemini. The more general version was written first and could be useful in the future. 

## `kzg` 

We slightly differ from the classic KZG formulation and write the protocol again as a reduction. The result of the interaction between P and V is a pair of points P1, P2, such that e(P1,[1]_2)•e(P2, [x]_2) = 1. 
We call this pair a "bilinear accumulator", since it is possible to batch multiple accumulators together so the verification of all of them requires only a single pairing.

cpp/src/aztec/polynomials/polynomial.hpp
```diff
@@ -65,100 +74,205 @@ class polynomial {
...
+        size_ -= roots.size();
+    };
+    void factor_roots(const Fr& root)
+    {
+        polynomial_arithmetic::factor_roots(std::span{ *this }, root);
+        size_--;
+    };
+
+    /**
+     * Implements requirements of `std::ranges::contiguous_range` and `std::ranges::sized_range`
```
ledwards2225, line 224: Is there a reason why this is preferable to simply making the coefficient storage a std::vector?

cpp/src/aztec/honk/pcs/commitment_key.hpp
```diff
@@ -0,0 +1,194 @@
...
+/**
+ * @brief CommitmentKey object over a pairing group 𝔾₁, using a structured reference string (SRS).
+ * The SRS is given as a list of 𝔾₁ points
+ *  { [xʲ]₁ }ⱼ where 'x' is unknown.
+ *
+ * @todo This class should take ownership of the SRS, and handle reading the file from disk.
+ */
+class CommitmentKey {
+    using Fr = typename barretenberg::g1::Fr;
+    using C = typename barretenberg::g1::affine_element;
```
ledwards2225, line 31: This is a nit-pick but I think single-character naming should be avoided, even for a small scope type alias like this. If you feel strongly otherwise, I won't push back.

cpp/src/aztec/honk/pcs/gemini/gemini.hpp
```diff
@@ -0,0 +1,466 @@
...
+#include "polynomials/polynomial.hpp"
+
+#include <common/assert.hpp>
+#include <vector>
+
+/**
+ * @brief Protocol for opening several multi-linear polynomials at the same point.
+ *
+ *
+ * m = number of variables
```
ledwards2225, line 13: Just a note that elsewhere we've adopted 'd' for log2(n). Should decide on one and be consistent.

cpp/src/aztec/honk/pcs/gemini/gemini.hpp
```diff
@@ -0,0 +1,466 @@
...
+ * n = 2ᵐ
+ * u = (u₀,...,uₘ₋₁)
+ * f₀, …, fₖ₋₁ = multilinear polynomials,
+ *  where we denote shifted polynomials as fⱼ = fⱼ↺.
```
ledwards2225: We talked about this a little bit but I think this fⱼ = fⱼ↺ notation has the potential to cause confusion. The ↺ is just used as a tag to say "this is a polynomial whose shift is of interest to us" but I think fⱼ↺ is easily misinterpreted as the shift of fⱼ, i.e. fⱼ↺ = fⱼ/X or something like this. I'm not sure I have a suggestion, just flagging this for now as a point of potential confusion.

cpp/src/aztec/honk/pcs/gemini/gemini.hpp
```diff
@@ -0,0 +1,466 @@
...
+ * v₀, …, vₖ₋₁ = multilinear evalutions s.t. fⱼ(u) = vⱼ
+ * Note that the set of shifted polynomials must be a subset of the non-shifted ones.
+ *
+ * We use a challenge ρ to create a random linear combination of all fⱼ,
+ * and actually define A₀ = B + B↺, where
+ *   B  = ∑ⱼ ρʲ fⱼ
+ *   B↺ = ∑ₖ ρʲ fⱼ↺
+ * where fⱼ is normal, and fⱼ is shifted.
```
ledwards2225: Looks like maybe you meant to say "fⱼ is normal, and fⱼ↺ is shifted"? But again, it's not accurate to say fⱼ↺ is _shifted_ right? The polynomials batched into B↺ are the unshifted polynomials whose shifts are required.

cpp/src/aztec/honk/pcs/gemini/gemini.hpp
```diff
@@ -0,0 +1,466 @@
...
+ * The evaluations are also batched, and
+ *   v  = ∑ ρʲ⋅vⱼ
+ *
+ * The prover then creates the folded polynomials A₀, ..., Aₘ₋₁,
+ * commits and opens them at different points, as univariates.
+ *
+ * We open A₀ as univariate at r and -r,
+ * When one of our polynomials fⱼ is shifted, we need to open
+ * fⱼ↺(X) = fⱼ(X)/X.
```
ledwards2225: haha, oh boy

cpp/src/aztec/honk/pcs/gemini/gemini.hpp
```diff
@@ -0,0 +1,466 @@
...
+        witness_polynomials.reserve(num_variables + 1);
+
+        // create batched polynomial which is a linear combination
+        // of polynomials fⱼ(X) with powers ρʲ
+        // A₀(X) = ∑ⱼ ρʲ fⱼ(X)
+        //
+        // In what follows, we use indices j, k for non-shifted and shifted polynomials respectively.
+        //
+        // We separate A₀(X) into two polynomials B(X), B↺(X)
+        // such that A₀(X) = B(X) + B↺(X)/X.
```
ledwards2225: Getting back to this point about notation: to me it would be clearer to say something like A₀(X) = B(X) + B↺(X), where B↺(X) = ∑_{ⱼ↺} ρʲ fⱼ(X)/X = ∑{ⱼ↺} ρʲ fⱼ↺(X) (where ∑{ⱼ↺} is the sum over the set of polynomials whose shift is required). In other words, ↺ always indicates an actual shift: fⱼ↺(X) = fⱼ(X)/X.

cpp/src/aztec/honk/pcs/gemini/gemini.hpp
```diff
@@ -0,0 +1,466 @@
...
+        A_0_pos.add(batched_polynomial_shifted);
+
+        std::swap(tmp, batched_polynomial_shifted);
+        // After the swap, we have
+        // tmp = B↺(X)/r
+        // A_0_neg = B(X) (since &batched_polynomial_shifted == &A_0_neg)
+        Polynomial& A_0_neg = witness_polynomials[1];
+
+        // A₀₋(X) = B(X) - B↺(X)/r, s.t. A₀₋(-r) = A₀(-r)
+        A_0_neg.sub(tmp);
```
ledwards2225: I'd like to see the method 'sub' named a bit less ambiguously, e.g. 'subtract' or 'minus'. In fact, any reason why you don't prefer using overloaded operators instead?

cpp/src/aztec/honk/pcs/gemini/gemini.hpp
```diff
@@ -0,0 +1,466 @@
...
+        // evaluate all new polynomials Aₗ at -rₗ
+        std::vector<Fr> evals;
+        evals.reserve(num_variables);
+        for (size_t l = 0; l < num_variables; ++l) {
+            const Polynomial& A_l = witness_polynomials[l + 1];
+            const Fr r_l_neg = -r_squares[l];
+            evals.emplace_back(A_l.evaluate(r_l_neg));
+        }
+
+        challenge_generator.consume(evals);
```
ledwards2225, line 306: I definitely like the idea of this 'consume' method. It's nice to abstract away all of the underlying details.

cpp/src/aztec/honk/pcs/gemini/gemini.hpp
```diff
@@ -0,0 +1,466 @@
...
+     * @return OutputClaim<Params>
+     */
+    static OutputClaim<Params> compute_output_claim_from_proof(std::span<const MLEOpeningClaim<Params>> claims,
+                                                               std::span<const Fr> mle_vars,
+                                                               std::span<const Fr> rhos,
+                                                               std::span<const Fr> r_squares,
+                                                               const Proof<Params>& proof)
+    {
+        const size_t m = proof.commitments.size() + 1;
+        const Fr r = r_squares[0];
```
ledwards2225, line 384: The single character name is not ideal here, especially since it's used in a different way within a sub-scope in between here and where this variable is actually used.

cpp/src/aztec/honk/pcs/gemini/gemini.hpp
```diff
@@ -0,0 +1,466 @@
...
+            mle_eval += claims[j].evaluation * rhos[j];
+        }
+
+        Fr eval_pos = mle_eval;
+        for (size_t l = m; l != 0; --l) {
+            const Fr eval_neg = evals[l - 1];
+            const Fr u = mle_vars[l - 1];
+            const Fr r = r_squares[l - 1];
+
+            eval_pos = ((r * eval_pos * 2) - eval_neg * (r * (Fr(1) - u) - u)) / (r * (Fr(1) - u) + u);
```
ledwards2225, line 417: A corresponding math comment would be helpful for this line

cpp/src/aztec/honk/pcs/gemini/gemini.test.cpp
```diff
@@ -0,0 +1,156 @@
...
+    auto commitment = this->commit(poly);
+    auto eval = poly.evaluate_mle(u);
+
+    // create opening claim
+    auto claims = { MLEOpeningClaim{ commitment, eval, false } };
+
+    this->consume(u);
+
+    auto [prover_claim, witness, proof] =
+        Gemini::reduce_prove(this->ck(), u, claims, { &poly }, this->prover_challenges);
```
ledwards2225: I had been wondering if a better name for `reduce_prove` might be something like `construct_batch_opening_claim`. These tests makes me think so :)

cpp/src/aztec/polynomials/polynomial.cpp
```diff
@@ -433,21 +444,159 @@ void polynomial::coset_ifft(const evaluation_domain& domain)
...
+template <typename Fr> void Polynomial<Fr>::scale(Fr scaling_facor)
+{
+    ASSERT(!mapped_);
+
+    for (size_t i = 0; i < size_; ++i) {
+        coefficients_[i] *= scaling_facor;
+    }
+}
+
+template <typename Fr> Fr Polynomial<Fr>::evaluate_mle(std::span<const Fr> evaluation_points, bool shift) const
```
ledwards2225, line 543: There's no difference between a normal evaluation and a shifted evaluation? What am I missing here

cpp/src/aztec/honk/pcs/kzg/kzg.hpp
```diff
@@ -0,0 +1,94 @@
...
+
+#include <utility>
+
+namespace honk::pcs::kzg {
+/**
+ * @brief A transformed polynomial commitment opening claim of the form (P₀, P₁) ∈ 𝔾₁
+ * which should satisfy e(P₀, [1]₂) ⋅ e(P₁, [x]₂)=1.
+ *
+ * @tparam Params CommitmentScheme parameters, where the verification key VK has a
+ * `paring_check` function.
```
ledwards2225: pairing_check*

cpp/src/aztec/honk/pcs/kzg/kzg.hpp
```diff
@@ -0,0 +1,94 @@
...
+
+    /**
+     * @brief Compute an accumulator for a single polynomial commitment opening claim
+     *
+     * @param ck CommitmentKey
+     * @param claim OpeningClaim = (C,r,v)
+     * @param polynomial the witness polynomial for C
+     * @return Output{Accumulator, Proof}
+     */
+    static Output reduce_prove(CK* ck, const OpeningClaim<Params>& claim, const Polynomial& polynomial)
```
ledwards2225, line 71: As with the other `reduce_prove` functions, would a better name be something like `construct_opening_proof`?

cpp/src/aztec/honk/pcs/kzg/kzg.hpp
```diff
@@ -0,0 +1,94 @@
...
+
+    /**
+     * @brief Computes the accumulator for a single polynomial commitment opening claim
+     * This reduction is non-interactive and always succeeds.
+     *
+     * @param claim OpeningClaim (C,r,v)
+     * @param proof a commitment to Q(X) = ( P(X) - v )/( X - r)
+     * @return Accumulator
+     */
+    static Accumulator reduce_verify(const OpeningClaim<Params>& claim, const Proof& proof)
```
ledwards2225, line 89: I can see how this name makes some sense but I think it's a bit misleading since it's not _verifying_ anything. Your brief suggests something like `accumulate_single_opening_claim`...

cpp/src/aztec/honk/pcs/gemini/gemini.hpp
```diff
@@ -0,0 +1,466 @@
...
+         * powers of ρ
+         * We use indices j, k for non-shifted and shifted polynomials respectively.
+         */
+
+        // 2 simulated polynomials and (m-1) polynomials from this round
+        Fr r_inv = r.invert();
+        // B↺(X) *= r⁻¹
+        batched_polynomial_shifted.scale(r_inv);
+
+        // To avoid an extra allocation, we reuse the following polynomials
```
ledwards2225, line 272: Took me a few re-reads of this function but I'm convinced that this is doing what it purports to do. There's no question that it could be written more clearly at the expense of some inefficiency but given the optimizations I think you've done a nice job with the comments.

cpp/src/aztec/honk/pcs/gemini/gemini.hpp
```diff
@@ -0,0 +1,466 @@
...
+            if (claims[j].shift) {
+                // B↺(X) += ρʲ fⱼ↺(X)
+                batched_polynomial_shifted.add_scaled(*mle_witness_polynomials[j], rhos[j]);
+            } else {
+                // B(X) += ρʲ fⱼ(X)
+                batched_polynomial.add_scaled(*mle_witness_polynomials[j], rhos[j]);
+            }
+        }
+
+        // A₀(X) = B(X) + B↺(X)/X.
```
ledwards2225: Using your existing notation, here `A_0` is actually set to A₀(X) = B(X) + B↺(X), not A₀(X) = B(X) + B↺(X)/X, right? This matches your documentation in the 'minimal sketch of honk'. Does this comment need to be corrected?

ledwards2225 (changes requested):
Overall looks very good to me. The documentation (comments/testing) is quite thorough in most cases. Most of my comments are suggestions for minor improvements to comments/notation/naming.

cpp/src/aztec/polynomials/polynomial.hpp
adr1anh: This is more powerful, we can now  directly iterate over the coefficients, and convert it to `std::span`
```c++
Polynomial<Fr> p;
for (const auto& coeff : p) {
  // do something with 'coeff'
}
std::span sp{p};
```

cpp/src/aztec/honk/pcs/commitment_key.hpp
adr1anh: Agreed, `C` isn't very descriptive in this case. The idea is that there should be a difference between "raw commitments" (`C`) and those that are derived homorphically (`Commitment`). It's an artifact of the previous work where `Commitment` represented a linear-combination of `C`s. In any case, these types are private and are not exposed. Adding a comment to adress this.

cpp/src/aztec/honk/pcs/gemini/gemini.hpp
adr1anh: I would've prefered to use `d`, but I think the reason I used `m` is because we have unicode ₘ, but not "_d". Not sure if this is a strong enough reason though.

cpp/src/aztec/honk/pcs/gemini/gemini.hpp
adr1anh: You're right, I added the operators to the Polynomial class and use those instead.

cpp/src/aztec/honk/pcs/gemini/gemini.hpp
adr1anh: Thanks for taking the time!

cpp/src/aztec/honk/pcs/gemini/gemini.hpp
adr1anh: Addressed more generally, see comment later on the PR

cpp/src/aztec/honk/pcs/gemini/gemini.hpp
adr1anh: Renamed to num_variables, to be consistent with the rest

cpp/src/aztec/honk/pcs/gemini/gemini.test.cpp
adr1anh: I'm more inclined to keep the term `reduce`, since we're reducing both the claims and the witness, and constructing a proof. 
However, I agree that `reduce_prove` isn't super descriptive. 
I'm thinking something like `reduce_to_batch_univariate_claim` ?

cpp/src/aztec/honk/pcs/kzg/kzg.hpp
adr1anh: We're constructing both an opening proof and an accumulator, so maybe something like `reduce_to_pairing` ?

cpp/src/aztec/polynomials/polynomial.cpp
adr1anh: If `shift==true` then we increase `prev` by 1, so that we skip the first coefficient.

adr1anh:
The update to `polynomial` replaces `add,sub,scale` by operators. The accept `std::span<const Fr>` so that we can compute 
`p += q.shifted()`, where `.shifted()` returns a span of the coefficients of the shifted polynomial. 

Also changes some assertions to be precise.

adr1anh:
The Gemini implementation was reworked to make the distinction between the non-shifted and shifted polynomials more clear. We now explicitly separate the sets of polynomials, and label them $\{f_j(X)\}$  and $\{g_j(X)\}$ respectively, and their batched counter parts are $F(X)$ and $G(X)$.

PR #30, opened 2023-01-04
Move verification_key and polynomial_manifest to proof_system dir

ledwards2225:
Moving files related to verification key and polynomial manifest from plonk/proof_system to shared proof_system directory since they are used by both plonk and honk


codygunton (approved):
Straightfoward, looks good.

PR #31, opened 2023-01-05
IPA PoC

arijitdutta67:
This PR implements a proof of concept version of the inner product argument (IPA) polynomial commitment scheme as in https://eprint.iacr.org/2019/1021.pdf. The spec for the implementation is in https://hackmd.io/q-A8y6aITWyWJrvsGGMWNA?view.

### DIFF
--- a/cpp/src/aztec/honk/commitment_scheme/ipa/ipa.hpp
+++ b/cpp/src/aztec/honk/commitment_scheme/ipa/ipa.hpp
@@ -35,8 +35,8 @@ template <typename Fr, typename Fq, typename G1> class InnerProductArgument {
         Fr a_zero;
     };
     // To contain the public inputs for IPA proof
-    // For now we are including the aux_generator in public input. It will be computed by the prover and the verifier
-    // separately when the challengeGenerator is defined.
+    // For now we are including the aux_generator and round_challenges in public input. They will be computed by the
+    // prover and the verifier by Fiat-Shamir when the challengeGenerator is defined.
     struct IpaPubInput {
         element commitment;
         Fr challenge_point;
@@ -79,14 +79,16 @@ template <typename Fr, typename Fq, typename G1> class InnerProductArgument {
     {
         IpaProof proof;
         auto& challenge_point = ipa_pub_input.challenge_point;
-        ASSERT(challenge_point != 0);
+        ASSERT(challenge_point != 0 && "The challenge point should not be zero");
         const size_t poly_degree = ipa_pub_input.poly_degree;
-        ASSERT(poly_degree != 0);
+        // To check poly_degree is greater than zero and a power of two
+        ASSERT((poly_degree > 0) && (!(poly_degree & (poly_degree - 1))) &&
+               "The poly_degree should be positive and a power of two");
         auto& aux_generator = ipa_pub_input.aux_generator;
         auto a_vec = polynomial;
-        std::vector<affine_element> G_vec(poly_degree);
+        std::vector<affine_element> G_vec_local(poly_degree);
         for (size_t i = 0; i < poly_degree; i++) {
-            G_vec[i] = G_vector[i];
+            G_vec_local[i] = G_vector[i];
         }
         // Construct b vector
         std::vector<Fr> b_vec(poly_degree);
@@ -113,12 +115,12 @@ template <typename Fr, typename Fq, typename G1> class InnerProductArgument {
             // L_i = < a_vec_lo, G_vec_hi > + inner_prod_L * aux_generator
             auto pippenger_runtime_state = barretenberg::scalar_multiplication::pippenger_runtime_state(round_size);
             element partial_L = barretenberg::scalar_multiplication::pippenger_without_endomorphism_basis_points(
-                &a_vec[0], &G_vec[round_size], round_size, pippenger_runtime_state);
+                &a_vec[0], &G_vec_local[round_size], round_size, pippenger_runtime_state);
             partial_L += aux_generator * inner_prod_L;
 
             // R_i = < a_vec_hi, G_vec_lo > + inner_prod_R * aux_generator
             element partial_R = barretenberg::scalar_multiplication::pippenger_without_endomorphism_basis_points(
-                &a_vec[round_size], &G_vec[0], round_size, pippenger_runtime_state);
+                &a_vec[round_size], &G_vec_local[0], round_size, pippenger_runtime_state);
             partial_R += aux_generator * inner_prod_R;
 
             std::vector<element> round_commitments(2);
@@ -143,12 +145,12 @@ template <typename Fr, typename Fq, typename G1> class InnerProductArgument {
                 b_vec[j] *= round_challenge_inv;
                 b_vec[j] += round_challenge * b_vec[round_size + j];
 
-                // element has (x,y,z) coord, affine_element has (x,y) coord, G_vec is a vector of affine_element,
+                // element has (x,y,z) coord, affine_element has (x,y) coord, G_vec_local is a vector of affine_element,
 
-                auto G_lo = element(G_vec[j]) * round_challenge_inv;
-                auto G_hi = element(G_vec[round_size + j]) * round_challenge;
+                auto G_lo = element(G_vec_local[j]) * round_challenge_inv;
+                auto G_hi = element(G_vec_local[round_size + j]) * round_challenge;
                 auto temp = G_lo + G_hi;
-                G_vec[j] = temp.normalize();
+                G_vec_local[j] = temp.normalize();
             }
         }
         proof.L_vec = std::vector<affine_element>(log_poly_degree);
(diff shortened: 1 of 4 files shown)


cpp/src/aztec/honk/commitment_scheme/ipa/ipa.hpp
```diff
@@ -0,0 +1,266 @@
...
+// Suggested by Zac: Future optimisations
+// 1: write a program that generates a large set of generator points (2^23?) and writes to a file on disk
+// 2: create a SRS class for IPA similar to existing SRS class, that loads these points from disk
+//    and stores in struct *and* applies the pippenger point table endomorphism transforation
+// 3: when constructing a InnerProductArgument class, pass std::shared_ptr<SRS> as input param and store as member
+// variable
+// 4: (SRS class should contain a `pippenger_runtime_state` object so it does not need to be repeatedly
+// generated)
+
+namespace waffle {
```
zac-williamson: We can remove the `waffle` namespace. It's a hangover from when barretenberg had a module named `waffle` (years ago). We should remove it from the codebase entirely at some point

cpp/src/aztec/honk/commitment_scheme/ipa/ipa.hpp
```diff
@@ -0,0 +1,266 @@
...
+    static IpaProof ipa_prove(const IpaPubInput& ipa_pub_input,
+                              const std::vector<Fr>& polynomial,
+                              const std::vector<affine_element>& G_vector)
+    {
+        IpaProof proof;
+        auto& challenge_point = ipa_pub_input.challenge_point;
+        ASSERT(challenge_point != 0 && "The challenge point should not be zero");
+        const size_t poly_degree = ipa_pub_input.poly_degree;
+        // To check poly_degree is greater than zero and a power of two
+        ASSERT((poly_degree > 0) && (!(poly_degree & (poly_degree - 1))) &&
```
zac-williamson, line 84: The commitment scheme should be able to accommodate commitments where the degree is not a power of 2

i.e. the commitment scheme increases the size of the `polynomial` vector until it is a power of 2 and fills the new entries with zeroes.

This work can be done in a future branch through and isn't a blocker to this PR.

cpp/src/aztec/honk/commitment_scheme/ipa/ipa.hpp
```diff
@@ -0,0 +1,266 @@
...
+            round_size >>= 1;
+            // Compute inner_prod_L := < a_vec_lo, b_vec_hi > and inner_prod_R := < a_vec_hi, b_vec_lo >
+            Fr inner_prod_L = Fr::zero();
+            Fr inner_prod_R = Fr::zero();
+            for (size_t j = 0; j < round_size; j++) {
+                inner_prod_L += a_vec[j] * b_vec[round_size + j];
+                inner_prod_R += a_vec[round_size + j] * b_vec[j];
+            }
+            // L_i = < a_vec_lo, G_vec_hi > + inner_prod_L * aux_generator
+            auto pippenger_runtime_state = barretenberg::scalar_multiplication::pippenger_runtime_state(round_size);
```
zac-williamson, line 122: For future work we need to remove the creation of multiple `pippenger_runtime_state` objects.

The commitment scheme should use a single `pippenger_runtime_state` object that is a member of the `SRS`

cpp/src/aztec/honk/commitment_scheme/ipa/ipa.hpp
```diff
@@ -0,0 +1,266 @@
...
+        ASSERT(challenge_point != 0 && "The challenge point should not be zero");
+        const size_t poly_degree = ipa_pub_input.poly_degree;
+        // To check poly_degree is greater than zero and a power of two
+        ASSERT((poly_degree > 0) && (!(poly_degree & (poly_degree - 1))) &&
+               "The poly_degree should be positive and a power of two");
+        auto& aux_generator = ipa_pub_input.aux_generator;
+        auto a_vec = polynomial;
+        std::vector<affine_element> G_vec_local(poly_degree);
+        for (size_t i = 0; i < poly_degree; i++) {
+            G_vec_local[i] = G_vector[i];
```
zac-williamson, line 93: It's quite inefficient to do a deep copy to `G_vec_local`


If I understand the code correctly, it would be more efficient to directly use `G_vector` for the input points when `i = 0` and write the output points to `G_vec_local`. Then use `G_vec_local` for rounds where `i>0`

cpp/src/aztec/honk/commitment_scheme/ipa/ipa.hpp
```diff
@@ -0,0 +1,266 @@
...
+        size_t round_size = poly_degree;
+
+        for (size_t i = 0; i < log_poly_degree; i++) {
+            round_size >>= 1;
+            // Compute inner_prod_L := < a_vec_lo, b_vec_hi > and inner_prod_R := < a_vec_hi, b_vec_lo >
+            Fr inner_prod_L = Fr::zero();
+            Fr inner_prod_R = Fr::zero();
+            for (size_t j = 0; j < round_size; j++) {
+                inner_prod_L += a_vec[j] * b_vec[round_size + j];
+                inner_prod_R += a_vec[round_size + j] * b_vec[j];
```
zac-williamson, line 118: For round `i=0` `b_vec` can be derived in-place.

This means that the size of `b_vec` can be 50% of the current size (i.e. we only write values to `b_vec` at the end of round 0 

I don't think this is a priority but should be added as a TODO comment

cpp/src/aztec/honk/commitment_scheme/ipa/ipa.hpp
```diff
@@ -0,0 +1,266 @@
...
+            element partial_R = barretenberg::scalar_multiplication::pippenger_without_endomorphism_basis_points(
+                &a_vec[round_size], &G_vec_local[0], round_size, pippenger_runtime_state);
+            partial_R += aux_generator * inner_prod_R;
+
+            std::vector<element> round_commitments(2);
+            round_commitments[0] = partial_L;
+            round_commitments[1] = partial_R;
+            element::batch_normalize(&round_commitments[0], 2);
+
+            L_elements[i] = affine_element({ round_commitments[0].x, round_commitments[0].y });
```
zac-williamson: you can directly construct `affine_element` objects from `element`.


e.g. `L_elements[I] = affine_element(round_commitments[0])`

zac-williamson: The code would be cleaner (with a negligible performance effect) if we wrote:
```
L_elements[i] = affine_element(partial_L);
R_elements[i] = affine_element(partial_R);
```

cpp/src/aztec/honk/commitment_scheme/ipa/ipa.hpp
```diff
@@ -0,0 +1,266 @@
...
+            // G_vec_next = G_vec_lo * round_challenge_inv + G_vec_hi * round_challenge
+            for (size_t j = 0; j < round_size; j++) {
+                a_vec[j] *= round_challenge;
+                a_vec[j] += round_challenge_inv * a_vec[round_size + j];
+                b_vec[j] *= round_challenge_inv;
+                b_vec[j] += round_challenge * b_vec[round_size + j];
+
+                // element has (x,y,z) coord, affine_element has (x,y) coord, G_vec_local is a vector of affine_element,
+
+                auto G_lo = element(G_vec_local[j]) * round_challenge_inv;
```
zac-williamson, line 160: We can improve performance here by using `element::batch_mul_with_endomorphism`


This method takes a vector of input points `points` and a scalar `x` and outputs a vector containing `points[i]*x`

It's 30% faster than a basic mul operation due to performing group additions in 2D affine coordinates instead of 3D projective coordinates (affine point additions are usually more expensive than projective additions due to the need to compute a modular inverse. However we get around this by computing a single batch inverse. This only works if you are adding a lot of independent point pairs so you can amortise the cost of the single batch inversion across multiple points).


Not a priority but please add this as a TODO

cpp/src/aztec/honk/commitment_scheme/ipa/ipa.hpp
zac-williamson: Please add this as a future TODO

zac-williamson (changes requested):
I've left some comments describing future improvements.

The only action items before we merge are to remove the `waffle` namespace and to tidy up how L_elements and R_elements are populated (directly construct from partial_L and partial_R)

Once this is done I'm happy for the branch to be merged.

However I wouldn't consider this project truly complete until the following points are achieved:

1. `commit`, `ipa_prove` and `ipa_verify` take in an SRS object instead of a vector of generators (and have this SRS class load the required points from disk like our KZG SRS)
2. the SRS object contains a single `pippenger_runtime_state` object that is used to perform all pippenger computations (constructing `pippenger_runtime_state` is expensive)
3. the Fiat-Shamir heuristic is used to generate the round challenges as part of the `ipa_prove` and `ipa_verify` algorithms. The `InnerProductArgument` class should have a `Hash` template parameter that defines which hash algorithm is used.
4. methods are added that batch prove and batch verify multiple polynomial commitments

The above 4 points are required for the IPA commitment scheme to be used in a protocol like Honk or Plonk.

cpp/src/aztec/honk/commitment_scheme/ipa/ipa.hpp
codygunton: :100: though I've had fun  trying to fill in the analogy 
"Plonk is to waffle as honk is to _____".
I ended up chuckling at the idea of having a namespace `road_rage`.

cpp/src/aztec/honk/commitment_scheme/ipa/ipa.hpp
arijitdutta67: Thanks for this. Works.

cpp/src/aztec/honk/commitment_scheme/ipa/ipa.hpp
arijitdutta67: Added a Todo for this. Will fix this in the upcoming PR.

PR #33, opened 2023-01-05
[REPLACED] Typed Plonk tests for Linearised and Non-Linearised Prover

suyash67:
# REPLACED BY PR https://github.com/AztecProtocol/barretenberg/pull/92


suyash67:
Closing because we got rid of linearised prover anyway.

PR #34, opened 2023-01-05
Cg/poc manifest

codygunton:
Write manifest and partially-integrate StandardHonk. Summary of major changes:
- Changed `Manifest` and `RoundManifest` constructors to take a vector instead of an initializer list. This was necessary to allow building  `RoundManifest`'s of varying length (there is no way to programatically specify an initializer list, as far as I can tell).
- Reorganized prover proof construction functions to have informative names and match more cleanly with the analogous functions for plonk (future todo: rename these).
- Mocked transcript interactions of the prover.
 - Added test `StandardHonkComposer.BaseCase` to ensure that manifest and new prover functions compile and run without segfaults. Test does not have any `EXPECT`s.
- Template `create_unrolled_prover` by flavor.  Composer no longer has any of its own functions, it is just a container coordinating several classes, so we needed a place to put the `create_manifest` function. I put this in the flavor. Perhaps the whole `ComposerHelper` class should be templated with Flavor and `CircuitConstructor` created from this.
- Move some objects of Honk from `waffle` to `honk` namespace for clarity. Eventually there should be no `waffle`s anywhere, and this should first happen inside Honk files.
- Added `onk` namespace as a place for proving system things shared between `honk` and `plonk`.


cpp/src/aztec/honk/composer/composer_helper/composer_helper.cpp
```diff
@@ -257,15 +262,15 @@ std::shared_ptr<verification_key> ComposerHelper<CircuitConstructor>::compute_ve
  * @return The verifier.
  * */
 template <typename CircuitConstructor>
-Verifier ComposerHelper<CircuitConstructor>::create_verifier(CircuitConstructor& circuit_constructor)
+waffle::Verifier ComposerHelper<CircuitConstructor>::create_verifier(CircuitConstructor& circuit_constructor)
```
ledwards2225, line 265: This needs to be a honk::Verifier. Just flagging for now, maybe this can wait

cpp/src/aztec/honk/composer/composer_helper/composer_helper.cpp
```diff
@@ -293,14 +300,18 @@ UnrolledVerifier ComposerHelper<CircuitConstructor>::create_unrolled_verifier(Ci
 }
 
 template <typename CircuitConstructor>
-UnrolledProver ComposerHelper<CircuitConstructor>::create_unrolled_prover(CircuitConstructor& circuit_constructor)
+template <typename Flavor>
+// TODO(Cody): this file should be generic with regard to flavor/arithmetization/whatever.
+StandardUnrolledProver ComposerHelper<CircuitConstructor>::create_unrolled_prover(
```
ledwards2225, line 305: Either rolled/unrolled has to go. My vote would be to get rid of the unrolled keyword but I can see why you'd want to keep it for consistency with PlonK.

cpp/src/aztec/honk/composer/composer_helper/composer_helper.hpp
```diff
@@ -1,68 +1,75 @@
 #include <srs/reference_string/file_reference_string.hpp>
 #include <proof_system/proving_key/proving_key.hpp>
-#include <plonk/proof_system/verification_key/verification_key.hpp>
+#include <honk/proof_system/prover.hpp>
+// keeping this include for now to use the proving key
```
ledwards2225: Is this still necessary? Doesn't seem like it should be. Something still isn't hooked up right?

cpp/src/aztec/honk/proof_system/program_settings.hpp
```diff
@@ -16,7 +16,7 @@ class standard_verifier_settings : public waffle::standard_settings {
     static constexpr size_t num_challenge_bytes = 32;
     static constexpr transcript::HashType hash_type = transcript::HashType::Keccak256;
 
-    static constexpr size_t num_polys = StandardArithmetization::NUM_POLYNOMIALS;
+    static constexpr size_t num_polys = onk::StandardArithmetization::NUM_POLYNOMIALS;
```
ledwards2225: Haha 'onk'? I mean, I see what you did there but..

cpp/src/aztec/honk/proof_system/prover.cpp
```diff
@@ -255,47 +264,109 @@ template <typename settings> void Prover<settings>::execute_third_round()
...
+    transcript.add_element("sigma_3", barretenberg::fr(105).to_buffer());
+    transcript.add_element("q_1", barretenberg::fr(106).to_buffer());
+    transcript.add_element("q_2", barretenberg::fr(107).to_buffer());
+    transcript.add_element("q_3", barretenberg::fr(108).to_buffer());
+    transcript.add_element("q_m", barretenberg::fr(109).to_buffer());
+    transcript.add_element("q_c", barretenberg::fr(110).to_buffer());
+    transcript.add_element("z_perm", barretenberg::fr(111).to_buffer());
+    transcript.add_element("z_perm_omega", barretenberg::fr(112).to_buffer());
+
+    transcript.apply_fiat_shamir("rho");
```
ledwards2225: This will be done in Gemini so I think it belongs in `execute_univariatization_round()`

cpp/src/aztec/honk/proof_system/prover.cpp
```diff
@@ -255,47 +264,109 @@ template <typename settings> void Prover<settings>::execute_third_round()
...
+}
+
+/**
+ * For Plonk: the polynomials are univariate, so this is a no-op.
+ * For Honk:
+ * - Get rho challenge
+ * - Compute Fold polynomials and commitments.
+ *
+ * */
+template <typename settings> void Prover<settings>::execute_univariatization_round()
```
ledwards2225, line 308: Ok so the two functions you have here,  `execute_univariatization_round` and `execute_pcs_evaluation_round` are handled by the single Gemini function `reduce_prove`. So these two functions can be combined into one, say `execute_gemini` (or whatever you want). This new function will mock the FOLD commitments and the uni evals as well as compute challenges rho and r.

cpp/src/aztec/honk/proof_system/prover.cpp
```diff
@@ -255,47 +264,109 @@ template <typename settings> void Prover<settings>::execute_third_round()
...
+
+/**
+ * For Plonk: Batching is combined with generation of opening proof polynomial commitments.
  *
+ * For Honk:
+ * - Do Fiat-Shamir to get "nu" challenge.
+ * - Compute Shplonk batched quotient commitment [Q]_1.
  * */
-template <typename settings> void Prover<settings>::execute_fifth_round()
+template <typename settings> void Prover<settings>::execute_shplonk_round()
```
ledwards2225, line 342: Similarly, I think `nu` should be computed in this function rather than the preceding one. This corresponds to what is done in the Shplonk function `reduce_prove`. (Looks like that's also consistent with your comment so maybe this was just an accident). Just FYI in case you take a look: the challenges in the Shplonk code are called `rho` and `r` right now (same names as gemini), but I suppose they should be changed to `nu` and `z` or whatever we decide on

cpp/src/aztec/honk/sumcheck/polynomials/multivariates.hpp
```diff
@@ -104,9 +104,14 @@ template <class FF_, size_t num_polys, size_t num_vars> class Multivariates {
...
     std::array<FF, num_polys> batch_evaluate(std::array<FF, num_vars> input)
     {
-        // TODO(Cody): IOU implementation.
+        // TODO(Cody): these just get extracted from the folded multivariates
+        // For now, at least initialize properly.
         static_cast<void>(input);
-        return { { 1 } };
+        std::array<FF, num_polys> result;
+        for (auto& elt : result) {
+            elt = 1;
```
ledwards2225: `elt`?

cpp/src/aztec/plonk/proof_system/verifier/verifier.test.cpp
```diff
@@ -298,6 +298,7 @@ TEST(verifier, verify_arithmetic_proof_small)
     EXPECT_EQ(result, true);
 }
 
+// TODO(Cody): This test is suddenly very slow to run?
```
ledwards2225, line 301: Not sure why this would've changed but this is constructing a proof of non-trivial size so makes sense that it would take some time I suppose. Prover stuff here could be mocked like I was planning to do for the equivalent honk verifier test.

cpp/src/aztec/proof_system/flavor/flavor.hpp
```diff
@@ -0,0 +1,169 @@
...
+              { .name = "sigma_2",      .num_bytes = fr_size, .derived_by_verifier = false, .challenge_map_index = 4 },
+              { .name = "sigma_3",      .num_bytes = fr_size, .derived_by_verifier = false, .challenge_map_index = 5 },
+              { .name = "q_1",          .num_bytes = fr_size, .derived_by_verifier = false, .challenge_map_index = 6 },
+              { .name = "q_2",          .num_bytes = fr_size, .derived_by_verifier = false, .challenge_map_index = 7 },
+              { .name = "q_3",          .num_bytes = fr_size, .derived_by_verifier = false, .challenge_map_index = 8 },
+              { .name = "q_m",          .num_bytes = fr_size, .derived_by_verifier = false, .challenge_map_index = 9 },
+              { .name = "q_c",          .num_bytes = fr_size, .derived_by_verifier = false, .challenge_map_index = 10 },
+              { .name = "z_perm",       .num_bytes = fr_size, .derived_by_verifier = false, .challenge_map_index = 11 },
+              { .name = "z_perm_omega", .num_bytes = fr_size, .derived_by_verifier = false, .challenge_map_index = -1 },
```
ledwards2225: I dont think `z_perm_omega` (or the appropriate analog for Honk) will exist here right? And maybe this will include id polys and maybe a couple of lagrange polys? Probably not necessary to hammer out for this PR

cpp/src/aztec/proof_system/proving_key/proving_key.cpp
```diff
@@ -27,6 +27,7 @@ proving_key::proving_key(const size_t num_gates,
                          waffle::ComposerType type = waffle::STANDARD)
     : composer_type(type)
     , n(num_gates)
+    , log_n(numeric::get_msb(num_gates))
```
ledwards2225, line 30: ah, handy

ledwards2225 (changes requested):
Looks good to me. Minor comments about making the new Prover structure consistent with the functionality in Gemini/Sphlonk. I would propose simply disabling the Honk Verifier test to get this PR in. It fails because it relies on the actual proof construction in the Prover (rather than some mocked thing) which is changed in this work. I will get it working again in a follow-on that also incorporates the full Honk Manifest.

cpp/src/aztec/plonk/proof_system/verifier/verifier.test.cpp
codygunton: Makes sense, just thought I'd flag the change for a time in case we noticed some of our tests were unexpectedly slow--maybe this would give a clue.

cpp/src/aztec/honk/sumcheck/polynomials/multivariates.hpp
codygunton: "element". But you're anti-abbreviation, I can change.

cpp/src/aztec/proof_system/flavor/flavor.hpp
codygunton: Indeed.

cpp/src/aztec/honk/sumcheck/polynomials/multivariates.hpp
ledwards2225: hah I genuinely couldn't decipher that one

cpp/src/aztec/honk/proof_system/program_settings.hpp
codygunton: I couldn't think of a good name so I chose a reasonable name... might become something like `proving_system` or `proving_system_shared` but we don't know what all will be in this namespace and I'd rather not start with that and then have it polluted.

cpp/src/aztec/honk/composer/composer_helper/composer_helper.cpp
codygunton: Yeah unrolled is a terrible name. Plan is to kill the linearization trick, then I think we should just have `create_prover`. But for now we need both, even here, because `ComposerHelper` is generic and needs to hold stuff for Plonk.

cpp/src/aztec/honk/proof_system/program_settings.hpp
ledwards2225: What I don't like is that 1) it looks like a typo at first glance and 2) it's incredibly hard to search for

cpp/src/aztec/honk/proof_system/prover.cpp
codygunton: I agree this should move forward one round.

cpp/src/aztec/honk/proof_system/prover.cpp
codygunton: Thanks for pointing that out. I will combine these into a single `execute_pcs_evaluation_round`.

cpp/src/aztec/honk/proof_system/prover.cpp
codygunton: I agree.

cpp/src/aztec/honk/proof_system/program_settings.hpp
codygunton: FINE ;)

cpp/src/aztec/honk/sumcheck/polynomials/multivariates.hpp
codygunton: This is a clear sign you were in applied math.

codygunton:
Per our conversation, it seems natural that `execute_univariatization_round` and `execute_pcs_evaluation_round` will be merged and `execute_shplonk_round`, `execute_kzg_round` will be merged, so that the final four rounds of the honk manifest become two. However, it's not totally clear what should happen here, so in the interest of getting this PR in to avoid blocking others I leave this as a future task. This has the benefit of closely matching what is in the proving system outline, which currently diverges, notationally and in terms of the grouping semantics, of what we have in Gemini.

cpp/src/aztec/honk/composer/composer_helper/composer_helper.hpp
codygunton: Nope, removed.

cpp/src/aztec/honk/composer/composer_helper/composer_helper.cpp
codygunton: Leaving as future task.

PR #35, opened 2023-01-06
Is/adding honk permutations

Rumata888:
Added the implementation of sigma vector computation.


cpp/src/aztec/honk/circuit_constructors/standard_circuit_constructor.hpp
```diff
@@ -10,7 +10,8 @@ inline std::vector<std::string> standard_selector_names()
     return result;
 }
 
-class StandardCircuitConstructor : public CircuitConstructorBase {
+#define STANDARD_HONK_WIDTH 3
```
codygunton: Can this go in flavor.hpp?

cpp/src/aztec/honk/composer/composer_helper/permutation_helper.hpp
```diff
@@ -105,4 +110,60 @@ void compute_wire_copy_cycles(CircuitConstructor& circuit_constructor, CycleColl
...
+ */
+template <size_t program_width, typename CircuitConstructor>
+void compute_standard_honk_sigma_permutations(CircuitConstructor& circuit_constructor, proving_key* key)
+{
+    // Compute wire copy cycles for public and private variables
+    CycleCollector wire_copy_cycles;
+    compute_wire_copy_cycles<program_width>(circuit_constructor, wire_copy_cycles);
+    auto n = key->n;
+    // Fill sigma polynomials with default values
+    barretenberg::polynomial sigma_polynomials_lagrange[program_width];
```
codygunton: Is there a reason we use a C-style array here over some less cowboy container?

cpp/src/aztec/honk/composer/composer_helper/permutation_helper.hpp
```diff
@@ -105,4 +110,60 @@ void compute_wire_copy_cycles(CircuitConstructor& circuit_constructor, CycleColl
...
+        for (uint64_t j = 0; j < key->n; j++) {
+            sigma_polynomial_lagrange.coefficients[j] = (i * n + j);
+        }
+    }
+    // Go through each cycle
+    for (size_t i = 0; i < wire_copy_cycles.size(); i++) {
+        ASSERT(wire_copy_cycles[i].size() > 0);
+        size_t cycle_size = wire_copy_cycles[i].size();
+        cycle_node current_element = wire_copy_cycles[i][cycle_size - 1];
+        // Get the index value of the last element
```
codygunton, line 156: It would be helpful to move this comment up by one line.

cpp/src/aztec/honk/composer/composer_helper/permutation_helper.hpp
```diff
@@ -105,4 +110,60 @@ void compute_wire_copy_cycles(CircuitConstructor& circuit_constructor, CycleColl
...
+        // Construct permutation polynomials in lagrange base
+        std::string index = std::to_string(i + 1);
+        sigma_polynomials_lagrange[i] = barretenberg::polynomial(key->n);
+        auto& sigma_polynomial_lagrange = sigma_polynomials_lagrange[i];
+        for (uint64_t j = 0; j < key->n; j++) {
+            sigma_polynomial_lagrange.coefficients[j] = (i * n + j);
+        }
+    }
+    // Go through each cycle
+    for (size_t i = 0; i < wire_copy_cycles.size(); i++) {
```
codygunton: Would be easier to parse, IMO, if you gave `wire_copy_cycles[i]` a name.

codygunton: In fact, can't this be a ranged `for` loop?

cpp/src/aztec/honk/composer/standard_honk_composer.hpp
```diff
@@ -20,10 +20,11 @@ class StandardHonkComposer {
     // 1) Proving and verification keys
     // 2) CRS
     // 3) Converting variables to witness vectors/polynomials
-    ComposerHelper composer_helper;
+    ComposerHelper<StandardCircuitConstructor> composer_helper;
 
     // Leaving it in for now just in case
     bool contains_recursive_proof = false;
+    static constexpr size_t program_width = 3;
```
codygunton: Maybe this also belongs in flavor.hpp? That's a proto version of the future settings/arithmetization abstraction that I started in the sumcheck work because I needed to put some constants someplace.

cpp/src/aztec/honk/composer/standard_honk_composer.test.cpp
```diff
@@ -0,0 +1,86 @@
...
+    composer.create_add_gate({ a_idx, b_idx, c_idx, fr::one(), fr::one(), fr::neg_one(), fr::zero() });
+    composer.create_add_gate({ a_idx, b_idx, c_idx, fr::one(), fr::one(), fr::neg_one(), fr::zero() });
+    composer.create_add_gate({ a_idx, b_idx, c_idx, fr::one(), fr::one(), fr::neg_one(), fr::zero() });
+    composer.create_add_gate({ a_idx, b_idx, c_idx, fr::one(), fr::one(), fr::neg_one(), fr::zero() });
+    composer.create_add_gate({ a_idx, b_idx, c_idx, fr::one(), fr::one(), fr::neg_one(), fr::zero() });
+    composer.create_add_gate({ a_idx, b_idx, c_idx, fr::one(), fr::one(), fr::neg_one(), fr::zero() });
+    composer.create_add_gate({ a_idx, b_idx, c_idx, fr::one(), fr::one(), fr::neg_one(), fr::zero() });
+
+    auto proving_key = composer.compute_proving_key();
+    barretenberg::fr left = barretenberg::fr::one();
```
codygunton, line 62: Is there a reason you prefer this to using the literal `1` for initialization?

cpp/src/aztec/honk/composer/standard_honk_composer.test.cpp
```diff
@@ -0,0 +1,86 @@
...
+    // Now let's check that witness values correspond to the permutation
+
+    composer.compute_witness();
+    for (size_t i = 0; i < composer.program_width; i++) {
+
+        std::string index = std::to_string(i + 1);
+        auto permutation_polynomial = proving_key->polynomial_cache.get("sigma_" + index + "_lagrange");
+        auto witness_polynomial = proving_key->polynomial_cache.get("w_" + index + "_lagrange");
+        for (size_t j = 0; j < proving_key->n; j++) {
+            auto current_witness = witness_polynomial.coefficients[j];
```
codygunton, line 92: This is cool.

cpp/src/aztec/honk/composer/standard_honk_composer.test.cpp
```diff
@@ -0,0 +1,86 @@
+#include "standard_honk_composer.hpp"
+#include <gtest/gtest.h>
+
+TEST(standard_honk_composer, test_sigma_correctness)
```
codygunton, line 13: Are all the components in place for us to add a test of `assert_equal` as well?

codygunton (changes requested):
Looks good. I only raise some small points.

cpp/src/aztec/honk/composer/standard_honk_composer.test.cpp
Rumata888: Because it's already a value and doesn't require any conversion. I don't have a strong opinion on this, though

cpp/src/aztec/honk/composer/standard_honk_composer.test.cpp
Rumata888: Checking that the permutation works?

cpp/src/aztec/honk/composer/standard_honk_composer.test.cpp
Rumata888: Did we have a test like this somewhere in legacy?

cpp/src/aztec/honk/composer/standard_honk_composer.test.cpp
codygunton: The quick-and-dirty product argument in a test. Meta!

cpp/src/aztec/honk/composer/standard_honk_composer.test.cpp
codygunton: Maybe not? I think we probably should, but it's also true that `assert_equal` will be tested elsewhere (e.g., we should test verifier fails when an untrue `assert_equal` is made), so it's not necessary to add anything. I was thinking we could add a test illustrating how it works under the hood.

cpp/src/aztec/honk/composer/composer_helper/permutation_helper.hpp
```diff
@@ -50,19 +51,21 @@ typedef std::vector<std::vector<cycle_node>> CycleCollector;
...
+    auto& w_4 = circuit_constructor.w_4;
 
     size_t number_of_cycles = 0;
-    // Initialize wire_copy_cycles of public input variables to point to themselves
-    for (size_t i = 0; i < public_inputs.size(); ++i) {
+
+    const uint32_t num_public_inputs = static_cast<uint32_t>(public_inputs.size());
+    // Initialize wire_copy_cycles of public input variables to point to themselves ( we could actually ignore this step
+    // for HONK because of the way we construct the permutation)
+    for (size_t i = 0; i < num_public_inputs; ++i) {
```
adr1anh: `i` should be initialized as a `uint32_t` to match `num_public_inputs`

cpp/src/aztec/honk/composer/composer_helper/permutation_helper.hpp
```diff
@@ -50,19 +51,21 @@ typedef std::vector<std::vector<cycle_node>> CycleCollector;
...
     auto& w_r = circuit_constructor.w_r;
     auto& w_o = circuit_constructor.w_o;
-    if constexpr (program_width > 3) {
-        auto& w_4 = circuit_constructor.w_4;
-    }
-    auto& w_o = circuit_constructor.w_o;
+    auto& n = circuit_constructor.n;
+    auto& w_4 = circuit_constructor.w_4;
 
     size_t number_of_cycles = 0;
```
adr1anh, line 64: `number_of_cycles` should be a `uint32_t`

cpp/src/aztec/honk/composer/composer_helper/permutation_helper.hpp
```diff
@@ -50,19 +51,21 @@ typedef std::vector<std::vector<cycle_node>> CycleCollector;
 template <size_t program_width, typename CircuitConstructor>
 void compute_wire_copy_cycles(CircuitConstructor& circuit_constructor, CycleCollector& wire_copy_cycles)
 {
+    // Reference circuit constructor members
     auto& real_variable_index = circuit_constructor.real_variable_index;
     auto& public_inputs = circuit_constructor.public_inputs;
```
adr1anh: Adding `const` to these methods indicates that it these members are not modified.

cpp/src/aztec/honk/composer/composer_helper/permutation_helper.hpp
Rumata888: Done

cpp/src/aztec/honk/composer/composer_helper/permutation_helper.hpp
Rumata888: Done

cpp/src/aztec/honk/composer/composer_helper/permutation_helper.hpp
Rumata888: Done

cpp/src/aztec/honk/composer/standard_honk_composer.hpp
Rumata888: I moved the constant there, but the member is needed to make it easier to copy the tests (for example, permutation tests). This way the test can simply use the value defined in the class.

cpp/src/aztec/honk/composer/composer_helper/permutation_helper.hpp
```diff
@@ -50,19 +51,21 @@ typedef std::vector<std::vector<cycle_node>> CycleCollector;
...
     auto& w_l = circuit_constructor.w_l;
     auto& w_r = circuit_constructor.w_r;
     auto& w_o = circuit_constructor.w_o;
-    if constexpr (program_width > 3) {
-        auto& w_4 = circuit_constructor.w_4;
-    }
-    auto& w_o = circuit_constructor.w_o;
+    auto& n = circuit_constructor.n;
```
adr1anh: Change to `const uint32_t n = circuit_constructor.n;`. The references is not necessary since it is a small value, and it ensures we're working with the same `uint32_t` type everywhere.

cpp/src/aztec/honk/composer/composer_helper/permutation_helper.hpp
```diff
@@ -76,16 +79,16 @@ void compute_wire_copy_cycles(CircuitConstructor& circuit_constructor, CycleColl
         cycle.emplace_back(right);
     }
 
-    const uint32_t num_public_inputs = static_cast<uint32_t>(public_inputs.size());
-
     // Go through all witnesses and add them to the wire_copy_cycles
     for (size_t i = 0; i < n; ++i) {
```
adr1anh: `i` should be initialized as `uint32_t`

cpp/src/aztec/honk/composer/composer_helper/permutation_helper.hpp
```diff
@@ -76,16 +79,16 @@ void compute_wire_copy_cycles(CircuitConstructor& circuit_constructor, CycleColl
...
         const auto w_1_index = real_variable_index[w_l[i]];
         const auto w_2_index = real_variable_index[w_r[i]];
         const auto w_3_index = real_variable_index[w_o[i]];
```
adr1anh: These are `uint32_t` and could be marked as such instead of `auto`.

cpp/src/aztec/honk/composer/composer_helper/permutation_helper.hpp
```diff
@@ -94,7 +97,9 @@ void compute_wire_copy_cycles(CircuitConstructor& circuit_constructor, CycleColl
         wire_copy_cycles[static_cast<size_t>(w_3_index)].emplace_back(static_cast<uint32_t>(i + num_public_inputs),
```
adr1anh, line 102: `static_cast`s are not required with the proposed changes from above.

cpp/src/aztec/honk/composer/composer_helper/permutation_helper.hpp
```diff
@@ -94,7 +97,9 @@ void compute_wire_copy_cycles(CircuitConstructor& circuit_constructor, CycleColl
...
                                                                       WireType::OUTPUT);
 
+        // Handle width 4 separately
         if constexpr (program_width > 3) {
+            static_assert(program_width == 4);
             const auto w_4_index = real_variable_index[w_4[i]];
             if (w_4_index >= number_of_cycles) {
                 wire_copy_cycles.resize(w_4_index + 1);
```
adr1anh, line 110: Resizing inside a loop can cause a large number of allocations since the size only increases by 1 each time it is called (the whole vector needs to be reallocated upon size change). By removing this line, the `emplace_back` will take care of growing the buffer to minimize the number of times the size needs to increase.

cpp/src/aztec/honk/composer/composer_helper/permutation_helper.hpp
```diff
@@ -76,16 +79,16 @@ void compute_wire_copy_cycles(CircuitConstructor& circuit_constructor, CycleColl
...
 
-    const uint32_t num_public_inputs = static_cast<uint32_t>(public_inputs.size());
-
     // Go through all witnesses and add them to the wire_copy_cycles
     for (size_t i = 0; i < n; ++i) {
         const auto w_1_index = real_variable_index[w_l[i]];
         const auto w_2_index = real_variable_index[w_r[i]];
         const auto w_3_index = real_variable_index[w_o[i]];
+        // Check the maximum index of a variable. If it is more or equal to the cycle vector size, extend the vector
         auto max_index = std::max({ w_1_index, w_2_index, w_3_index });
```
adr1anh: Replace `auto` with `const uint32_t`

adr1anh (commented):
Added some comments that came up while trying to understand the public input permutation trick.

cpp/src/aztec/honk/composer/composer_helper/permutation_helper.hpp
Rumata888: Nice catch. Addressed

cpp/src/aztec/honk/composer/composer_helper/permutation_helper.hpp
Rumata888: I agree. This is probably the most inefficient way to do it, too, since by just starting from the end, I'd quickly get the actual maximum value. I'll delete it. Thanks

Rumata888: Ah, no. I was wrong. We can't simply emplace_back wire_copy_cycles vector. It needs to be resized, because it has holes. I will start from the end, though, to minimize the number of reallocations.

cpp/src/aztec/honk/composer/composer_helper/permutation_helper.hpp
Rumata888: Yep

cpp/src/aztec/honk/composer/composer_helper/permutation_helper.hpp
Rumata888: Done

cpp/src/aztec/honk/composer/composer_helper/permutation_helper.hpp
Rumata888: I'm not sure that's the best way. Whenever we are using a variable as indexer in vectors/arrays it is better to initialize it as size_t and then convert it to uint32_t, when we need to.

cpp/src/aztec/honk/composer/composer_helper/permutation_helper.hpp
Rumata888: It's actually size_t originally.

cpp/src/aztec/honk/composer/composer_helper/permutation_helper.hpp
Rumata888: I agree that it's better to use const here.

cpp/src/aztec/honk/composer/composer_helper/permutation_helper.hpp
Rumata888: We use it to resize a vector, so it's better to leave size_t.

codygunton (approved):

PR #36, opened 2023-01-06
[REPLACED] SRS Transcript Cleanup

suyash67:
# REPLACED BY PR https://github.com/AztecProtocol/barretenberg/pull/93

PR #39, opened 2023-01-09
[REPLACED] Commit to Wires Using Lagrange SRS

suyash67:
# REPLACED BY PR https://github.com/AztecProtocol/barretenberg/pull/94


suyash67:
Note: we need to merge #36 before this one.

PR #40, opened 2023-01-09
Lde/honk verifier base

ledwards2225:
- Adding Honk verifier skeleton plus initial tests 
- Adding program_settings for Standard Honk
- Making Prover/Verifier naming more consistent

### DIFF
--- a/cpp/src/aztec/honk/proof_system/verifier.cpp
+++ b/cpp/src/aztec/honk/proof_system/verifier.cpp
@@ -1,0 +1,114 @@
+#include <common/throw_or_abort.hpp>
+#include <plonk/proof_system/constants.hpp>
+#include "./verifier.hpp"
+#include "../../plonk/proof_system/public_inputs/public_inputs.hpp"
+#include <ecc/curves/bn254/fq12.hpp>
+#include <ecc/curves/bn254/pairing.hpp>
+#include <ecc/curves/bn254/scalar_multiplication/scalar_multiplication.hpp>
+#include <polynomials/polynomial_arithmetic.hpp>
+
+using namespace barretenberg;
+using namespace honk::sumcheck;
+
+namespace honk {
+template <typename program_settings>
+HonkVerifier<program_settings>::HonkVerifier(std::shared_ptr<waffle::verification_key> verifier_key,
+                                             const transcript::Manifest& input_manifest)
+    : manifest(input_manifest)
+    , key(verifier_key)
+{}
+
+template <typename program_settings>
+HonkVerifier<program_settings>::HonkVerifier(HonkVerifier&& other)
+    : manifest(other.manifest)
+    , key(other.key)
+    , commitment_scheme(std::move(other.commitment_scheme))
+{}
+
+template <typename program_settings>
+HonkVerifier<program_settings>& HonkVerifier<program_settings>::operator=(HonkVerifier&& other)
+{
+    key = other.key;
+    manifest = other.manifest;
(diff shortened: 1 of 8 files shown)


cpp/src/aztec/honk/proof_system/verifier.cpp
```diff
@@ -0,0 +1,121 @@
...
+        w_i(X),        i = 1,2,3
+        sigma_i(X),    i = 1,2,3
+        q_i(X),        i = 1,2,3,4,5
+        z_perm(X),
+        L_0(X),
+        id(X)
+
+    Univariate evaluations:
+        a_0 = Fold_{-r}^(0)(-r),
+        a_l = Fold^(l)(-r^{2^l}), i = 1,...,d-1
```
codygunton: Index is `l`. Also, this is just uniformly written as `a_l = Fold_{-r}^(l)(-r^{2^l}), i = 0,...,d-1`

cpp/src/aztec/honk/proof_system/verifier.cpp
```diff
@@ -0,0 +1,121 @@
...
+                             static_cast<uint8_t>(key->n >> 8),
+                             static_cast<uint8_t>(key->n) });
+
+    transcript.add_element("public_input_size",
+                           { static_cast<uint8_t>(key->num_public_inputs >> 24),
+                             static_cast<uint8_t>(key->num_public_inputs >> 16),
+                             static_cast<uint8_t>(key->num_public_inputs >> 8),
+                             static_cast<uint8_t>(key->num_public_inputs) });
+
+    // Compute challenges from the proof data, based on the manifest, using the Fiat-Shamir heuristic
```
codygunton, line 97: Don't we have to get more challenges than this? E.g., we need to generate `u` and then compute the values of some polynomials there.

cpp/src/aztec/honk/proof_system/verifier.test.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+    StandardVerifier verifier = TestFixture::generate_verifier(state.key);
+
+    // construct proof
+    waffle::plonk_proof proof = state.construct_proof();
+
+    // verify proof
+    bool result = verifier.verify_proof(proof);
+
+    EXPECT_EQ(result, true);
```
codygunton, line 230: Note: we currently just define `result` to be `1` :wink:

codygunton (changes requested):

cpp/src/aztec/honk/proof_system/verifier.cpp
ledwards2225: I don't think so - a_0 is an evaluation of the partially evaluated batch polynomial, whereas l = 1, ..., d-1 are not

cpp/src/aztec/honk/proof_system/verifier.cpp
ledwards2225: Yeah but they have to exist in the manifest. To me it made sense to just wait for that to be done rather than duplicating your manifest work by mocking it

cpp/src/aztec/honk/proof_system/verifier.cpp
codygunton: My bad -- I was assuming a typo in the writeup, but I was wrong about that.

cpp/src/aztec/honk/proof_system/verifier.cpp
ledwards2225: Improved comment for future TODO on this

PR #43, opened 2023-01-10
Is/adding power polynomial

Rumata888:
Added power polynomial generation and evaluation


cpp/src/aztec/honk/utils/power_polynomial.hpp
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * @brief Evaluate the power polynomial on {x_0,..,x_d}
+ *
+ * @details The power polynomial can be efficiently evaluated as ∏( ( b^{2^i} - 1 ) * x_i + 1)
+ *
+ * @param zeta ζ
+ * @param variables
+ * @return barretenberg::fr
+ */
+barretenberg::fr evaluate(barretenberg::fr zeta, const std::vector<barretenberg::fr>& variables)
```
adr1anh: Consider `std::span<const barretenberg::fr> variables)` instead of `const std::vector<barretenberg::fr>& variables)`, so that this function can accept any type of contiguous range (like `std::array`, `std::vector`, `std::initializer_list`, or even classic fixed size C arrays)

cpp/src/aztec/honk/utils/power_polynomial.hpp
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * @brief Generate the power polynomial vector
+ *
+ * @details Generates a vector, where v[i]=ζⁱ
+ *
+ * @param zeta
+ * @param vector_size
+ * @return barretenberg::polynomial
+ */
+barretenberg::polynomial generate_vector(barretenberg::fr zeta, size_t vector_size)
```
adr1anh: In the upcoming Gemini PR, the polynomial class has been templated to `Polynomial<Fr>`, since we don't do anything specific to BN254Fr, this would be a good candidate to be made generic.

cpp/src/aztec/honk/utils/power_polynomial.hpp
```diff
@@ -0,0 +1,65 @@
...
+{
+    barretenberg::polynomial pow_vector(vector_size);
```
adr1anh: If the size does not grow, consider `pow_vector(vector_size, vector_size);` so that we don't over-allocate.

cpp/src/aztec/honk/utils/power_polynomial.hpp
```diff
@@ -0,0 +1,65 @@
...
+    if ((num_threads % thread_size) != 0) {
+        thread_size += 1;
+    }
+#ifndef NO_MULTITHREADING
+#pragma omp parallel for
+#endif
+    for (size_t i = 0; i < num_threads; i++) {
+        barretenberg::fr starting_power = zeta.pow(i * thread_size);
+        for (size_t j = 0; j < (thread_size - 1); j++) {
+            pow_vector.coefficients[i * thread_size + j] = starting_power;
```
adr1anh: Consider `pow_vector[i * thread_size + j]` as this allows for bound checking in debug builds? The only problem is we run tests in release with asserts so that might slow down testing. Curious to hear your thoughts.

cpp/src/aztec/honk/utils/power_polynomial.test.cpp
```diff
@@ -0,0 +1,34 @@
...
+    barretenberg::polynomial power_polynomial = honk::power_polynomial::generate_vector(zeta, n);
+
+    barretenberg::fr current_power = barretenberg::fr::one();
+    for (size_t i = 0; i < n; i++) {
+        EXPECT_EQ(power_polynomial.coefficients[i], current_power);
+        if (power_polynomial.coefficients[i] != current_power) {
+            break;
+        }
+        current_power *= zeta;
+    }
```
adr1anh: This is nit-picking, but we could use the existing `polynomial::evaluate()` function with the polynomial whose coefficients are all ones, and evaluate it in `zeta`.

cpp/src/aztec/honk/utils/power_polynomial.test.cpp
```diff
@@ -0,0 +1,34 @@
...
+}
+
+TEST(power_polynomial, test_evaluation_correctness)
```
adr1anh, line 22: Once the Gemini PR comes in, we could test this function against the `polynomial::evaluate_mle()` in a random multi-variate point to test the correctness.

adr1anh (commented):
Good stuff! Perhaps the use of multi-threading at this time is a bit premature since the rest of honk still lacks this. It would be nice to make this generic once the Gemini PR gets in, but that can come at a later time.

cpp/src/aztec/honk/utils/power_polynomial.hpp
Rumata888: Will do

cpp/src/aztec/honk/utils/power_polynomial.test.cpp
Rumata888: Wouldn't it be inefficient? Here I am sequentially checking the whole vector

cpp/src/aztec/honk/utils/power_polynomial.hpp
Rumata888: Are you talking about the new polynomial? The old one only checks if the polynomial is empty

cpp/src/aztec/honk/utils/power_polynomial.hpp
Rumata888: Did

cpp/src/aztec/honk/utils/power_polynomial.hpp
Rumata888: Yes, you were. Changed

PR #45, opened 2023-01-10
Adding ID and edge Lagrange polynomials to the proving key

Rumata888:


cpp/src/aztec/honk/composer/composer_helper/permutation_helper.hpp
```diff
@@ -175,4 +175,50 @@ void compute_standard_honk_sigma_permutations(CircuitConstructor& circuit_constr
...
+        key->polynomial_cache.put("id_" + index + "_lagrange", std::move(id_polynomials_lagrange[i]));
+    }
+}
+
+/**
+ * @brief Compute Lagrange Polynomials L_0 and L_{n-1} and put them in the polynomial cache
+ *
+ * @param key Proving key where we will save the polynomials
+ */
+void compute_edge_lagrange_polynomials(proving_key* key)
```
ledwards2225: I have two requests for this function: 1) Change the name to avoid the term 'edge'. We've used that term to refer to an 'edge' of the hypercube and I thought at first glance that this function had something to do with that - maybe `compute_first_and_last_lagrange_polynomials()`? 2) Add a simple test just for completeness even though the function is quite simple.

ledwards2225 (changes requested):
Looks great. Minor requests for name change + a simple test

PR #49, opened 2023-01-11
Add transcript_tests and remove unneeded script.

codygunton:


codygunton:
These commits were cherry-picked into #50.

PR #50, opened 2023-01-11
Lde/fix univariate serialization

ledwards2225:
- Fix univariate serialization
- Reinstate transcript and proof_system tests on CCI


codygunton:
We discussed this PR on a call.

PR #53, opened 2023-01-11
Cg/some commitments in prover

codygunton:
Compute actual commitments in the pre-sumcheck prover rounds.

### DIFF
--- a/cpp/src/aztec/polynomials/polynomial.hpp
+++ b/cpp/src/aztec/polynomials/polynomial.hpp
@@ -76,6 +76,8 @@ template <typename Fr> class Polynomial {
 
     Fr* get_coefficients() const { return coefficients_; };
     Fr* get_coefficients() { return coefficients_; };
+    // TODO(Cody): This is a hack.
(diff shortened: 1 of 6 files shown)


cpp/src/aztec/polynomials/polynomial.hpp
```diff
@@ -76,6 +76,8 @@ template <typename Fr> class Polynomial {
 
     Fr* get_coefficients() const { return coefficients_; };
     Fr* get_coefficients() { return coefficients_; };
+    // TODO(Cody): This is a hack.
+    std::span<Fr> get_coefficients_span() { return std::span<Fr>(coefficients_, size()); };
```
Rumata888: Wouldn't it be better to define a conversion function?

Rumata888 (changes requested):
Everything seems ok. I'd just make 1 small change

cpp/src/aztec/polynomials/polynomial.hpp
codygunton: Yes, addressed that as you suggested.

Rumata888 (approved):
Everything looks good

PR #58, opened 2023-01-12
Fix debug build in x86_64-linux-gcc10

codygunton:
An implicit conversion inside of an `ASSERT` prevented building using the x86_64-linux-gcc10 toolchain in debug mode. We fix this.


codygunton:
@ledwards2225 pointed out this issue out to me. The change is minuscule and clearly uncontroversial, so I don't bother requesting review.

PR #59, opened 2023-01-12
Move IPA into pcs/

codygunton:
IPA was in honk/commitment_scheme. I move IPA rather than Gemini because IPA is not currently hooked up to other parts of Honk.


codygunton:
This is a very simple and uncontroversial move-only PR, so I don't request review.

PR #61, opened 2023-01-12
Prover constructs sumcheck

codygunton:
This PR makes the Honk Prover class construct a sumcheck object and execute this. Some of the higher-level changes this necessitates.

- It may be desirable to pre-compute sumcheck classes for some expected sizes (values of `multivariate_d`, but at least for now I make this size a runtime parameter. This necessitates replacing some uses of `std::array` with `std::vector`.
- Since Standard Honk and Standard Plonk have different sets of polynomials, we cannot reuse the Standard Plonk polynomial manifest (or its size). The polynomial manifest is selected using the Composer type, so I added another composer type for Standard Honk.
- I add a constructor of a `Multivariates` object from a proving key (needed for prover to create the object) and from a transcript (Sumcheck class in verifier mode had previously not constructed multivariates).

I regret leaving the insertion of the sumcheck evaluations in a rough form in the prover. This is done for expediency.

### DIFF
--- a/cpp/src/aztec/honk/composer/composer_helper/composer_helper.cpp
+++ b/cpp/src/aztec/honk/composer/composer_helper/composer_helper.cpp
@@ -43,8 +43,8 @@ std::shared_ptr<waffle::proving_key> ComposerHelper<CircuitConstructor>::compute
 
     // Initialize circuit_proving_key
     // TODO: replace composer types.
-    circuit_proving_key =
-        std::make_shared<waffle::proving_key>(subgroup_size, public_inputs.size(), crs, waffle::ComposerType::STANDARD);
+    circuit_proving_key = std::make_shared<waffle::proving_key>(
+        subgroup_size, public_inputs.size(), crs, waffle::ComposerType::STANDARD_HONK);
 
     for (size_t i = 0; i < constructor.num_selectors; ++i) {
         std::vector<barretenberg::fr>& selector_values = selectors[i];
@@ -225,7 +225,7 @@ std::shared_ptr<waffle::proving_key> ComposerHelper<CircuitConstructor>::compute
         return circuit_proving_key;
     }
     // Compute q_l, q_r, q_o, etc polynomials
-    ComposerHelper::compute_proving_key_base(circuit_constructor, waffle::ComposerType::STANDARD);
+    ComposerHelper::compute_proving_key_base(circuit_constructor, waffle::ComposerType::STANDARD_HONK);
 
     // Compute sigma polynomials (we should update that late)
     compute_standard_honk_sigma_permutations<CircuitConstructor::program_width>(circuit_constructor,
--- a/cpp/src/aztec/honk/composer/composer_helper/composer_helper.cpp
+++ b/cpp/src/aztec/honk/composer/composer_helper/composer_helper.cpp
@@ -234,6 +234,7 @@ std::shared_ptr<waffle::proving_key> ComposerHelper<CircuitConstructor>::compute
 
     compute_first_and_last_lagrange_polynomials(circuit_proving_key.get());
 
+    // TODO(Cody): this is a workaround
     circuit_proving_key->polynomial_cache.put("z_perm", Polynomial<barretenberg::fr>(1));
 
     return circuit_proving_key;
--- a/cpp/src/aztec/honk/composer/composer_helper/composer_helper.cpp
+++ b/cpp/src/aztec/honk/composer/composer_helper/composer_helper.cpp
@@ -1,4 +1,5 @@
 #include "composer_helper.hpp"
+#include "polynomials/polynomial.hpp"
 #include <cstddef>
 #include <proof_system/flavor/flavor.hpp>
 #include <honk/pcs/commitment_key.hpp>
@@ -232,6 +233,8 @@ std::shared_ptr<waffle::proving_key> ComposerHelper<CircuitConstructor>::compute
     compute_standard_honk_id_polynomials<CircuitConstructor::program_width>(circuit_proving_key.get());
 
     compute_first_and_last_lagrange_polynomials(circuit_proving_key.get());
+
+    circuit_proving_key->polynomial_cache.put("z_perm", Polynomial<barretenberg::fr>(1));
 
     return circuit_proving_key;
 }
--- a/cpp/src/aztec/honk/composer/standard_honk_composer.test.cpp
+++ b/cpp/src/aztec/honk/composer/standard_honk_composer.test.cpp
@@ -2,7 +2,7 @@
 #include "numeric/uint256/uint256.hpp"
 #include <cstdint>
 #include <honk/proof_system/prover.hpp>
-
+#include <honk/sumcheck/polynomials/multivariates.hpp>
 #include <gtest/gtest.h>
 
 using namespace honk;
@@ -264,7 +264,10 @@ TEST(StandarHonkComposer, BaseCase)
 
     auto prover = composer.create_unrolled_prover();
     // waffle::Verifier verifier = composer.create_verifier();
+    // TODO(Cody): multivariate_d can't be a template parameter.
+    auto multivariates = honk::sumcheck::Multivariates<fr, 17, 1>(prover.proving_key);
 
+    // Next up: construct sumcheck module from multivariates.
     waffle::plonk_proof proof = prover.construct_proof();
 
     // bool result = verifier.verify_proof(proof); // instance, prover.reference_string.SRS_T2);
--- a/cpp/src/aztec/honk/composer/standard_honk_composer.test.cpp
+++ b/cpp/src/aztec/honk/composer/standard_honk_composer.test.cpp
@@ -265,8 +265,8 @@ TEST(StandarHonkComposer, BaseCase)
     auto prover = composer.create_unrolled_prover();
     // waffle::Verifier verifier = composer.create_verifier();
     // TODO(Cody): multivariate_d can't be a template parameter.
-    auto multivariates = honk::sumcheck::Multivariates<fr, 17, 1>(prover.proving_key);
-    (void)multivariates;
+    // auto multivariates = honk::sumcheck::Multivariates<fr, 17, 1>(prover.proving_key);
+    // (void)multivariates;
     // Next up: construct sumcheck module from multivariates.
     waffle::plonk_proof proof = prover.construct_proof();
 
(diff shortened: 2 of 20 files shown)


cpp/src/aztec/honk/composer/standard_honk_composer.test.cpp
```diff
@@ -264,7 +264,8 @@ TEST(StandarHonkComposer, BaseCase)
 
     auto prover = composer.create_unrolled_prover();
     // waffle::Verifier verifier = composer.create_verifier();
-
+    auto multivariates = honk::sumcheck::Multivariates<fr, 17>(prover.proving_key);
```
ledwards2225: TODO: Replace magic 17 with `STANDARD_HONK_MANIFEST_SIZE`

cpp/src/aztec/honk/proof_system/prover.cpp
```diff
@@ -273,32 +280,40 @@ template <typename settings> void Prover<settings>::execute_relation_check_round
...
-    transcript.add_element("q_3", barretenberg::fr(108).to_buffer());
-    transcript.add_element("q_m", barretenberg::fr(109).to_buffer());
-    transcript.add_element("q_c", barretenberg::fr(110).to_buffer());
-    transcript.add_element("z_perm", barretenberg::fr(111).to_buffer());
+    auto multivariates = Multivariates(proving_key);
+    auto sumcheck = Sumcheck(multivariates, transcript);
+
+    sumcheck.execute_prover();
+
+    // TODO(Cody): Execute as a loop over polynomial manifest? Things thare are called *_lagrange
```
ledwards2225, line 299: Oh yeah that should not be the case. I think Kesha may have added the _lagrange suffix because he didn't like that it was explicitly applied in various places in the code. But _lagrange was only used in plonk because there are alternatives (_monomial, _fft). It was kept for Honk for consistency. We should revert to applying it explicitly where it's needed or just get rid of it entirely since there is no ambiguity in Honk.

cpp/src/aztec/honk/sumcheck/polynomials/multivariates.hpp
```diff
@@ -54,25 +57,54 @@ bool span_arrays_equal(auto& lhs, auto& rhs)
...
+        : multivariate_n(full_polynomials[0].size())
+        , multivariate_d(numeric::get_msb(multivariate_n))
+        , full_polynomials(full_polynomials)
+    {
+        for (auto& polynomial : folded_polynomials) {
+            polynomial.resize(multivariate_n >> 1);
+        }
+    };
+
+    explicit Multivariates(const std::shared_ptr<waffle::proving_key>& proving_key)
```
ledwards2225, line 86: To be updated to handle shifted polynomials in follow-on

cpp/src/aztec/honk/sumcheck/sumcheck.test.cpp
```diff
@@ -70,7 +74,18 @@ TEST(Sumcheck, Prover)
         q_c, sigma_1, sigma_2, sigma_3, id_1,         id_2, id_3, lagrange_1
     };
 
-    auto transcript = Transcript(transcript::Manifest());
+    std::vector<transcript::Manifest::RoundManifest> manifest_rounds;
```
ledwards2225, line 77: Follow-on: Stuff like this and `mock_prover_contributions_to_transcript` could be replaced with the real manifest and the `mock_inputs_prior_to_challenge()` fctn

cpp/src/aztec/honk/sumcheck/sumcheck_round.test.cpp
```diff
@@ -22,11 +22,11 @@ namespace test_sumcheck_round {
 TEST(SumcheckRound, ComputeUnivariateProver)
 {
     const size_t num_polys(proving_system::StandardArithmetization::NUM_POLYNOMIALS);
-    const size_t multivariate_d(1);
+    // const size_t multivariate_d(1);
```
ledwards2225, line 25: Looks like some of the commented out stuff in this file can just be deleted? can be done in follow-on

cpp/src/aztec/plonk/proof_system/types/polynomial_manifest.hpp
```diff
@@ -170,6 +170,29 @@ static constexpr PolynomialDescriptor ultra_polynomial_manifest[ULTRA_UNROLLED_M
...
 };
 
+// TODO(Cody): Get this right; just using for now to extract names.
+static constexpr size_t STANDARD_HONK_MANIFEST_SIZE = 17;
+static constexpr PolynomialDescriptor standard_honk_polynomial_manifest[STANDARD_HONK_MANIFEST_SIZE]{
+    PolynomialDescriptor("W_1", "w_1_lagrange", false, false, WITNESS, W_1), //
+    PolynomialDescriptor("W_2", "w_2_lagrange", false, false, WITNESS, W_2), //
+    PolynomialDescriptor("W_3", "w_3_lagrange", false, false, WITNESS, W_3), //
+    PolynomialDescriptor("Z_PERM", "z_perm", true, true, WITNESS, Z),        //
+    // PolynomialDescriptor("Z_PERM_SHIFT", "z_perm_shift", true, true, WITNESS, Z_LOOKUP),                     //
```
ledwards2225, line 180: To be reinstated in follow-on

cpp/src/aztec/proof_system/flavor/flavor.hpp
```diff
@@ -11,7 +11,7 @@ struct StandardArithmetization {
         W_R,
         W_O,
         Z_PERM,
-        Z_PERM_SHIFT,
+        Z_PERM_SHIFT, // TODO(Cody): Hid ethis.
```
ledwards2225, line 14: To be addressed in follow-on

cpp/src/aztec/proof_system/flavor/flavor.hpp
```diff
@@ -36,7 +36,8 @@ struct StandardHonk {
   public:
     using Arithmetization = proving_system::StandardArithmetization;
     using MULTIVARIATE = Arithmetization::POLYNOMIAL;
-    static constexpr size_t STANDARD_UNROLLED_MANIFEST_SIZE = 12; // cf waffle::STANDARD_UNROLLED_MANIFEST_SIZE
+    // TODO(Cody): Where to specify? is this polynomial manifest size?
+    static constexpr size_t STANDARD_HONK_MANIFEST_SIZE = 16;
```
ledwards2225: Can this just be deleted? Not currently used anywhere. Or should the value be specified here then used by the polynomial manifest?

cpp/src/aztec/proof_system/flavor/flavor.hpp
```diff
@@ -100,23 +101,28 @@ struct StandardHonk {
...
+              { .name = "sigma_3", .num_bytes = fr_size, .derived_by_verifier = false, .challenge_map_index = 11 },
+              { .name = "id_1",    .num_bytes = fr_size, .derived_by_verifier = false, .challenge_map_index = 12 },
+              { .name = "id_2",    .num_bytes = fr_size, .derived_by_verifier = false, .challenge_map_index = 13 },
+              { .name = "id_3",    .num_bytes = fr_size, .derived_by_verifier = false, .challenge_map_index = 14 },
+              { .name = "L_first", .num_bytes = fr_size, .derived_by_verifier = false, .challenge_map_index = 15 },
+              { .name = "L_last",  .num_bytes = fr_size, .derived_by_verifier = false, .challenge_map_index = 16 },
             },
             /* challenge_name = */ "rho",
-            /* num_challenges_in = */ STANDARD_UNROLLED_MANIFEST_SIZE - 1, /* TODO(Cody): this is bad. */
+            /* num_challenges_in = */ 11, /* TODO(Cody): magic number! Where should this be specified? */
```
ledwards2225, line 125: To be addressed in follow-on

cpp/src/aztec/honk/composer/composer_helper/composer_helper.cpp
```diff
@@ -233,6 +234,9 @@ std::shared_ptr<waffle::proving_key> ComposerHelper<CircuitConstructor>::compute
 
     compute_first_and_last_lagrange_polynomials(circuit_proving_key.get());
 
+    // TODO(Cody): this is a workaround
```
ledwards2225, line 237: To be addressed in follow-on

ledwards2225 (approved):

cpp/src/aztec/honk/composer/standard_honk_composer.test.cpp
codygunton: Fixed.

cpp/src/aztec/honk/proof_system/prover.cpp
codygunton: Agree. Let's address this in a future PR.

cpp/src/aztec/proof_system/flavor/flavor.hpp
codygunton: I left it in only as a comment, as a suggestion that perhaps this data should be specified as part of the flavor.

cpp/src/aztec/honk/proof_system/prover.cpp
codygunton: It requires changes to `ComposerHelper`, which is supposed to be Plonk/Honk agnostic, so I think it makes sense to defer until we do the unification work.

cpp/src/aztec/honk/sumcheck/sumcheck_round.test.cpp
codygunton: Yeah, some of it can go, but it might be beneficial to preserve some of the info in comments (maybe it adds context to make the example more legible).

cpp/src/aztec/honk/sumcheck/sumcheck.test.cpp
codygunton: Yeah, I think, more generally, we need to decide on a policy about when we use real things vs mocked things. @arijitdutta67 maybe you can keep this in mind as you work on the sumcheck tests and implement @ledwards2225's suggestion if you think it makes sense to do so?

cpp/src/aztec/honk/proof_system/prover.cpp
ledwards2225: Gotcha, ok

PR #62, opened 2023-01-13
Public input permutation

### LINKED ISSUE #22: Handle public inputs in grand product in Honk PoC
Include public inputs via permutation argument in Honk.

### LINKED ISSUE #28: Update grand product relation
Assuming we adopt the form where 'polynomials' have 0 in the first and last index, we only need one grand product relation of the form [(Z_{shift,i} + L_{n-1,i}) * G_i] - [(Z_i + L_{0,i}) * F_i].
- Update the current grand product computation relation
- Remove the no longer needed grand product initialization relation

adr1anh:
Fixes #28, #22 

- comment-out work queue
- change sumcheck initialization round to check that the last element is 0 (not necessary now since we don't do ZK, but creates the least amount of change) 
- reinstore gamma challenge 

## Clang-tidy related

There are many changes that silence warnings from `clang-tidy`. 
- Initializing class members when possible 
- Explicitly deleting copy constructors/assignment operators for non-copyable classes
- Using `= default` for trivial destructors
- Adding `noexcept` to move constructors/assignment
- Using `auto` after `static_cast` to avoid duplication.
- Using `std::move` when copying an argument like a `string` or `vector` to help the compiler (references may alias, so passing by value tells the compiler that this value is really constant)
- Removing unused headers, and adding `#pragma once` to prevent import clashes. 
- Reorder some imports, and use the module-relative paths
- Replace some `typedef` with `using` 
- Use `const auto& element : container` to prevent unnecessary copies.

## Circuit Constructor

-  Replace `n` by `num_gates` to prevent clashes with `n` we use to refer to the `subgroup_size` 
- Added comments about things that seemed fishy, (all marked with `TODO(Adrian)` 
- Removed `NUM_RESERVED_GATES` and `WireType` since they are not used/duplicated
- Simplified `set_public_input`, though not sure if the removal of the `ASSERT` is correct. 

## Composer 

- Refactor `compute_proving_key_base`
  - Take the `CircuitConstructor` by `const` reference and modify the behavior so we only modify the `proving_key`. We no longer add any padding or dummy gates to the constructor, and instead do the padding entirely over the polynomials. 
  - Remove unnecessary zero-ing out of selector values by using the fact that polynomials are initialized to 0. 
  - Explicitly state where the public inputs are stored. 
- Refactor `compute_witness_base`
  - Use explicit types when referencing the `circuit_constructor` and ensure we only get objects that we don't modify. 
  - Create `array` of wires to handle the `program_width` more generally. 
  - No longer modify the `circuit_constructor` wires to add padding, and instead add 0-padding to the wire polynomials. 
  - Remove `fr::__copy` 
- Ensure all calls to `circuit_constructor` are `const` 
- PermutationHelper: Big refactor to clarify handling of public inputs. 
  - Simplify `cycle_node` behavior, and more generically handle different number of columns, and remove confusing `WireType` enum. 
  - Make `compute_wire_copy_cycles` return a vector of `CyclicPermutation` for clarity. 
  - Remove `resize` by noting that the number of cycles is equal to the number of `variables`
  - Reverse the order in which we were applying each `CyclicPermutation` to the `sigma` polynomials. Now, each `cycle_node` will map to the next one in the list. 
  - Add comments to explain what parts of the function are necessary for our public input handling. 
  - Changed the `composer_test` to also test for public inputs. 
    - Add many more tests to ensure the permutation polynomials that we create have the expected form. 
    - Compare results with the `public_input_delta`

## Prover & Verifier

- Comment-out the `work_queue` related members. 
- Restore `gamma` challenge that was removed due to a misunderstanding
- Remove default argument `beta = 1` for the `grand_product` computation.

## Sumcheck 

- Add `#pragma once` to headers
- Modify the `GrandProductComputationRelation` to work with `Z_perm` that has the first coefficient equal to 0. 
- Handle `public_input_delta` in `GrandProductComputationRelation` 
- Modify `GrandProductInitializationRelation` to instead check that the last "real" value of `Z_perm_shift` is 0. In the current ZK-less situation, this is not necessary since it will be guaranteed by Gemini shift opening. But leaving it here for later. 
- Renamed `LAGRANGE_1` to `LAGRANGE_FIRST` to be consistent with other parts of the code.

### DIFF
--- a/cpp/src/aztec/common/assert.hpp
+++ b/cpp/src/aztec/common/assert.hpp
@@ -10,8 +10,12 @@
         true ? static_cast<void>(0) : static_cast<void>((expression));                                                 \
     }
 
+// NOLINTBEGIN
+
 #if NDEBUG
 #define ASSERT(expression) DONT_EVALUATE((expression))
 #else
 #define ASSERT(expression) assert((expression))
 #endif // NDEBUG
+
+// NOLINTEND
--- a/cpp/src/aztec/honk/circuit_constructors/circuit_constructor_base.hpp
+++ b/cpp/src/aztec/honk/circuit_constructors/circuit_constructor_base.hpp
@@ -1,5 +1,6 @@
 #pragma once
 #include <ecc/curves/bn254/fr.hpp>
+#include <utility>
 
 namespace honk {
 static constexpr uint32_t DUMMY_TAG = 0;
@@ -98,55 +99,60 @@ template <size_t program_width_> class CircuitConstructorBase {
   public:
     static constexpr size_t program_width = program_width_;
     std::vector<std::string> selector_names_;
-    size_t n; // the circuit size; we should rename
+    size_t num_gates = 0;
+    // TODO(Adrian): It would be better to store an array of size program_width_
+    // to make the composer agnostic of the wire name.
     std::vector<uint32_t> w_l;
     std::vector<uint32_t> w_r;
     std::vector<uint32_t> w_o;
     std::vector<uint32_t> w_4;
     std::vector<uint32_t> public_inputs;
     std::vector<barretenberg::fr> variables;
-    std::vector<uint32_t> next_var_index; // index of next variable in equivalence class (=REAL_VARIABLE if you're last)
-    std::vector<uint32_t>
-        prev_var_index; // index of  previous variable in equivalence class (=FIRST if you're in a cycle alone)
-    std::vector<uint32_t> real_variable_index; // indices of corresponding real variables
+    // index of next variable in equivalence class (=REAL_VARIABLE if you're last)
+    std::vector<uint32_t> next_var_index;
+    // index of  previous variable in equivalence class (=FIRST if you're in a cycle alone)
+    std::vector<uint32_t> prev_var_index;
+    // indices of corresponding real variables
+    std::vector<uint32_t> real_variable_index;
     std::vector<uint32_t> real_variable_tags;
     uint32_t current_tag = DUMMY_TAG;
-    std::map<uint32_t, uint32_t>
-        tau; // The permutation on variable tags. See
-             // https://github.com/AztecProtocol/plonk-with-lookups-private/blob/new-stuff/GenPermuations.pdf
-             // DOCTODO: replace with the relevant wiki link.
+    // The permutation on variable tags. See
+    // https://github.com/AztecProtocol/plonk-with-lookups-private/blob/new-stuff/GenPermuations.pdf
+    // DOCTODO: replace with the relevant wiki link.
+    std::map<uint32_t, uint32_t> tau;
 
     size_t num_selectors;
     std::vector<std::vector<barretenberg::fr>> selectors;
-    numeric::random::Engine* rand_engine;
+    numeric::random::Engine* rand_engine = nullptr;
     bool _failed = false;
     std::string _err;
     static constexpr uint32_t REAL_VARIABLE = UINT32_MAX - 1;
     static constexpr uint32_t FIRST_VARIABLE_IN_CLASS = UINT32_MAX - 2;
-    static constexpr size_t NUM_RESERVED_GATES = 4; // this must be >= num_roots_cut_out_of_vanishing_polynomial
 
     // Enum values spaced in increments of 30-bits (multiples of 2 ** 30).
-    enum WireType { LEFT = 0U, RIGHT = (1U << 30U), OUTPUT = (1U << 31U), FOURTH = 0xc0000000 };
+    // TODO(Adrian): This is unused, and this type of hard coded data should be avoided
+    // enum WireType { LEFT = 0U, RIGHT = (1U << 30U), OUTPUT = (1U << 31U), FOURTH = 0xc0000000 };
 
     CircuitConstructorBase(std::vector<std::string> selector_names, size_t num_selectors = 0, size_t size_hint = 0)
-        : selector_names_(selector_names)
-        , n(0)
+        : selector_names_(std::move(selector_names))
         , num_selectors(num_selectors)
         , selectors(num_selectors)
-        , rand_engine(nullptr)
     {
         for (auto& p : selectors) {
             p.reserve(size_hint);
         }
     }
 
-    CircuitConstructorBase(CircuitConstructorBase&& other) = default;
-    CircuitConstructorBase& operator=(CircuitConstructorBase&& other) = default;
-    virtual ~CircuitConstructorBase(){};
+    CircuitConstructorBase(const CircuitConstructorBase& other) = delete;
+    CircuitConstructorBase(CircuitConstructorBase&& other) noexcept = default;
+    CircuitConstructorBase& operator=(const CircuitConstructorBase& other) = delete;
+    CircuitConstructorBase& operator=(CircuitConstructorBase&& other) noexcept = default;
+    virtual ~CircuitConstructorBase() = default;
 
-    virtual size_t get_num_gates() const { return n; }
-    virtual void print_num_gates() const { std::cout << n << std::endl; }
+    virtual size_t get_num_gates() const { return num_gates; }
+    virtual void print_num_gates() const { std::cout << num_gates << std::endl; }
     virtual size_t get_num_variables() const { return variables.size(); }
+    // TODO(Adrian): Feels wrong to let the zero_idx be changed.
     uint32_t zero_idx = 0;
 
     virtual void create_add_gate(const add_triple& in) = 0;
@@ -264,24 +270,24 @@ template <size_t program_width_> class CircuitConstructorBase {
      * */
     virtual void set_public_input(const uint32_t witness_index)
     {
-        bool does_not_exist = true;
-        for (size_t i = 0; i < public_inputs.size(); ++i) {
-            does_not_exist = does_not_exist && (public_inputs[i] != witness_index);
+        for (const uint32_t public_input : public_inputs) {
+            if (public_input == witness_index) {
+                if (!failed()) {
+                    failure("Attempted to set a public input that is already public!");
+                }
+                return;
+            }
         }
-        if (does_not_exist) {
-            public_inputs.emplace_back(witness_index);
-        }
-        ASSERT(does_not_exist);
-        if (!does_not_exist && !failed()) {
-            failure("Attempted to set a public input that is already public!");
-        }
+        public_inputs.emplace_back(witness_index);
     }
 
     virtual void assert_equal(const uint32_t a_idx, const uint32_t b_idx, std::string const& msg = "assert_equal");
 
-    size_t get_circuit_subgroup_size(const size_t num_gates)
+    // TODO(Adrian): This method should belong in the ComposerHelper, where the number of reserved gates can be
+    // correctly set
+    size_t get_circuit_subgroup_size(const size_t num_gates) const
     {
-        size_t log2_n = static_cast<size_t>(numeric::get_msb(num_gates));
+        auto log2_n = static_cast<size_t>(numeric::get_msb(num_gates));
         if ((1UL << log2_n) != (num_gates)) {
             ++log2_n;
         }
@@ -299,19 +305,16 @@ template <size_t program_width_> class CircuitConstructorBase {
     // uint32::MAX number of variables
     void assert_valid_variables(const std::vector<uint32_t>& variable_indices)
     {
-        for (size_t i = 0; i < variable_indices.size(); i++) {
-            ASSERT(is_valid_variable(variable_indices[i]));
+        for (const auto& variable_index : variable_indices) {
+            ASSERT(is_valid_variable(variable_index));
         }
     }
-    bool is_valid_variable(uint32_t variable_index)
-    {
-        return static_cast<uint32_t>(variables.size()) > variable_index;
-    };
+    bool is_valid_variable(uint32_t variable_index) { return variable_index < variables.size(); };
 
     bool failed() const { return _failed; };
     const std::string& err() const { return _err; };
 
-    void set_err(std::string msg) { _err = msg; }
+    void set_err(std::string msg) { _err = std::move(msg); }
     void failure(std::string msg)
     {
         _failed = true;
--- a/cpp/src/aztec/honk/circuit_constructors/standard_circuit_constructor.hpp
+++ b/cpp/src/aztec/honk/circuit_constructors/standard_circuit_constructor.hpp
@@ -19,6 +19,7 @@ class StandardCircuitConstructor : public CircuitConstructorBase<STANDARD_HONK_W
 
     // These are variables that we have used a gate on, to enforce that they are
     // equal to a defined value.
+    // TODO(Adrian): Why is this not in CircuitConstructorBase
     std::map<barretenberg::fr, uint32_t> constant_variable_indices;
 
     StandardCircuitConstructor(const size_t size_hint = 0)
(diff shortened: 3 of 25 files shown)


cpp/src/aztec/common/assert.hpp
```diff
@@ -10,8 +10,12 @@
...
 
+// NOLINTBEGIN
+
 #if NDEBUG
 #define ASSERT(expression) DONT_EVALUATE((expression))
 #else
 #define ASSERT(expression) assert((expression))
 #endif // NDEBUG
+
+// NOLINTEND
```
adr1anh, line 21: Prevents a warning with `clang-tidy`

cpp/src/aztec/honk/circuit_constructors/circuit_constructor_base.hpp
```diff
@@ -264,24 +270,24 @@ template <size_t program_width_> class CircuitConstructorBase {
...
     virtual void set_public_input(const uint32_t witness_index)
     {
-        bool does_not_exist = true;
-        for (size_t i = 0; i < public_inputs.size(); ++i) {
-            does_not_exist = does_not_exist && (public_inputs[i] != witness_index);
-        }
-        if (does_not_exist) {
-            public_inputs.emplace_back(witness_index);
-        }
-        ASSERT(does_not_exist);
```
adr1anh, line 274: Not sure if this `ASSERT` is useful, since if it is triggered, the remaining composer failure is not triggered.

cpp/src/aztec/honk/circuit_constructors/circuit_constructor_base.hpp
```diff
@@ -98,55 +99,60 @@ template <size_t program_width_> class CircuitConstructorBase {
...
     }
 
-    CircuitConstructorBase(CircuitConstructorBase&& other) = default;
-    CircuitConstructorBase& operator=(CircuitConstructorBase&& other) = default;
-    virtual ~CircuitConstructorBase(){};
+    CircuitConstructorBase(const CircuitConstructorBase& other) = delete;
+    CircuitConstructorBase(CircuitConstructorBase&& other) noexcept = default;
+    CircuitConstructorBase& operator=(const CircuitConstructorBase& other) = delete;
+    CircuitConstructorBase& operator=(CircuitConstructorBase&& other) noexcept = default;
+    virtual ~CircuitConstructorBase() = default;
```
adr1anh, line 150: The make `clang-tidy` happy

cpp/src/aztec/honk/circuit_constructors/circuit_constructor_base.hpp
```diff
@@ -98,55 +99,60 @@ template <size_t program_width_> class CircuitConstructorBase {
...
     bool _failed = false;
     std::string _err;
     static constexpr uint32_t REAL_VARIABLE = UINT32_MAX - 1;
     static constexpr uint32_t FIRST_VARIABLE_IN_CLASS = UINT32_MAX - 2;
-    static constexpr size_t NUM_RESERVED_GATES = 4; // this must be >= num_roots_cut_out_of_vanishing_polynomial
 
     // Enum values spaced in increments of 30-bits (multiples of 2 ** 30).
-    enum WireType { LEFT = 0U, RIGHT = (1U << 30U), OUTPUT = (1U << 31U), FOURTH = 0xc0000000 };
+    // TODO(Adrian): This is unused, and this type of hard coded data should be avoided
+    // enum WireType { LEFT = 0U, RIGHT = (1U << 30U), OUTPUT = (1U << 31U), FOURTH = 0xc0000000 };
```
adr1anh, line 134: This was unused, and if necessary we can add it back later. `NUM_RESERVED_GATES` should be defined by the `ComposerHelper` anyway

cpp/src/aztec/honk/circuit_constructors/circuit_constructor_base.hpp
```diff
@@ -299,19 +305,16 @@ template <size_t program_width_> class CircuitConstructorBase {
...
-            ASSERT(is_valid_variable(variable_indices[i]));
+        for (const auto& variable_index : variable_indices) {
+            ASSERT(is_valid_variable(variable_index));
         }
     }
-    bool is_valid_variable(uint32_t variable_index)
-    {
-        return static_cast<uint32_t>(variables.size()) > variable_index;
-    };
+    bool is_valid_variable(uint32_t variable_index) { return variable_index < variables.size(); };
```
adr1anh, line 312: The `static_cast` was not necessary since the comparison operator will promote `variable_index` to the same width as `variables.size()`.

cpp/src/aztec/honk/circuit_constructors/standard_circuit_constructor.hpp
```diff
@@ -27,12 +28,15 @@ class StandardCircuitConstructor : public CircuitConstructorBase<STANDARD_HONK_W
...
+        // TODO(Adrian): This should be done in a constant way, maybe by initializing the constant_variable_indices map
         zero_idx = put_constant_variable(barretenberg::fr::zero());
     };
 
+    StandardCircuitConstructor(const StandardCircuitConstructor& other) = delete;
     StandardCircuitConstructor(StandardCircuitConstructor&& other) = default;
+    StandardCircuitConstructor& operator=(const StandardCircuitConstructor& other) = delete;
     StandardCircuitConstructor& operator=(StandardCircuitConstructor&& other) = default;
-    ~StandardCircuitConstructor() {}
+    ~StandardCircuitConstructor() override = default;
```
adr1anh, line 41: Makes `clang-tidy` happy.

cpp/src/aztec/honk/composer/composer_helper/composer_helper.hpp
```diff
@@ -23,53 +26,60 @@ template <typename CircuitConstructor> class ComposerHelper {
...
     ComposerHelper(std::unique_ptr<waffle::ReferenceStringFactory>&& crs_factory)
         : crs_factory_(std::move(crs_factory))
     {}
-    ComposerHelper(std::shared_ptr<waffle::proving_key> const& p_key,
-                   std::shared_ptr<waffle::verification_key> const& v_key)
-        : circuit_proving_key(p_key)
-        , circuit_verification_key(v_key)
+    ComposerHelper(std::shared_ptr<waffle::proving_key> p_key, std::shared_ptr<waffle::verification_key> v_key)
+        : circuit_proving_key(std::move(p_key))
+        , circuit_verification_key(std::move(v_key))
```
adr1anh, line 38: Makes `clang-tidy` happy

cpp/src/aztec/honk/composer/composer_helper/permutation_helper.hpp
```diff
@@ -1,118 +1,89 @@
...
+#include <cstdint>
+#include <initializer_list>
+#include <cstdint>
+#include <cstddef>
+#include <utility>
+#include <vector>
+
 namespace honk {
-// Enum values spaced in increments of 30-bits (multiples of 2 ** 30).
-enum WireType { LEFT = 0U, RIGHT = (1U << 30U), OUTPUT = (1U << 31U), FOURTH = 0xc0000000 };
```
adr1anh, line 7: This was not used directly, and made things a lot more confusing. Instead, we use a `uint32_t` to describe which column we are using explicitly.

cpp/src/aztec/honk/composer/composer_helper/permutation_helper.hpp
```diff
@@ -1,118 +1,89 @@
...
-        wire_type = other.wire_type;
-        return *this;
-    }
-    bool operator==(const cycle_node& other) const
-    {
-        return ((gate_index == other.gate_index) && (wire_type == other.wire_type));
-    }
 };
-typedef std::vector<std::vector<cycle_node>> CycleCollector;
+using CyclicPermutation = std::vector<cycle_node>;
```
adr1anh: All these constructors could be removed, making `cycle_node` a POD (plain-old data). Also changed the typedef to be more explicit.

cpp/src/aztec/honk/composer/composer_helper/permutation_helper.hpp
```diff
@@ -1,118 +1,89 @@
...
- * Then go through all witnesses in w_l, w_r, w_o and w_4 (if program width is > 3) and
- * add them to cycles of their real indexes.
+ * Compute all CyclicPermutations of the circuit. Each CyclicPermutation represents the indices of the values in the
+ * witness wires that must have the same value.
  *
  * @tparam program_width Program width
  * */
 template <size_t program_width, typename CircuitConstructor>
-void compute_wire_copy_cycles(CircuitConstructor& circuit_constructor, CycleCollector& wire_copy_cycles)
+std::vector<CyclicPermutation> compute_wire_copy_cycles(const CircuitConstructor& circuit_constructor)
```
adr1anh, line 37: Returning a value is cheap thanks to RVO (return value optimization) and makes the control flow clearer.

cpp/src/aztec/honk/composer/composer_helper/permutation_helper.hpp
```diff
@@ -1,118 +1,89 @@
...
+    }
 
-        const auto public_input_index = real_variable_index[public_inputs[i]];
-        if (static_cast<size_t>(public_input_index) >= number_of_cycles) {
-            wire_copy_cycles.resize(public_input_index + 1);
-        }
-        std::vector<cycle_node>& cycle = wire_copy_cycles[static_cast<size_t>(public_input_index)];
+    // Each variable represents one cycle
+    const size_t number_of_cycles = circuit_constructor.variables.size();
+    std::vector<CyclicPermutation> copy_cycles(number_of_cycles);
```
adr1anh, line 56: Initializes the `copy_cycles` to be the same size as `variables` means that there is a 1-to-1 correspondence between the two.

cpp/src/aztec/honk/composer/standard_honk_composer.test.cpp
```diff
@@ -19,90 +20,108 @@ namespace test_standard_honk_composer {
  */
 TEST(standard_honk_composer, test_sigma_and_id_correctness)
 {
+    auto test_permutation = [](StandardHonkComposer& composer) {
```
adr1anh, line 26: This was put into a lambda so it would be possible to test different circuit configurations, but so far we still use the same configuration as before (with a public variable)

cpp/src/aztec/honk/proof_system/prover.cpp
```diff
@@ -91,13 +91,12 @@ template <typename settings> void Prover<settings>::compute_wire_commitments()
...
  */
-template <typename settings> void Prover<settings>::compute_grand_product_polynomial(Fr beta)
+template <typename settings>
+void Prover<settings>::compute_grand_product_polynomial(barretenberg::fr beta, barretenberg::fr gamma)
 {
     using barretenberg::polynomial_arithmetic::copy_polynomial;
     static const size_t program_width = settings::program_width;
 
-    Fr gamma = beta * beta; // TODO(Cody): We already do this and it's kosher, right?
-
```
adr1anh, line 100: This was not kosher, so `gamma` was put back.

cpp/src/aztec/honk/proof_system/prover.hpp
```diff
@@ -28,23 +28,13 @@ template <typename settings> class Prover {
...
 
-    void flush_queued_work_items() { queue.flush_queue(); }
-
-    waffle::work_queue::work_item_info get_queued_work_item_info() const { return queue.get_queued_work_item_info(); }
-
-    size_t get_scalar_multiplication_size(const size_t work_item_number) const
-    {
-        return queue.get_scalar_multiplication_size(work_item_number);
-    }
-
```
adr1anh, line 47: Commented this out and placed it along with the other `work_queue` related declarations.

cpp/src/aztec/honk/proof_system/verifier.test.cpp
```diff
@@ -114,11 +114,11 @@ template <class FF> class VerifierTests : public testing::Test {
...
-            fr::__copy(fr::one(), q_l.at(2 * i));
-            fr::__copy(fr::one(), q_r.at(2 * i));
-            fr::__copy(fr::neg_one(), q_o.at(2 * i));
-            fr::__copy(fr::one(), q_c.at(2 * i));
-            fr::__copy(fr::one(), q_m.at(2 * i));
+            q_l.at(2 * i) = fr::one();
+            q_r.at(2 * i) = fr::one();
+            q_o.at(2 * i) = fr::neg_one();
+            q_c.at(2 * i) = fr::one();
+            q_m.at(2 * i) = fr::one();
```
adr1anh, line 117: There is no reason the compiler cannot optimize the default assignment to be as performant as `__copy`. This makes things clearer

adr1anh (commented):
Explanations for some changes

cpp/src/aztec/honk/composer/composer_helper/composer_helper.cpp
```diff
@@ -139,69 +135,53 @@ std::shared_ptr<waffle::verification_key> ComposerHelper<CircuitConstructor>::co
...
-    auto zero_idx = circuit_constructor.zero_idx;
-
-    const size_t total_num_gates = std::max(minimum_circuit_size, n + public_inputs.size());
-    const size_t subgroup_size = circuit_constructor.get_circuit_subgroup_size(total_num_gates + NUM_RESERVED_GATES);
+    const size_t num_gates = circuit_constructor.num_gates;
+    std::span<const uint32_t> public_inputs = circuit_constructor.public_inputs;
+    const size_t num_public_inputs = public_inputs.size();
+
+    const size_t num_constraints = std::max(minimum_circuit_size, num_gates + num_public_inputs);
+    // TODO(Adrian): We should o
```
Rumata888: What?

cpp/src/aztec/honk/composer/composer_helper/permutation_helper.hpp
```diff
@@ -1,118 +1,89 @@
...
-// Enum values spaced in increments of 30-bits (multiples of 2 ** 30).
-enum WireType { LEFT = 0U, RIGHT = (1U << 30U), OUTPUT = (1U << 31U), FOURTH = 0xc0000000 };
 
 /**
- * @brief cycle_node represents a particular witness at a particular gate. Used to collect permutation sets
- *
+ * @brief cycle_node represents the index of a value of the circuit.
+ * It will belong to a CyclicPermutation, such that all nodes in a CyclicPermutation
+ * must have the value.
+ * The total number of constaints is always <2^32 since that is the type used to represent variables, so we can save
```
Rumata888: *constraints

cpp/src/aztec/honk/composer/composer_helper/permutation_helper.hpp
```diff
@@ -1,118 +1,89 @@
...
+    // Represents the index of a variable in circuit_constructor.variables
+    std::span<const uint32_t> real_variable_index = circuit_constructor.real_variable_index;
+
+    // We use the permutation argument to enforce the public input variables to be equal to values provided by the
+    // verifier. The convension we use is to place the public input values as the first rows of witness vectors.
+    // All selectors are zero at these rows, so they are fully unconstrained.
+    // The "real" gates that follow can use references to these variables.
+    //
+    // The copy cycle for the i-th public variable looks like
+    //   (i) -> (n+i) -> (i') -> ... -> (i'')
```
Rumata888, line 68: It's not intuitive, because the schemes in HONK and Plonk differ, so maybe it's best to rephrase "(i) -> (n+i)". Probably better to explain that the permutation uses the left and right wires at index i. Also you have the comment a bit lower unfinished

Rumata888 (approved):
Everything seems great

cpp/src/aztec/honk/composer/composer_helper/permutation_helper.hpp
adr1anh: Left a bit more context with regards to the Left Right wires. And finished the comments below.

cpp/src/aztec/honk/composer/composer_helper/composer_helper.cpp
adr1anh: Expanded the comment.

PR #63, opened 2023-01-16
Is/verification key construction

Rumata888:
Implemented verification key computation
https://github.com/AztecProtocol/aztec2-internal/issues/1818


codygunton:
This became #68.

PR #65, opened 2023-01-17
Cg/PoC partial verification

codygunton:
Working toward a full proof construction and verification flow using a partial proof system.


codygunton:
Some fixes:

- `waffle::Verifier` $\leadsto$ `honk::StandardUnrolledVerifier`
- Sumcheck evals are added to transcript from within sumcheck module (old method lead to all zeros in transcript)
- Verifier algorithm adds `u` challenges to transcript.
- Start on verifier-computed polynomial values.
- Temporarily end prover and verifier at sumcheck.
- Verifier test uses real manifest.
- Circuit size is correctly extracted from transcript into `Multivariates`.
- Multivariate evals recorded as a single element in the manifest.

codygunton:
Subsumed by #72.

PR #68, opened 2023-01-18
Constructing verification key

Rumata888:
Adds the creation of verification key (basically, commitments to selectors and permutation polynomials)


cpp/src/aztec/honk/composer/standard_honk_composer.test.cpp
```diff
@@ -257,6 +257,27 @@ TEST(standard_honk_composer, test_assert_equal)
     EXPECT_EQ(get_maximum_cycle(composer_with_assert_equal), get_maximum_cycle(composer_no_assert_equal) * 2);
 }
 
+TEST(standard_honk_composer, test_verification_key_creation)
```
ledwards2225, line 260: Might be worth adding a few more comments to this test. Took me a second to determine what it's actually checking

ledwards2225 (approved):

PR #87, opened 2023-01-27
[REPLACED] Making sumcheck work with challenges

Rumata888:
# REPLACED BY PR https://github.com/AztecProtocol/barretenberg/pull/102

### DIFF
--- a/cpp/src/aztec/honk/proof_system/prover.cpp
+++ b/cpp/src/aztec/honk/proof_system/prover.cpp
@@ -295,7 +295,7 @@ template <typename settings> void Prover<settings>::execute_relation_check_round
     using Sumcheck = sumcheck::Sumcheck<Multivariates,
                                         Transcript,
                                         sumcheck::ArithmeticRelation,
-                                        // sumcheck::GrandProductComputationRelation,
+                                        sumcheck::GrandProductComputationRelation,
                                         sumcheck::GrandProductInitializationRelation>;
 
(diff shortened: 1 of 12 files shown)


ledwards2225:
All four Sumcheck tests were failing due to a mock manifest/transcript issue (and in some cases some other bad inputs) Commit [e8d4ad9](https://github.com/AztecProtocol/barretenberg/pull/87/commits/e8d4ad9d4d48641e7cacf7e4d4205189b5ecd130) fixes the manifest/transcript issue and makes all tests pass. To make `ProverAndVerifier` pass I had to make the inputs a bit more trivial since they were not set up to pass with all three relations. (Not sure you can make an `n=2` case that will pass with interesting inputs). The test `ProverAndVerifierLonger` has slightly more interesting multivariates but I did see failure when I tried to set `num_public_inputs > 0` while also setting the lagrange polys correctly. Could be an easy fix but I'm not fully up to date on how the PI stuff works. We should make sure that we're testing that functionality somewhere if not in that test. (To be clear, public inputs were not relevant to this test before this update so we're not losing something we were previously testing).

ledwards2225 (approved):
Looks good overall. Talked with Kesha via slack about possibly adding a comment documenting the funny looking empty template arg in the relations that dont require `relation_parameters`. To me it looks like it could be a typo or something otherwise. Could update the Sumcheck tests with some more interesting inputs but that could also wait until later.

PR #24, opened 2023-01-04
[REPLACED] Ah/byte-array-test

adr1anh:
# REPLACED BY PR https://github.com/AztecProtocol/barretenberg/pull/91

PR #91, opened 2023-01-28
Ah/byte-array-test

### LINKED ISSUE #656: Honk proofs are structs(?)
This seems much better--you can read what's in the proof and you can explicitly see how thing serialize because there's a function that shows the serialization.

Luke: Also, this would allow for more explicit testing of invalid proofs in the recursive verification setting. E.g. this would make it easy to supply an incorrect multilinear evaluation or an bad wire commitment. See the honk recursive verifier tests for an example.

dbanks12:
# **Copied from PR https://github.com/AztecProtocol/barretenberg/pull/24**
   - **Original author: @adr1anh**

---

# Description

COPEID from https://github.com/AztecProtocol/aztec2-internal/pull/1418

Modifies `byte_array` tests to work over all composers, and fixes #656. 

This PR implements a different strategy for typed tests, and includes a new header [stdlib/testing/testing.hpp](https://github.com/AztecProtocol/aztec2-internal/blob/c98438346f1ff31f6978f93724e1ee01ff305246/barretenberg/src/aztec/stdlib/testing/testing.hpp) that should provide some remove some boiler plate code required with typed tests. One disadvantage is that type aliases for the given composer cannot be used within the `TYPED_TEST` block, and requires additional `using` directives. I'd be interested in hearing folk's feedback on this approach compared to defining tests as static methods in the fixture class. 

When implementing a new test to check if byte arrays for smaller `safe_uint` were being constructed correctly, I noticed that range constraints would fail when `num_bits == 0`, due to a missing `else if`. This was fixed and the corresponding test for `stdlib::field` was added to check for this. I also improved the speed of the test by reusing the same composer for all successful tests, and stopping verification earlier for failing tests by checking that the composer did not give an error. 

This PR also modifies one the `safe_uint` constructors that accepted a `stdlib::bool` to **only** accept bools and not other types, which would prevent accidental conversions from `witness_t`. 

# Checklist:

## Barretenberg-specific requirements:

---


codygunton:
It would be nice to have something like this, but some of it will become moot when the splitting of the composers is complete, and for other parts we have taken a more modest approach as in #325.

PR #93, opened 2023-01-28
SRS Transcript Cleanup

dbanks12:
# **Copied from PR https://github.com/AztecProtocol/barretenberg/pull/36**
   - **Original author: @suyash67**

---

This PR slightly modifies the way we handle the SRS data in barretenberg. 
1. First of all, the directory structure of the transcripts changes slightly:
    - On running `./bootstrap.sh` in `cpp/`, the `ignition` directory would create the `monomial` and the `lagrange` directories which will store the respective transcripts downloaded from our s3 bucket.
    - Monomial transcripts are downloaded from https://aztec-ignition.s3.eu-west-2.amazonaws.com/MAIN+IGNITION/monomial
    - Lagrange transcripts are downloaded from https://aztec-ignition.s3.eu-west-2.amazonaws.com/MAIN+IGNITION/lagrange
2. The first monomial transcript `ignition/monomial/transcript00.dat` now contains 5040001 points and rest of the 19 monomial transcripts contain 5040000 points. The generator $[1]_1$ was added to `transcript00.dat`. This allows us to remove conditional logic on including $[1]_1$ while reading monomial transcripts.
3. The code to generate a Lagrange transcript from a monomial transcript now also works for size $>2^{24}$. To store Lagrange transcripts with size $>2^{24}$, we split it into multiple files each containing $2^{24}$ points (file size is 1.1GB). For example, we now store the Lagrange transcript of size $2^{25}$ in $\frac{2^{25}}{2^{24}}=2$ files named `transcript_ 33554432.dat` and `transcript_33554432_1.dat`. This was done because it isn't possible to write more than 1.1GB of data to a .dat file in our code. 
---


cpp/src/aztec/lagrange_base_gen/lagrange_base_processor.cpp
```diff
@@ -44,10 +44,35 @@ int main(int argc, char** argv)
     g2_elements.push_back(verifier_ref_string->get_g2x());
     g2_elements.push_back(barretenberg::g2::affine_one);
 
-    barretenberg::io::Manifest manifest{
-        0, 1, static_cast<uint32_t>(subgroup_size), 2, static_cast<uint32_t>(subgroup_size), 2, 0
-    };
-    barretenberg::io::write_transcript(&lagrange_base_srs[0], &g2_elements[0], manifest, lagrange_srs_path, true);
+    const uint32_t subgroup_size_ = static_cast<uint32_t>(subgroup_size);
+    const uint32_t max_lagrange_transcript_size = static_cast<uint32_t>((size_t(1) << 24));
+    if (subgroup_size > max_lagrange_transcript_size) {
```
ledwards2225: Do we really need the if-else here? Seems like the code within the if could handle both cases (with some minor modifications perhaps). Maybe I'm missing a subtlety. Not a big deal either way

ledwards2225 (approved):
Looks good to me!

cpp/src/aztec/lagrange_base_gen/lagrange_base_processor.cpp
suyash67: Good point, simplified it. Thank you!

PR #94, opened 2023-01-28
Commit to Wires Using Lagrange SRS

dbanks12:
# **Copied from PR https://github.com/AztecProtocol/barretenberg/pull/39**
   - **Original author: @suyash67**

---

Committing to the wire polynomials with Lagrange srs is expected to improve the proof construction time. This is because the wire values are very small in magnitude as compared to the field size, and the resulting MSM while computing the commitments to the wire polynomials is faster.

This is going to be very useful if/when we start using the new lookup argument for doing permutation check ([paper](https://eprint.iacr.org/2022/1530.pdf)).
---

# Comments from original PR

## **Copied from comment (https://github.com/AztecProtocol/barretenberg/pull/39#issuecomment-1375487455) by author @suyash67**

Note: we need to merge #36 before this one.

### DIFF
--- a/cpp/src/aztec/plonk/proof_system/prover/prover.cpp
+++ b/cpp/src/aztec/plonk/proof_system/prover/prover.cpp
@@ -67,6 +67,7 @@ template <typename settings> ProverBase<settings>& ProverBase<settings>::operato
 /**
  * - Compute wire commitments and add them to the transcript.
  * - Add public_inputs from w_2_fft to transcript.
+ * - We use Lagrange srs to commit to wire polynomials (in their Lagrange form).
  *
  * @tparam settings Program settings.
  * */
@@ -77,9 +78,9 @@ template <typename settings> void ProverBase<settings>::compute_wire_commitments
     for (size_t i = 0; i < end; ++i) {
         std::string wire_tag = "w_" + std::to_string(i + 1);
         std::string commit_tag = "W_" + std::to_string(i + 1);
-        barretenberg::fr* coefficients = key->polynomial_cache.get(wire_tag).get_coefficients();
+        barretenberg::fr* coefficients = key->polynomial_cache.get(wire_tag + "_lagrange").get_coefficients();
         // This automatically saves the computed point to the transcript
-        commitment_scheme->commit(coefficients, commit_tag, work_queue::MSMSize::N, queue);
+        commitment_scheme->commit(coefficients, commit_tag, work_queue::MSMType::LAGRANGE_N, queue);
     }
 
     // add public inputs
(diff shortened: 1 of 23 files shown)


cpp/src/aztec/plonk/proof_system/prover/prover.cpp
```diff
@@ -251,6 +243,20 @@ template <typename settings> void ProverBase<settings>::execute_first_round()
     start = std::chrono::steady_clock::now();
 #endif
     compute_wire_commitments();
+
```
ledwards2225, line 246: Thank you for moving this!

cpp/src/aztec/plonk/proof_system/prover/prover.cpp
```diff
@@ -285,30 +291,31 @@ template <typename settings> void ProverBase<settings>::execute_second_round()
...
             .work_type = work_queue::WorkType::SCALAR_MULTIPLICATION,
-            .mul_scalars = key->polynomial_cache.get(wire_tag).get_coefficients(),
+            .mul_scalars = w_4_lagrange.get_coefficients(),
             .tag = "W_4",
-            .constant = barretenberg::fr(0),
+            .constant = work_queue::MSMType::LAGRANGE_N,
+            .index = 0,
+        });
+
+        // convert w_4 to the coefficient form.
```
ledwards2225, line 313: Was wondering why this was not previously done in the queue. Just an oversight I guess?

cpp/src/aztec/proof_system/work_queue/work_queue.cpp
```diff
@@ -50,14 +50,28 @@ size_t work_queue::get_scalar_multiplication_size(const size_t work_item_number)
...
-                return item.constant == MSMSize::N ? key->circuit_size : key->circuit_size + 1;
+                return (item.constant == MSMType::MONOMIAL_N_PLUS_ONE) ? key->circuit_size + 1 : key->circuit_size;
             }
             ++count;
         }
     }
     return 0;
 }
 
+bool work_queue::get_scalar_multiplication_type(const size_t work_item_number) const
```
ledwards2225, line 66: A comment here (or in header) would be helpful since since it's not immediately obvious which return value corresponds to which SM type

ledwards2225 (approved):
Left a few comments/suggestions but overall this looks great so I'll go ahead and approve and let you do what you like with the comments.

cpp/src/aztec/plonk/proof_system/prover/prover.cpp
suyash67: I believe it wasn't an oversight. Previously we were committing to $w_4$ using the monomial SRS. So what we did is: 
1. $w_{4, \textsf{mono}} = \textsf{iFFT}(w_{4, \textsf{lag}})$  // this happens right away
2. $W_4 = \textsf{commit}(w_{4, \textsf{mono}})$  // this is pushed to the queue and is processed after the round
These two steps happens sequentially. So step 1 can be computed "locally" without having to wait for the queue processing at the end of this round. We could do the $\textsf{iFFT}$ operation on the queue but that would only work iff the iFFT and MSM are executed sequentially in WASM (so our typescript code would need to have this knowledge). 

Now, when we want to commit using the Lagrange SRS, we need to two operations:
1. $w_{4, \textsf{mono}} = \textsf{iFFT}(w_{4, \textsf{lag}})$  
2. $W_4 = \textsf{commit}(w_{4, \textsf{lag}})$ 
These two operations are computing the output on the same input, i.e. $w_{4, \textsf{lag}}$. If we did the iFFT "locally" and not on queue, that would change the polynomial $w_4$ to monomial form, and hence the MSM operation would give a wrong result (Lagrange SRS with coefficients of $w_4$). Therefore, in this case, there is no choice but to use the queue to compute iFFT. One downside of this though is: we will have to ensure that the MSM operation is processed before the iFFT operation in WASM (because iFFT modifies the same memory).

cpp/src/aztec/proof_system/work_queue/work_queue.cpp
suyash67: Done, thank you!

PR #60, opened 2023-01-12
[REPLACED] Skip some recursion tests.

codygunton:
# REPLACED BY PR https://github.com/AztecProtocol/barretenberg/pull/96

PR #96, opened 2023-01-28
Skip some recursion tests.

dbanks12:
# **Copied from PR https://github.com/AztecProtocol/barretenberg/pull/60**
   - **Original author: @codygunton**

---

The set of stdlib recursion tests seems to be overkill, given that we are going to significantly change recursion before it's ever put into production again. Reducing that set of tests to speed up CI time.
---


codygunton:
This may become unnecessary after the composers are finally split, at which point we will start to use `check_circuit` for most stdlib tests. We also have a totally new kind of recursion coming in with cycle of curves, and testing this will look very different. We are also about to start running tests from the Noir repo which will blow up CI time to > 15 mins, most likely. Finally, to really get CI time down we would have to also change how we test biggroup. Given all of this, I'm going to close this PR with the understanding that we will address the underlying issue in an appropriate manner in the future.

PR #103, opened 2023-01-28
Cg/pow zeta

dbanks12:
# **Copied from PR https://github.com/AztecProtocol/barretenberg/pull/88**
   - **Original author: @codygunton**

---

# Description

Inefficiently introduce $pow_\zeta$ into Honk PoC.

This PR mistakenly includes some comments that were intended to go into another PR, but since they are just comments, I think it's fine to leave them in here.

# Checklist:

---

### DIFF
--- a/cpp/src/aztec/honk/composer/composer_helper/composer_helper.cpp
+++ b/cpp/src/aztec/honk/composer/composer_helper/composer_helper.cpp
@@ -251,6 +251,7 @@ std::shared_ptr<waffle::verification_key> ComposerHelper<CircuitConstructor>::co
  *
  * @return The verifier.
  * */
+// TODO(Cody): This should go away altogether.
 template <typename CircuitConstructor>
 StandardVerifier ComposerHelper<CircuitConstructor>::create_verifier(const CircuitConstructor& circuit_constructor)
 {
--- a/cpp/src/aztec/honk/composer/composer_helper/composer_helper.cpp
+++ b/cpp/src/aztec/honk/composer/composer_helper/composer_helper.cpp
@@ -218,6 +218,8 @@ std::shared_ptr<waffle::proving_key> ComposerHelper<CircuitConstructor>::compute
 
     // TODO(Cody): this is a workaround
     circuit_proving_key->polynomial_cache.put("z_perm_lagrange", Polynomial<barretenberg::fr>(1));
+    // used in test_verification_key_creation
+    circuit_proving_key->polynomial_cache.put("pow_zeta", Polynomial<barretenberg::fr>(1));
 
     return circuit_proving_key;
 }
--- a/cpp/src/aztec/honk/composer/composer_helper/composer_helper.cpp
+++ b/cpp/src/aztec/honk/composer/composer_helper/composer_helper.cpp
@@ -216,11 +216,6 @@ std::shared_ptr<waffle::proving_key> ComposerHelper<CircuitConstructor>::compute
 
     compute_first_and_last_lagrange_polynomials(circuit_proving_key.get());
 
-    // TODO(Cody): this is a workaround
-    circuit_proving_key->polynomial_cache.put("z_perm_lagrange", Polynomial<barretenberg::fr>(1));
-    // used in test_verification_key_creation
-    circuit_proving_key->polynomial_cache.put("pow_zeta", Polynomial<barretenberg::fr>(1));
-
     return circuit_proving_key;
 }
 
--- a/cpp/src/aztec/honk/proof_system/prover.cpp
+++ b/cpp/src/aztec/honk/proof_system/prover.cpp
@@ -78,7 +78,7 @@ template <typename settings> void Prover<settings>::compute_wire_commitments()
  * Z_perm[i] = ∏ --------------------------------------------------------------------------------
  *                  (w_1(j) + β⋅σ_1(j) + γ) ⋅ (w_2(j) + β⋅σ_2(j) + γ) ⋅ (w_3(j) + β⋅σ_3(j) + γ)
  *
- * where ∏ := ∏_{j=0:i-1} and id_i(X) = id(X) + n*(i-1). These evaluations are constructed over the
+ * where ∏ := ∏_{j=0:i-1} and id_i(X) = id(X) + n*(i-1). These evaluations are constructed over t;he
  * course of four steps. For expositional simplicity, write Z_perm[i] as
  *
  *                A_1(j) ⋅ A_2(j) ⋅ A_3(j)
@@ -300,14 +300,18 @@ template <typename settings> void Prover<settings>::execute_relation_check_round
                                         sumcheck::GrandProductInitializationRelation>;
 
     // Compute alpha challenge
-    transcript.apply_fiat_shamir("alpha");
+    transcript.apply_fiat_shamir("zeta");
 
     // TODO(Cody): This is just temporary of course. Very inefficient, e.g., no commitment needed.
-    Fr zeta_challenge = transcript.get_challenge_field_element("alpha", 1);
+    Fr zeta_challenge = transcript.get_challenge_field_element("zeta");
     barretenberg::polynomial pow_zeta = power_polynomial::generate_vector(zeta_challenge, key->n);
-    auto commitment = commitment_key->commit(pow_zeta);
-    transcript.add_element("POW_ZETA", commitment.to_buffer());
     key->polynomial_cache.put("pow_zeta", std::move(pow_zeta));
+    std::span<Fr> pow_zeta_fetched = key->polynomial_cache.get("pow_zeta");
+    // The actual polynomial is of length n+1, but commitment key is just n, so we need to limit it
+    auto commitment = commitment_key->commit(pow_zeta_fetched);
+    transcript.add_element("POW_ZETA", commitment.to_buffer());
+
+    transcript.apply_fiat_shamir("alpha");
 
     auto multivariates = Multivariates(key);
     auto sumcheck = Sumcheck(multivariates, transcript);
--- a/cpp/src/aztec/honk/proof_system/prover.cpp
+++ b/cpp/src/aztec/honk/proof_system/prover.cpp
@@ -3,6 +3,7 @@
 #include <honk/sumcheck/sumcheck.hpp> // will need
 #include <array>
 #include <honk/sumcheck/polynomials/univariate.hpp> // will go away
+#include <honk/utils/power_polynomial.hpp>
 #include <honk/pcs/commitment_key.hpp>
 #include <memory>
 #include <vector>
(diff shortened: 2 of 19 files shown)


cpp/src/aztec/honk/proof_system/prover.cpp
```diff
@@ -78,7 +78,7 @@ template <typename settings> void Prover<settings>::compute_wire_commitments()
  * Z_perm[i] = ∏ --------------------------------------------------------------------------------
  *                  (w_1(j) + β⋅σ_1(j) + γ) ⋅ (w_2(j) + β⋅σ_2(j) + γ) ⋅ (w_3(j) + β⋅σ_3(j) + γ)
  *
- * where ∏ := ∏_{j=0:i-1} and id_i(X) = id(X) + n*(i-1). These evaluations are constructed over the
+ * where ∏ := ∏_{j=0:i-1} and id_i(X) = id(X) + n*(i-1). These evaluations are constructed over t;he
```
ledwards2225: artistic but unnecessary

cpp/src/aztec/honk/proof_system/prover.cpp
```diff
@@ -299,6 +300,17 @@ template <typename settings> void Prover<settings>::execute_relation_check_round
                                         sumcheck::GrandProductInitializationRelation>;
 
     // Compute alpha challenge
+    transcript.apply_fiat_shamir("zeta");
```
ledwards2225, line 303: This naming is now conflicting with what we call zeta_challenge  inShplonk. That's probably my fault - when I proposed zeta in that context I was forgetting about this zeta. We should change that back to z_challenge or something.

cpp/src/aztec/honk/proof_system/prover.cpp
```diff
@@ -299,6 +300,17 @@ template <typename settings> void Prover<settings>::execute_relation_check_round
...
+
+    // TODO(Cody): This is just temporary of course. Very inefficient, e.g., no commitment needed.
+    Fr zeta_challenge = transcript.get_challenge_field_element("zeta");
+    barretenberg::polynomial pow_zeta = power_polynomial::generate_vector(zeta_challenge, key->n);
+    key->polynomial_cache.put("pow_zeta", std::move(pow_zeta));
+    std::span<Fr> pow_zeta_fetched = key->polynomial_cache.get("pow_zeta");
```
ledwards2225: Don't need to 'get' here do we? Can't we just call commit on `pow_zeta`

cpp/src/aztec/honk/proof_system/verifier.hpp
```diff
@@ -24,7 +24,11 @@ template <typename program_settings> class Verifier {
     Verifier& operator=(const Verifier& other) = delete;
     Verifier& operator=(Verifier&& other);
 
-    // TODO: plonk_proof is just an std::vector<uint8_t>; probably shouldn't even exist
+    // TODO(luke): plonk_proof is just an std::vector<uint8_t>; probably shouldn't even exist
```
ledwards2225, line 27: Not very informative is it? All over the code we have lines like this `waffle::plonk_proof proof = prover.construct_proof();`. I'd say it's both redundant and misleading (given that `std::vector<uint8>` has nothing to do with Plonk). I think this is a case where you need a good variable name not a type alias. But even if you did, why not `using plonk_proof = std::vector<uint8>` instead of the struct. To me, though, even aliasing here is a bit like doing something like this: `using vector_size = size_t`.

cpp/src/aztec/honk/sumcheck/relations/grand_product_computation_relation.hpp
```diff
@@ -8,7 +8,7 @@ namespace honk::sumcheck {
 template <typename FF> class GrandProductComputationRelation : public Relation<FF> {
   public:
     // 1 + polynomial degree of this relation
-    static constexpr size_t RELATION_LENGTH = 5;
+    static constexpr size_t RELATION_LENGTH = 5; // TODO(Cody): Introduce pow_zeta; increment this by 1.
```
ledwards2225: Leaving this as a TODO?

cpp/src/aztec/honk/sumcheck/sumcheck.test.cpp
```diff
@@ -331,12 +335,13 @@ TEST(Sumcheck, ProverAndVerifierLonger)
     std::array<FF, multivariate_n> id_3           = { 0,  0,  0, 0 };
     std::array<FF, multivariate_n> lagrange_first = { 1,  0,  0, 0 };
     std::array<FF, multivariate_n> lagrange_last  = { 0,  0,  0, 1 };
+    std::array<FF, multivariate_n> pow_zeta       = { 2,  4,  8, 16 }; // TODO(Cody): fails with non-1 entry
```
ledwards2225, line 338: Is this comment still true? Maybe I'm misunderstanding its meaning

cpp/src/aztec/honk/composer/composer_helper/composer_helper.cpp
```diff
@@ -218,6 +218,8 @@ std::shared_ptr<waffle::proving_key> ComposerHelper<CircuitConstructor>::compute
 
     // TODO(Cody): this is a workaround
     circuit_proving_key->polynomial_cache.put("z_perm_lagrange", Polynomial<barretenberg::fr>(1));
+    // used in test_verification_key_creation
+    circuit_proving_key->polynomial_cache.put("pow_zeta", Polynomial<barretenberg::fr>(1));
```
ledwards2225: Is this necessary? Same question for `z_perm_lagrange`

cpp/src/aztec/honk/sumcheck/sumcheck.test.cpp
```diff
@@ -87,9 +87,11 @@ Transcript produce_mocked_transcript(size_t multivariate_d, size_t num_public_in
 
 TEST(Sumcheck, PolynomialNormalization)
 {
+    // Todo(Cody): We should not use real constants like this in the tests, at least not in so many of them.
     const size_t num_polys(bonk::StandardArithmetization::NUM_POLYNOMIALS);
     const size_t multivariate_d(3);
     const size_t multivariate_n(1 << multivariate_d);
+    ;
```
ledwards2225, line 94: Remove

cpp/src/aztec/honk/sumcheck/sumcheck_round.test.cpp
```diff
@@ -100,13 +101,15 @@ TEST(SumcheckRound, ComputeUnivariateVerifier)
     FF id_3 = { 0 };
     FF lagrange_first = { 0 };
     FF lagrange_last = { 0 };
+    FF pow_zeta = { 1 };
 
     // q_m * w_l * w_r + q_l * w_l + q_r * w_r + q_o * w_o + q_c
```
ledwards2225: Add pow_zeta to this comment? Seems to me we should be using the real expression for this test anyway rather than a hard coded `54`..

cpp/src/aztec/proof_system/flavor/flavor.hpp
```diff
@@ -87,10 +88,18 @@ struct StandardHonk {
         // Round 3
         manifest_rounds.emplace_back(transcript::Manifest::RoundManifest(
             { { .name = "Z_PERM", .num_bytes = g1_size, .derived_by_verifier = false } },
+            /* challenge_name = */ "zeta",
+            /* num_challenges_in = */ 1)
+        ); 
+
+        // Round 4
+        manifest_rounds.emplace_back(transcript::Manifest::RoundManifest(
+            { { .name = "POW_ZETA", .num_bytes = g1_size, .derived_by_verifier = false } },
```
ledwards2225, line 99: I think you said ultimately the commitment to pow_zeta is not needed? Should we just be generating two challenges (zeta, alpha) from the transcript after adding `Z_PERM`?

cpp/src/aztec/proof_system/types/polynomial_manifest.hpp
```diff
@@ -193,7 +194,8 @@ static constexpr PolynomialDescriptor standard_honk_polynomial_manifest[STANDARD
     PolynomialDescriptor("ID_2", "id_2_lagrange", false, false, PERMUTATION, ID_2),                  //
     PolynomialDescriptor("ID_3", "id_3_lagrange", true, false, PERMUTATION, ID_3),                   //
     PolynomialDescriptor("LAGRANGE_FIRST", "L_first_lagrange", false, false, OTHER, LAGRANGE_FIRST), //
-    PolynomialDescriptor("LAGRANGE_LAST", "L_last_lagrange", false, false, OTHER, LAGRANGE_LAST)     //
+    PolynomialDescriptor("LAGRANGE_LAST", "L_last_lagrange", false, false, OTHER, LAGRANGE_LAST),    //
+    PolynomialDescriptor("POW_ZETA", "pow_zeta", false, false, WITNESS, POW_ZETA)                    //
```
ledwards2225, line 198: Any particular reason why this is a WITNESS? Maybe this is all temporary and it doesn't matter?

ledwards2225 (changes requested):
Looks good to me once pow_zeta is added into the grand product computation relation. Just a few things I want to get some clarity on

cpp/src/aztec/honk/proof_system/prover.cpp
codygunton: Thanks

cpp/src/aztec/honk/proof_system/prover.cpp
codygunton: I think we caught that at the time? I'm not seeing another "zeta".

cpp/src/aztec/honk/proof_system/prover.cpp
codygunton: You're right, this was just a temporary thing due to some issue (I think I was just trying to access after the `std::move`). Fixed now.

cpp/src/aztec/honk/proof_system/verifier.hpp
codygunton: Yeah, fair enough.

cpp/src/aztec/honk/sumcheck/relations/grand_product_computation_relation.hpp
codygunton: Nope, got it, thanks.

cpp/src/aztec/honk/sumcheck/sumcheck.test.cpp
codygunton: Thanks

cpp/src/aztec/honk/composer/composer_helper/composer_helper.cpp
codygunton: I just tried removing running some tests. No, it seems they are not needed. I don't recall what was being worked around. Will remove.

cpp/src/aztec/honk/sumcheck/sumcheck_round.test.cpp
codygunton: Yeah, we should generate a full, nontrivial example and use that in various places.

cpp/src/aztec/proof_system/flavor/flavor.hpp
codygunton: Yeah, definitely not needed, the polynomial is public information. Basically, I had to determine if type should be `WITNESS` `SELECTOR` or `OTHER` and I went with `WITNESS` because `OTHER` automatically triggers a commitment when generating the vk, and rather than filling with some dummy value to enable that commitment, I took this approach (maybe a bad decision). Anyway, this should all go away soon!

cpp/src/aztec/proof_system/types/polynomial_manifest.hpp
codygunton: Cf the previous comment. It can't be computed without `zeta`, which depends on the witnesses.

ledwards2225 (approved):

PR #109, opened 2023-01-31
Polynomial cleanup

### LINKED ISSUE #115: Refactor polynomial class
Clean out outdated/unnecessary memory management stuff from the Polynomial class.

ledwards2225:
# Description

The purpose of this work is to clean out outdated/unnecessary memory management stuff from the Polynomial class.

- Reduce Polynomial data to these three: `mapped_`, `coefficients_`, and `size_`
- No more concept of `max_size_`, there is only `size_`
- Remove unused and/or unnecessary operations like 'bump_memory', resize, reserve. These have been replaced with appropriate ASSERTs or avoided altogether as appropriate
- Remove some unused or unnecessary functions like `add_lagrange_base_coefficient()`

# Checklist:

### DIFF
--- a/cpp/src/aztec/honk/pcs/shplonk/shplonk_multi.hpp
+++ b/cpp/src/aztec/honk/pcs/shplonk/shplonk_multi.hpp
@@ -109,9 +109,10 @@ template <typename Params> class MultiBatchOpeningScheme {
 
         // initialize Q(X) = 0
         Polynomial Q(max_poly_size);
+        Polynomial tmp(max_poly_size);
         for (size_t k = 0; k < num_multi_claims; ++k) {
             // Bₖ(X) into temp_poly
-            Polynomial tmp = merged_polynomials[k];
+            tmp = merged_polynomials[k];
             // subtract Bₖ(X) - Tₖ(X) in-place
             tmp -= interpolated_polynomials[k];
 
--- a/cpp/src/aztec/honk/pcs/shplonk/shplonk_multi.hpp
+++ b/cpp/src/aztec/honk/pcs/shplonk/shplonk_multi.hpp
@@ -82,7 +82,7 @@ template <typename Params> class MultiBatchOpeningScheme {
             // - Cₖ = ∑ⱼ ρʲ⋅Cⱼ
             // - Yₖ = {yᵏ₁, …, yᵏₘₖ}, where yᵏᵢ = ∑ⱼ ρʲ⋅yʲᵢ
             // create an empty polynomial in which we merge all polynomials openened at the same claim
-            auto& B_k = merged_polynomials.emplace_back(merged_poly_size, merged_poly_size);
+            auto& B_k = merged_polynomials.emplace_back(merged_poly_size);
             auto& C_k = merged_commitments.emplace_back(Commitment::zero());
             std::vector<Fr> evals_k(num_queries_k, Fr::zero());
 
@@ -108,7 +108,7 @@ template <typename Params> class MultiBatchOpeningScheme {
         }
 
         // initialize Q(X) = 0
-        Polynomial Q(max_poly_size, max_poly_size);
+        Polynomial Q(max_poly_size);
         for (size_t k = 0; k < num_multi_claims; ++k) {
             // Bₖ(X) into temp_poly
             Polynomial tmp = merged_polynomials[k];
--- a/cpp/src/aztec/honk/pcs/shplonk/shplonk_multi.hpp
+++ b/cpp/src/aztec/honk/pcs/shplonk/shplonk_multi.hpp
@@ -109,13 +109,8 @@ template <typename Params> class MultiBatchOpeningScheme {
 
         // initialize Q(X) = 0
         Polynomial Q(max_poly_size, max_poly_size);
-        // Polynomial tmp(size_t(0), max_poly_size);
         for (size_t k = 0; k < num_multi_claims; ++k) {
             // Bₖ(X) into temp_poly
-            // info("k = ", k);
-            // info("tmp.get_max_size() = ", tmp.get_max_size());
-            // info("merged_polynomials[j].get_max_size() = ", merged_polynomials[k].get_max_size());
-            // info("interpolated_polynomials[j].get_max_size() = ", interpolated_polynomials[k].get_max_size());
             Polynomial tmp = merged_polynomials[k];
             // subtract Bₖ(X) - Tₖ(X) in-place
             tmp -= interpolated_polynomials[k];
--- a/cpp/src/aztec/honk/pcs/shplonk/shplonk_multi.hpp
+++ b/cpp/src/aztec/honk/pcs/shplonk/shplonk_multi.hpp
@@ -82,7 +82,7 @@ template <typename Params> class MultiBatchOpeningScheme {
             // - Cₖ = ∑ⱼ ρʲ⋅Cⱼ
             // - Yₖ = {yᵏ₁, …, yᵏₘₖ}, where yᵏᵢ = ∑ⱼ ρʲ⋅yʲᵢ
             // create an empty polynomial in which we merge all polynomials openened at the same claim
-            auto& B_k = merged_polynomials.emplace_back(Polynomial(size_t(0), merged_poly_size));
+            auto& B_k = merged_polynomials.emplace_back(merged_poly_size, merged_poly_size);
             auto& C_k = merged_commitments.emplace_back(Commitment::zero());
             std::vector<Fr> evals_k(num_queries_k, Fr::zero());
 
@@ -108,7 +108,7 @@ template <typename Params> class MultiBatchOpeningScheme {
         }
 
         // initialize Q(X) = 0
-        Polynomial Q(size_t(0), max_poly_size);
+        Polynomial Q(max_poly_size, max_poly_size);
         // Polynomial tmp(size_t(0), max_poly_size);
         for (size_t k = 0; k < num_multi_claims; ++k) {
             // Bₖ(X) into temp_poly
--- a/cpp/src/aztec/honk/pcs/shplonk/shplonk_multi.hpp
+++ b/cpp/src/aztec/honk/pcs/shplonk/shplonk_multi.hpp
@@ -109,10 +109,14 @@ template <typename Params> class MultiBatchOpeningScheme {
 
         // initialize Q(X) = 0
         Polynomial Q(size_t(0), max_poly_size);
-        Polynomial tmp(size_t(0), max_poly_size);
+        // Polynomial tmp(size_t(0), max_poly_size);
         for (size_t k = 0; k < num_multi_claims; ++k) {
             // Bₖ(X) into temp_poly
-            tmp = merged_polynomials[k];
+            // info("k = ", k);
+            // info("tmp.get_max_size() = ", tmp.get_max_size());
+            // info("merged_polynomials[j].get_max_size() = ", merged_polynomials[k].get_max_size());
+            // info("interpolated_polynomials[j].get_max_size() = ", interpolated_polynomials[k].get_max_size());
+            Polynomial tmp = merged_polynomials[k];
             // subtract Bₖ(X) - Tₖ(X) in-place
             tmp -= interpolated_polynomials[k];
 
--- a/cpp/src/aztec/honk/pcs/shplonk/shplonk_single.hpp
+++ b/cpp/src/aztec/honk/pcs/shplonk/shplonk_single.hpp
@@ -45,6 +45,7 @@ template <typename Params> class SingleBatchOpeningScheme {
         }
         // Q(X) = ∑ⱼ ρʲ ⋅ ( fⱼ(X) − vⱼ) / ( X − xⱼ )
         Polynomial Q(max_poly_size);
+        Polynomial tmp(max_poly_size);
 
         Fr current_nu = Fr::one();
         for (size_t j = 0; j < num_claims; ++j) {
@@ -52,7 +53,7 @@ template <typename Params> class SingleBatchOpeningScheme {
             const auto& [commitment_j, opening_j, eval_j] = claims[j];
 
             // tmp = ρʲ ⋅ ( fⱼ(X) − vⱼ) / ( X − xⱼ )
-            Polynomial tmp = witness_polynomials[j];
+            tmp = witness_polynomials[j];
             tmp[0] -= eval_j;
             tmp.factor_roots(opening_j);
 
@@ -93,7 +94,7 @@ template <typename Params> class SingleBatchOpeningScheme {
             const auto& [commitment_j, opening_j, eval_j] = claims[j];
 
             // tmp = ρʲ ⋅ ( fⱼ(X) − vⱼ) / ( r − xⱼ )
-            Polynomial tmp = witness_polynomials[j];
+            tmp = witness_polynomials[j];
             tmp[0] -= eval_j;
             Fr scaling_factor = current_nu * inverse_vanishing_evals[j]; // = ρʲ / ( r − xⱼ )
 
--- a/cpp/src/aztec/honk/pcs/shplonk/shplonk_single.hpp
+++ b/cpp/src/aztec/honk/pcs/shplonk/shplonk_single.hpp
@@ -44,7 +44,7 @@ template <typename Params> class SingleBatchOpeningScheme {
             max_poly_size = std::max(max_poly_size, poly.size());
         }
         // Q(X) = ∑ⱼ ρʲ ⋅ ( fⱼ(X) − vⱼ) / ( X − xⱼ )
-        Polynomial Q(max_poly_size, max_poly_size);
+        Polynomial Q(max_poly_size);
 
         Fr current_nu = Fr::one();
         for (size_t j = 0; j < num_claims; ++j) {
--- a/cpp/src/aztec/honk/pcs/shplonk/shplonk_single.hpp
+++ b/cpp/src/aztec/honk/pcs/shplonk/shplonk_single.hpp
@@ -45,7 +45,6 @@ template <typename Params> class SingleBatchOpeningScheme {
         }
         // Q(X) = ∑ⱼ ρʲ ⋅ ( fⱼ(X) − vⱼ) / ( X − xⱼ )
         Polynomial Q(max_poly_size, max_poly_size);
-        Polynomial tmp(max_poly_size, max_poly_size);
 
         Fr current_nu = Fr::one();
         for (size_t j = 0; j < num_claims; ++j) {
@@ -53,7 +52,7 @@ template <typename Params> class SingleBatchOpeningScheme {
             const auto& [commitment_j, opening_j, eval_j] = claims[j];
 
             // tmp = ρʲ ⋅ ( fⱼ(X) − vⱼ) / ( X − xⱼ )
-            tmp = witness_polynomials[j];
+            Polynomial tmp = witness_polynomials[j];
             tmp[0] -= eval_j;
             tmp.factor_roots(opening_j);
 
@@ -94,7 +93,7 @@ template <typename Params> class SingleBatchOpeningScheme {
             const auto& [commitment_j, opening_j, eval_j] = claims[j];
 
             // tmp = ρʲ ⋅ ( fⱼ(X) − vⱼ) / ( r − xⱼ )
-            tmp = witness_polynomials[j];
+            Polynomial tmp = witness_polynomials[j];
             tmp[0] -= eval_j;
             Fr scaling_factor = current_nu * inverse_vanishing_evals[j]; // = ρʲ / ( r − xⱼ )
 
--- a/cpp/src/aztec/plonk/proof_system/prover/prover.test.cpp
+++ b/cpp/src/aztec/plonk/proof_system/prover/prover.test.cpp
@@ -243,11 +243,11 @@ waffle::Prover generate_test_data(const size_t n)
     q_m.ifft(key->small_domain);
     q_c.ifft(key->small_domain);
 
-    polynomial q_1_fft(q_l, n * 2);
-    polynomial q_2_fft(q_r, n * 2);
-    polynomial q_3_fft(q_o, n * 2);
-    polynomial q_m_fft(q_m, n * 2);
-    polynomial q_c_fft(q_c, n * 2);
+    polynomial q_1_fft(q_l, n * 4);
+    polynomial q_2_fft(q_r, n * 4);
+    polynomial q_3_fft(q_o, n * 4);
+    polynomial q_m_fft(q_m, n * 4);
+    polynomial q_c_fft(q_c, n * 4);
 
     q_1_fft.coset_fft(key->large_domain);
     q_2_fft.coset_fft(key->large_domain);
--- a/cpp/src/aztec/plonk/proof_system/prover/prover.test.cpp
+++ b/cpp/src/aztec/plonk/proof_system/prover/prover.test.cpp
@@ -115,14 +115,14 @@ waffle::Prover generate_test_data(const size_t n)
     auto reference_string = std::make_shared<waffle::FileReferenceString>(n + 1, "../srs_db/ignition");
     std::shared_ptr<proving_key> key = std::make_shared<proving_key>(n, 0, reference_string, waffle::STANDARD);
 
-    polynomial w_l(n, n);
-    polynomial w_r(n, n);
-    polynomial w_o(n, n);
-    polynomial q_l(n, n);
-    polynomial q_r(n, n);
-    polynomial q_o(n, n);
-    polynomial q_c(n, n);
-    polynomial q_m(n, n);
+    polynomial w_l(n);
+    polynomial w_r(n);
+    polynomial w_o(n);
+    polynomial q_l(n);
+    polynomial q_r(n);
+    polynomial q_o(n);
+    polynomial q_c(n);
+    polynomial q_m(n);
 
     for (size_t i = 0; i < n / 4; ++i) {
         w_l.at(2 * i) = fr::random_element();
--- a/cpp/src/aztec/plonk/proof_system/prover/prover.test.cpp
+++ b/cpp/src/aztec/plonk/proof_system/prover/prover.test.cpp
@@ -115,23 +115,14 @@ waffle::Prover generate_test_data(const size_t n)
     auto reference_string = std::make_shared<waffle::FileReferenceString>(n + 1, "../srs_db/ignition");
     std::shared_ptr<proving_key> key = std::make_shared<proving_key>(n, 0, reference_string, waffle::STANDARD);
 
-    polynomial w_l;
-    polynomial w_r;
-    polynomial w_o;
-    polynomial q_l;
-    polynomial q_r;
-    polynomial q_o;
-    polynomial q_c;
-    polynomial q_m;
-
-    w_l.resize(n);
-    w_r.resize(n);
-    w_o.resize(n);
-    q_l.resize(n);
-    q_r.resize(n);
-    q_o.resize(n);
-    q_m.resize(n);
-    q_c.resize(n);
+    polynomial w_l(n, n);
+    polynomial w_r(n, n);
+    polynomial w_o(n, n);
+    polynomial q_l(n, n);
+    polynomial q_r(n, n);
+    polynomial q_o(n, n);
+    polynomial q_c(n, n);
+    polynomial q_m(n, n);
 
     for (size_t i = 0; i < n / 4; ++i) {
         w_l.at(2 * i) = fr::random_element();
--- a/cpp/src/aztec/polynomials/polynomial.cpp
+++ b/cpp/src/aztec/polynomials/polynomial.cpp
@@ -77,12 +77,16 @@ Polynomial<Fr>::Polynomial(Fr* buf, const size_t size_)
 
 template <typename Fr> Polynomial<Fr>& Polynomial<Fr>::operator=(const Polynomial<Fr>& other)
 {
+    if (is_empty()) {
+        size_ = other.size();
+        coefficients_ = (Fr*)(aligned_alloc(32, sizeof(Fr) * other.capacity()));
+    }
+
     mapped_ = false;
 
     ASSERT(other.size_ <= size_);
 
     if (other.coefficients_ != nullptr) {
-        ASSERT(coefficients_);
         memcpy(static_cast<void*>(coefficients_), static_cast<void*>(other.coefficients_), sizeof(Fr) * other.size_);
     }
 
--- a/cpp/src/aztec/polynomials/polynomial.cpp
+++ b/cpp/src/aztec/polynomials/polynomial.cpp
@@ -48,17 +48,16 @@ Polynomial<Fr>::Polynomial(std::string const& filename)
 }
 
 template <typename Fr>
-Polynomial<Fr>::Polynomial(const size_t initial_size_, const size_t initial_max_size_hint)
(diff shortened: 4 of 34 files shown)


cpp/src/aztec/honk/pcs/shplonk/shplonk_multi.hpp
```diff
@@ -82,7 +82,7 @@ template <typename Params> class MultiBatchOpeningScheme {
             // - Cₖ = ∑ⱼ ρʲ⋅Cⱼ
             // - Yₖ = {yᵏ₁, …, yᵏₘₖ}, where yᵏᵢ = ∑ⱼ ρʲ⋅yʲᵢ
             // create an empty polynomial in which we merge all polynomials openened at the same claim
-            auto& B_k = merged_polynomials.emplace_back(Polynomial(size_t(0), merged_poly_size));
```
adr1anh, line 85: I can't remember whether it was necessary to have an initial size of zero, but it should be ok.

cpp/src/aztec/honk/pcs/shplonk/shplonk_single.hpp
```diff
@@ -44,16 +44,15 @@ template <typename Params> class SingleBatchOpeningScheme {
             max_poly_size = std::max(max_poly_size, poly.size());
         }
         // Q(X) = ∑ⱼ ρʲ ⋅ ( fⱼ(X) − vⱼ) / ( X − xⱼ )
-        Polynomial Q(max_poly_size, max_poly_size);
-        Polynomial tmp(max_poly_size, max_poly_size);
```
adr1anh, line 48: Leaving the `tmp` declaration outside the loop was done to prevent reallocations. With this change, a new large buffer gets reallocated to store the copy. However, the copy assignment operator should be able to reuse the existing buffer if the size is correct.

cpp/src/aztec/plonk/proof_system/prover/prover.test.cpp
```diff
@@ -252,11 +243,11 @@ waffle::Prover generate_test_data(const size_t n)
     q_m.ifft(key->small_domain);
     q_c.ifft(key->small_domain);
 
-    polynomial q_1_fft(q_l, n * 2);
-    polynomial q_2_fft(q_r, n * 2);
-    polynomial q_3_fft(q_o, n * 2);
-    polynomial q_m_fft(q_m, n * 2);
-    polynomial q_c_fft(q_c, n * 2);
```
adr1anh, line 259: Just checking, but why is the extra `2 * n` needed? Was this a bug before?

cpp/src/aztec/polynomials/polynomial.cpp
```diff
@@ -27,16 +18,13 @@ namespace barretenberg {
 template <typename Fr>
 Polynomial<Fr>::Polynomial(std::string const& filename)
     : mapped_(true)
```
adr1anh, line 21: Are we actually using `mapped` anywhere? What's the use-case for it?

cpp/src/aztec/polynomials/polynomial.cpp
```diff
@@ -48,112 +36,77 @@ Polynomial<Fr>::Polynomial(std::string const& filename)
...
 {
-    ASSERT(page_size_ != 0);
-    size_t target_max_size = std::max(initial_size_, initial_max_size_hint + DEFAULT_PAGE_SPILL);
-    if (target_max_size > 0) {
+    size_t target_size = size_ + DEFAULT_SIZE_INCREASE;
+    if (target_size > 0) {
 
-        max_size_ = target_max_size;
-        coefficients_ = (Fr*)(aligned_alloc(32, sizeof(Fr) * target_max_size));
+        coefficients_ = (Fr*)(aligned_alloc(32, sizeof(Fr) * target_size));
```
adr1anh: Just a thought, but we really need a wrapper for this at some point. 

Also, usually C-style casts are [discouraged](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-casts-named) in favor of `static_cast`. 

In this specific case, it should be okay, and most C code would be written like this anyway. However, `static_cast` should be preferred.

cpp/src/aztec/polynomials/polynomial.cpp
```diff
@@ -48,112 +36,77 @@ Polynomial<Fr>::Polynomial(std::string const& filename)
 }
 
 template <typename Fr>
-Polynomial<Fr>::Polynomial(const size_t initial_size_, const size_t initial_max_size_hint)
+Polynomial<Fr>::Polynomial(const size_t size_)
     : mapped_(false)
     , coefficients_(nullptr)
```
adr1anh: I wonder if these should not be defaulted in the header? It seems like these values are the default we'd want to enforce everywhere.

cpp/src/aztec/polynomials/polynomial.cpp
```diff
@@ -48,112 +36,77 @@ Polynomial<Fr>::Polynomial(std::string const& filename)
...
     : mapped_(other.mapped_)
     , coefficients_(other.coefficients_)
-    , initial_size_(other.initial_size_)
     , size_(other.size_)
-    , page_size_(other.page_size_)
-    , max_size_(other.max_size_)
-    , allocated_pages_(other.allocated_pages_)
 {
-    ASSERT(page_size_ != 0);
     other.coefficients_ = 0;
```
adr1anh: Should be `nullptr`

adr1anh: Also, you can change the constructor of the pointer argument to use [`std::exchange`](https://en.cppreference.com/w/cpp/utility/exchange).

cpp/src/aztec/polynomials/polynomial.hpp
```diff
@@ -245,20 +229,16 @@ template <typename Fr> class Polynomial {
...
-    void add_coefficient_internal(const Fr& coefficient);
-    void bump_memory(const size_t new_size);
+    void zero_memory(const size_t start_position, const size_t end_position);
+    // When a polynomial is instantiated from a size alone, the memory allocated corresponds to
+    // input size + DEFAULT_SIZE_INCREASE. A DEFAULT_SIZE_INCREASE of >= 1 is required to ensure
+    // that polynomials can be 'shifted' via a span of the 1st to size+1th coefficients.
+    const static size_t DEFAULT_SIZE_INCREASE = 1;
 
   public:
     bool mapped_;
```
adr1anh: Default to `false`

cpp/src/aztec/polynomials/polynomial.hpp
```diff
@@ -245,20 +229,16 @@ template <typename Fr> class Polynomial {
...
     Fr* coefficients_;
```
adr1anh: Default to `nullptr`

cpp/src/aztec/polynomials/polynomial.hpp
```diff
@@ -126,15 +119,8 @@ template <typename Fr> class Polynomial {
...
 
     Fr compute_kate_opening_coefficients(const Fr& z);
-    void add_lagrange_base_coefficient(const Fr& coefficient);
-    void add_coefficient(const Fr& coefficient);
-
-    void reserve(const size_t new_max_size);
-    void resize(const size_t new_size);
-    void resize_unsafe(const size_t new_size);
 
     bool empty() const
```
adr1anh: For this method, I'd add `[[nodiscard]]` so that someone gets a warning if they don't check the return value. They might think that this empties the polynomial .

cpp/src/aztec/polynomials/polynomial.hpp
```diff
@@ -78,7 +72,6 @@ template <typename Fr> class Polynomial {
     Fr* get_coefficients() { return coefficients_; };
 
     size_t get_size() const { return size_; };
```
adr1anh: Maybe we could deprecate this in favor of `.size()`

cpp/src/aztec/polynomials/polynomial.hpp
```diff
@@ -78,7 +72,6 @@ template <typename Fr> class Polynomial {
     Fr* get_coefficients() { return coefficients_; };
```
adr1anh, line 80: Should be `.data()` to match with usual containers.

cpp/src/aztec/polynomials/polynomial.hpp
```diff
@@ -78,7 +72,6 @@ template <typename Fr> class Polynomial {
...
 
     size_t get_size() const { return size_; };
-    size_t get_max_size() const { return max_size_; };
 
     // Const and non const versions of coefficient accessors
     Fr const& operator[](const size_t i) const
```
adr1anh, line 83: Not related to what you did, but I think the general consensus on the subscript operator, is that is is not bounds checked, though often enabled for DEBUG builds. The different with the `.at()` accessors is that the bounds check is never disabled. I think it may throw an exception.

cpp/src/aztec/polynomials/polynomial.hpp
```diff
@@ -47,11 +45,7 @@ template <typename Fr> class Polynomial {
 
         mapped_ = false;
         coefficients_ = 0;
```
adr1anh, line 46: `nullptr`

cpp/src/aztec/polynomials/polynomial.hpp
```diff
@@ -15,10 +15,8 @@ template <typename Fr> class Polynomial {
     // Creates a read only polynomial using mmap.
     Polynomial(std::string const& filename);
 
-    // TODO: add a 'spill' factor when allocating memory - we sometimes needs to extend poly degree by 2/4,
-    // if page size = power of two, will trigger unneccesary copies
-    Polynomial(const size_t initial_size, const size_t initial_max_size_hint = DEFAULT_SIZE_HINT);
-    Polynomial(const Polynomial& other, const size_t target_max_size = 0);
+    Polynomial(const size_t initial_size);
+    Polynomial(const Polynomial& other, const size_t target_size = 0);
```
adr1anh, line 19: With a default second argument, does this still count as the copy constructor? I need to check.

cpp/src/aztec/polynomials/polynomial.hpp
```diff
@@ -15,10 +15,8 @@ template <typename Fr> class Polynomial {
...
 
     Polynomial(Polynomial&& other) noexcept;
```
adr1anh, line 22: Can't comment the line below, but the signature looks a bit dangerous. Can we maybe remove it?

cpp/src/aztec/polynomials/polynomial.hpp
```diff
@@ -47,11 +45,7 @@ template <typename Fr> class Polynomial {
...
         mapped_ = false;
         coefficients_ = 0;
-        initial_size_ = 0;
         size_ = 0;
-        page_size_ = DEFAULT_SIZE_HINT;
-        max_size_ = 0;
-        allocated_pages_ = 0;
     }
 
     bool operator==(Polynomial const& rhs) const
```
adr1anh, line 51: What if two sizes are not equal, but the last coefficients of the larger vector are zero. This would happen if we did some operations like quotienting.

cpp/src/aztec/polynomials/polynomial.hpp
adr1anh: Also, the accessor just above violates const-ness. If should return `Fr const*`

cpp/src/aztec/polynomials/polynomial.cpp
```diff
@@ -496,37 +350,33 @@ template <typename Fr> Polynomial<Fr>& Polynomial<Fr>::operator+=(std::span<cons
...
 {
     ASSERT(!mapped_);
     const size_t other_size = other.size();
-    if (other_size > max_size_) {
-        std::cout << "bumping memory! hey this shouldn't really happen in prod!" << std::endl;
-        bump_memory(other_size);
-    }
-    size_ = std::max(size_, other_size);
+
+    ASSERT(size_ >= other_size);
```
adr1anh: I prefer this new approach. You're only allowed to do arithmetic on a polynomial if the sizes are good.

cpp/src/aztec/polynomials/polynomial.hpp
```diff
@@ -245,20 +229,16 @@ template <typename Fr> class Polynomial {
...
 
   public:
     bool mapped_;
     Fr* coefficients_;
-    size_t initial_size_;
-    size_t size_;      // This is the size() of the `coefficients` vector.
-    size_t page_size_; // DOCTODO: what does 'page' mean? Explain this.
-    size_t max_size_;
-    size_t allocated_pages_;
+    size_t size_; // This is the size() of the `coefficients` vector.
```
adr1anh: Making this `const` would prevent any tampering. I think we don't do resizing anymore.

Need to double check, because there are subtleties involved with having data member be const.

adr1anh (changes requested):
These are really good changes, and I'm happy the class is getting a nice upgrade!

I noted down a lot of minor C++ stuff more as suggestions and sharing tips. 

After reading the whole class through, I'm no longer sure that removing `max_size` is the right approach. There are many operations we perform that change the size of the polynomial in-place such as coset FFTs or even arithmetic operations between polynomials of different sizes. We also need a "hidden" extra slot at the end for the shift. 

When we only have the `size` variable, changing it means that we lose the information about the total size we have allocated. This is OK if we are only shrinking the size, but we may stil want to save that space for later. 
What would you say about having a `const size_t capacity` data member, which by default would be set to `size_ + 1`, but can be specified to allow for `4*n` in the case where we know the polynomial will be used to compute an in-place FFT of larger size?
The semantics for bounds checking could then be implemented as follows:
- `operator[]()` performs an `ASSERT(i < capacity_)`, so that accessing the shift via this operator is well defined.
- `at()` performs an `ASSERT(i < size_)` to check that we access a valid coefficient.
- When we perform an in-place operation, we increase the size to the appropriate domain size, if and only if the new size is smaller than the capacity.

Reducing the size should always be okay, but increasing it should always check that the new size is less than the capacity.

cpp/src/aztec/honk/pcs/shplonk/shplonk_single.hpp
ledwards2225: I agree for sure - should have left a comment here. I did this because this was the only point in production code that was calling the 'bump_memory' function, which doesn't seem like a desirable behavior. My thought was that this code has not been optimized and there should be a way to simply initialize a poly with sufficient memory then reuse that. (I think that was your intent here but I don't think that was the result). I'll take another look at this and see if there's an easy fix.

cpp/src/aztec/plonk/proof_system/prover/prover.test.cpp
ledwards2225: I'd say it was a typo/bug yes. Before my changes it would've called bump_memory under the hood but just setting the right value in the first place is obviously preferable.

cpp/src/aztec/polynomials/polynomial.cpp
ledwards2225: I don't think it's actually used in production (don't quote me on that) but it is needed for the memory mapping (mmap, munmap) functionality for polynomials. I think it's best to let it be for now.

cpp/src/aztec/polynomials/polynomial.cpp
ledwards2225: Yeah I just left this how it's always been. My ultimate goal would be to replace this c-style array with a better container anyway. If there's a performance benefit that can't be lost then so be it but it's something worth testing.

cpp/src/aztec/polynomials/polynomial.hpp
ledwards2225: I agree. I hadn't gotten to that but it's on the list. This was something Emerson added a long time ago.

cpp/src/aztec/polynomials/polynomial.hpp
ledwards2225: Makes sense. This was another Emerson addition and I'm actually hoping it can be gotten rid of altogether.

cpp/src/aztec/polynomials/polynomial.cpp
ledwards2225: Done

cpp/src/aztec/polynomials/polynomial.cpp
ledwards2225: Good call re `std::exchange`, I didn't know about that. Done

cpp/src/aztec/polynomials/polynomial.hpp
ledwards2225: All defaults now set appropriately

cpp/src/aztec/polynomials/polynomial.hpp
ledwards2225: done

cpp/src/aztec/polynomials/polynomial.hpp
ledwards2225: done

cpp/src/aztec/polynomials/polynomial.hpp
ledwards2225: Done

cpp/src/aztec/polynomials/polynomial.hpp
ledwards2225: This quickly starts to spiral so leaving a note but saving this for a follow on.

cpp/src/aztec/polynomials/polynomial.hpp
ledwards2225: Changed to `is_empty()` which should avoid any confusion

cpp/src/aztec/polynomials/polynomial.hpp
ledwards2225: Updated to account for this possibility.

cpp/src/aztec/honk/pcs/shplonk/shplonk_single.hpp
ledwards2225: This is now restored.

cpp/src/aztec/polynomials/polynomial.cpp
```diff
@@ -48,112 +37,63 @@ Polynomial<Fr>::Polynomial(std::string const& filename)
...
-    , initial_size_(initial_size_)
-    , size_(initial_size_)
-    , page_size_(DEFAULT_SIZE_HINT)
-    , max_size_(initial_size_)
-    , allocated_pages_(0)
+Polynomial<Fr>::Polynomial(Fr* buf, const size_t size_)
+    : coefficients_(buf)
+    , size_(size_)
+    , mapped_(false)
 {}
```
adr1anh: If this function is not used it should probably be removed

cpp/src/aztec/polynomials/polynomial.cpp
```diff
@@ -48,112 +37,63 @@ Polynomial<Fr>::Polynomial(std::string const& filename)
...
         ASSERT(coefficients_);
-        memcpy(static_cast<void*>(coefficients_), static_cast<void*>(other.coefficients_), sizeof(Fr) * size_);
+        memcpy(static_cast<void*>(coefficients_), static_cast<void*>(other.coefficients_), sizeof(Fr) * other.size_);
     }
 
-    zero_memory(max_size_);
+    zero_memory(other.size_, size_);
 
     return *this;
 }
```
adr1anh: I think we may want to handle the case where we assign to a default constructed polynomial (i.e. `coefficients_ == nullptr` and `size_ == 0`) and allocate in this very specific situation. Default initialization is good to have in the proof system since we often store polynomials in containers.

cpp/src/aztec/polynomials/polynomial.cpp
```diff
@@ -165,18 +105,11 @@ template <typename Fr> Polynomial<Fr>& Polynomial<Fr>::operator=(Polynomial&& ot
...
-    ASSERT(page_size_ != 0);
-
-    other.coefficients_ = nullptr;
-    // Clear other, so we can detect use on free after poly's are put in cache
-    other.clear();
+    // simultaneously set members and clear other
+    coefficients_ = std::exchange(other.coefficients_, nullptr);
+    size_ = std::exchange(other.size_, 0);
+    mapped_ = std::exchange(other.mapped_, false);
+
```
adr1anh, line 106: Nice!

cpp/src/aztec/polynomials/polynomial.cpp
```diff
@@ -465,7 +297,7 @@ Fr Polynomial<Fr>::evaluate_from_fft(const EvaluationDomain<Fr>& large_domain,
 
 template <typename Fr>
 Polynomial<Fr>::Polynomial(std::span<const Fr> interpolation_points, std::span<const Fr> evaluations)
-    : Polynomial(interpolation_points.size(), interpolation_points.size())
+    : Polynomial(interpolation_points.size())
```
adr1anh, line 269: Not sure if it was noted before, but this could probably be a static method to avoid too many constructors.

cpp/src/aztec/polynomials/polynomial.hpp
adr1anh: Seems good https://stackoverflow.com/questions/2787569/copy-constructor-with-default-arguments

cpp/src/aztec/polynomials/polynomial.hpp
```diff
@@ -242,23 +218,23 @@ template <typename Fr> class Polynomial {
...
-    size_t size_;      // This is the size() of the `coefficients` vector.
-    size_t page_size_; // DOCTODO: what does 'page' mean? Explain this.
-    size_t max_size_;
-    size_t allocated_pages_;
+    Fr* coefficients_ = nullptr;
+    // The size_ effectively represents the 'usable' length of the coefficients array but may be less than the true
+    // 'capacity' of the array. It is not explicitly tied to the degree and is not changed by any operations on the
+    // polynomial.
+    size_t size_ = 0;
+    bool mapped_ = false;
```
adr1anh, line 244: Since these are defaulted, we can add the default constructor `Polynomial() = default;`

cpp/src/aztec/proof_system/work_queue/work_queue.cpp
```diff
@@ -248,16 +248,15 @@ void work_queue::process_queue()
         case WorkType::FFT: {
             using namespace barretenberg;
             polynomial& wire = key->polynomial_cache.get(item.tag);
-            polynomial wire_fft(4 * key->circuit_size + 4, 4 * key->circuit_size + 4);
+            polynomial wire_fft(4 * key->circuit_size + 4);
 
             polynomial_arithmetic::copy_polynomial(
                 &wire[0], &wire_fft[0], key->circuit_size, 4 * key->circuit_size + 4);
```
adr1anh: We can replace `&wire[0]` with `wire.data()` and the same for `wire_fft`.

cpp/src/aztec/proof_system/work_queue/work_queue.cpp
```diff
@@ -248,16 +248,15 @@ void work_queue::process_queue()
...
 
             wire_fft.coset_fft(key->large_domain);
-            wire_fft.add_lagrange_base_coefficient(wire_fft[0]);
-            wire_fft.add_lagrange_base_coefficient(wire_fft[1]);
-            wire_fft.add_lagrange_base_coefficient(wire_fft[2]);
-            wire_fft.add_lagrange_base_coefficient(wire_fft[3]);
+            for (size_t i = 0; i < 4; i++) {
+                wire_fft[4 * key->circuit_size + i] = wire_fft[i];
+            }
```
adr1anh, line 259: Just a thought that's maybe not 100% relevant for this PR. What we're doing here is saying that `wire_fft` is shiftable by 4. In fact, this is just saying that the polynomial has a "cyclic shift". In Honk, by setting the first and (n+1)-th to zero, we are also enabling a cyclic shift, but by 1 instead. When the polynomial is in Lagrange form, then the shift is 1, but the initial coefficient may be non-zero. 

When we get to modernizing PlonK, we may want to unify this observation, to allow us for to use `.shifted()` on different representations, and let the "excess capacity" be customizable, and automatically ensure that the initial values are copies to the end of the vector.

cpp/src/aztec/proof_system/polynomial_cache/polynomial_store_mem.cpp
```diff
@@ -17,7 +18,7 @@ template <typename... Args> inline void debug(Args...) {}
 void PolynomialStoreMem::put(std::string const& key, polynomial const& value)
 {
     debug("put: taking copy of polynomial: ", key);
-    map_[key] = value;
+    map_.emplace(key, std::move(value));
     auto& stats = stats_[key];
     stats.first++;
 }
```
adr1anh: I think this may cause an issue for the caller. `value` is passed as reference, but it gets moved into the store. The caller will then have a polynomial in an invalid state. 

Solution:
- Either remove the `std::move(value)` and replace with `value`, thereby creating a copy.
- Overload this method in the `value` parameter to accept both a reference and movable argument. This allows the caller to call `put("w1", std::move(w1))` when it wants to signal that `w1` is no longer needed in this scope, and it will access it later using the cache.


I think the code would look like this. 
```c++
void PolynomialStoreMem::put(std::string const& key, polynomial && value)
 {
     debug("put: taking copy of polynomial: ", key);
     map_.emplace(key, std::move(value));
     auto& stats = stats_[key];
     stats.first++;
 }

void PolynomialStoreMem::put(std::string const& key, polynomial const& value)
 {
     debug("put: taking copy of polynomial: ", key);
     map_.emplace(key, value);
     auto& stats = stats_[key];
     stats.first++;
 }
```

The expression `std::move(value)` doesn't actually call the move constructor, it's just a way of casting `value` to a `polynomial&&` type. When we use `std::move` on an argument passed to a function, the compiler will call the method which accepts `polynomial &&` instead. However, inside the `put` function, `polynomial&& value` decays back to a reference, so we need another `std::move` in the `emplace` line to convert `value` to a `polynomial&&` again, so that `emplace` calls `polynomial`'s move constructor instead of the copy constructor.

 The `&&` notation is just to signal that we can rip the guts out of `value` since the caller doesn't need it anymore. But it behaves like a `polynomial&` inside the function.

cpp/src/aztec/polynomials/polynomial.cpp
```diff
@@ -48,112 +37,63 @@ Polynomial<Fr>::Polynomial(std::string const& filename)
...
-    : mapped_(false)
-    , coefficients_(0)
-    , initial_size_(0)
+    : coefficients_(nullptr)
     , size_(0)
-    , page_size_(DEFAULT_SIZE_HINT)
-    , max_size_(0)
-    , allocated_pages_(0)
+    , mapped_(false)
 {}
```
adr1anh: Can replace with `Polynomial<Fr>::Polynomial() = default`

cpp/src/aztec/polynomials/polynomial.cpp
```diff
@@ -48,112 +37,63 @@ Polynomial<Fr>::Polynomial(std::string const& filename)
...
 
-    ASSERT(max_size_ >= size_);
+    ASSERT(other.size_ <= size_);
 
-    if (other.coefficients_ != 0) {
+    if (other.coefficients_ != nullptr) {
         ASSERT(coefficients_);
-        memcpy(static_cast<void*>(coefficients_), static_cast<void*>(other.coefficients_), sizeof(Fr) * size_);
+        memcpy(static_cast<void*>(coefficients_), static_cast<void*>(other.coefficients_), sizeof(Fr) * other.size_);
     }
```
adr1anh: Since we allow default-constructible polynomials, we should allocate the memory here. This would only be in the situation where `coefficients_ == nullptr` and `size_ == 0`.

cpp/src/aztec/polynomials/polynomial.cpp
ledwards2225: This function is used for reading polys from a `mmap`. It may be possible to get rid of it at some point but it's used at various points in our testing suite, for example, to reuse a proving key that has been written to file.

cpp/src/aztec/polynomials/polynomial.hpp
ledwards2225: Good call, done.

cpp/src/aztec/proof_system/work_queue/work_queue.cpp
ledwards2225: Done

cpp/src/aztec/proof_system/work_queue/work_queue.cpp
ledwards2225: Agreed

cpp/src/aztec/polynomials/polynomial.cpp
ledwards2225: Done

cpp/src/aztec/proof_system/polynomial_cache/polynomial_store_mem.cpp
ledwards2225: Good catch, this was a bug left over from a debug experiment. What I actually wanted here is simply `map_.emplace(key, value);`. The point of my change in the first place was to get rid of unnecessary default polynomial construction (which happens with `map_[key] = value;`).

cpp/src/aztec/polynomials/polynomial.cpp
ledwards2225: Yeah I debated about this one. We don't currently do this anywhere and I don't feel we should need to. (The closest thing is something like `barretenberg::polynomial my_poly; my_poly = polynomial(n)` which will use the move assignment operator. Even that should not be necessary but it happens in several places and I didn't want to root it out. Do you see some reason why we need to have an container of empty polynomials rather than constructing polynomials in place in an empty container?

cpp/src/aztec/polynomials/polynomial.cpp
ledwards2225: This is a good point but as I mentioned above, I actually think we should get away from default constructed polys all together. If there is good reason to have them, I agree this change is necessary.

cpp/src/aztec/polynomials/polynomial.cpp
ledwards2225: I agree but I'd like to just leave things like this for a follow on to include moving things like this out of Polynomial entirely and into, say, polynomial_arithmetic

cpp/src/aztec/polynomials/polynomial.cpp
ledwards2225: I've added functionality to handle (non-move) assignment to a default constructed polynomial

cpp/src/aztec/polynomials/polynomial.cpp
ledwards2225: Now handling this case and doing memory allocation here as suggested. Test added to verify correctness.

cpp/src/aztec/polynomials/polynomial.hpp
```diff
@@ -219,10 +225,11 @@ template <typename Fr> class Polynomial {
 
     std::size_t size() const { return size_; }
     std::size_t capacity() const { return size_ + DEFAULT_CAPACITY_INCREASE; }
+    std::size_t mapped() const { return mapped_; }
```
adr1anh: should return `bool`

adr1anh (commented):
Except for the `mapped()` return type. This looks good to me!

cpp/src/aztec/polynomials/polynomial.hpp
ledwards2225: Woops. done

adr1anh (approved):
LGTM

PR #111, opened 2023-02-01
Added RAM abstraction into UltraPlonk

zac-williamson:
# Description

Updates UltraComposer to support RAM gates.

A RAM gate allows a use to read and write into a memory table, where both the index entry and value are witness values, not known at circuit-compile time.

See (https://hackmd.io/@aztec-network/Byky93bXs?type=view) for full description of circuit algebra.

Cost is 3.25 gates per read/write op. Each table index must have a minimum of 1 read/write op (enforced automatically by composer), which adds a constant gate overhead per table.

Also fixed an error in UltraPlonk where the AuxiliaryWidget was not correctly constraining memory lookups.

# Checklist:


cpp/src/aztec/plonk/proof_system/widgets/transition_widgets/plookup_auxiliary_widget.hpp
```diff
@@ -146,84 +171,150 @@ template <class Field, class Getters, typename PolyContainer> class PlookupAuxil
...
-         * A record gate is structured such that each of the above 4 fields maps to wires 1, 2, 3, 4
-         **/
-
-        /**
-         * RAM SORTED LIST CHECK
+         * A ROM memory record contains a tuple of the following fields:
+         *  * i: `index` of memory cell being accessed
+         *  * v: `value1` of memory cell being accessed (ROM tables can store up to 2 values per index)
+         *  * v2:`value2` of memory cell being accessed (ROM tables can store up to 2 values per index)
+         *  * r: `record` of memory cell. record = index * eta + timestamp * eta^2 + value * eta^3
```
adr1anh: Should be `record = index * eta + value2 * eta^2 + value1 * eta^3`

cpp/src/aztec/plonk/composer/ultra_composer.cpp
```diff
@@ -2498,10 +2800,159 @@ void UltraComposer::process_ROM_array(const size_t rom_id, const size_t gate_off
...
+            break;
+        }
+        case RamRecord::AccessType::WRITE: {
+            memory_write_records.push_back(
+                static_cast<uint32_t>(sorted_record.gate_index + gate_offset_from_public_inputs));
+            memory_write_records.push_back(static_cast<uint32_t>(record.gate_index + gate_offset_from_public_inputs));
+            break;
+        }
+        default: {
+            ASSERT(false); // shouldn't get here!
```
adr1anh, line 2840: https://en.cppreference.com/w/cpp/utility/unreachable is coming, but for now maybe `__builtin_unreachable();`. Is this equivalent to `ASSERT(false)`?

cpp/src/aztec/plonk/composer/ultra_composer.cpp
```diff
@@ -2498,10 +2800,159 @@ void UltraComposer::process_ROM_array(const size_t rom_id, const size_t gate_off
...
+        }
+        }
+    }
+
+    // Step 2: Create gates that validate correctness of RAM timestamps
+
+    std::vector<uint32_t> timestamp_deltas;
+    for (size_t i = 0; i < ram_array.records.size() - 1; ++i) {
+        // create_RAM_timestamp_gate(sorted_records[i], sorted_records[i + 1])
```
adr1anh: Remove comment?

cpp/src/aztec/plonk/composer/ultra_composer.cpp
```diff
@@ -2498,10 +2800,159 @@ void UltraComposer::process_ROM_array(const size_t rom_id, const size_t gate_off
...
+    // e.g. if a RAM record is uninitialized but the index of that record is a function of public/private inputs,
+    // different public iputs will produce different circuit constraints.
+    for (size_t i = 0; i < ram_array.state.size(); ++i) {
+        if (ram_array.state[i] == UNINITIALIZED_MEMORY_RECORD) {
+            init_RAM_element(ram_id, static_cast<uint32_t>(i), zero_idx);
+        }
+    }
+
+#ifdef NO_TBB
+    std::sort(ram_array.records.begin(), ram_array.records.end());
```
adr1anh, line 2828: sort according to `(index, timestamp)`

cpp/src/aztec/plonk/composer/ultra_composer.cpp
```diff
@@ -2498,10 +2800,159 @@ void UltraComposer::process_ROM_array(const size_t rom_id, const size_t gate_off
...
+#else
+    std::sort(std::execution::par_unseq, ram_array.records.begin(), ram_array.records.end());
+#endif
+
+    // Iterate over all but final RAM record.
```
adr1anh, line 2833:
```cpp
// This loop adds constraints to ensure the records are sorted according to (index, timestamp)
```

cpp/src/aztec/plonk/composer/ultra_composer.cpp
adr1anh: unused

cpp/src/aztec/plonk/proof_system/widgets/transition_widgets/plookup_auxiliary_widget.hpp
```diff
@@ -146,84 +171,150 @@ template <class Field, class Getters, typename PolyContainer> class PlookupAuxil
...
+        /**
+         * RAM Consistency Check
+         *
+         * The 'access' type of the record is extracted with the expression `w_4 - partial_record_check`
+         * (i.e. for an honest Prover `w1 * eta + w2 * eta^2 + w3 * eta^3 - w4 = access`.
+         * This is validated by requiring `access` to be boolean
+         *
+         * For two adjacent entries in the sorted list, their values must match iff:
+         *  * index values match
+         *  * adjacent access value is 0 (i.e. next gate is a READ)
```
adr1anh: The condition is not "if and only if", but rather:
If `(adjacent access value is 0) AND (index values match)` then `adjacent values match`.

It seems like if the other direction held, then we would not be able to over-write with the existing value.

cpp/src/aztec/plonk/composer/ultra_composer.hpp
```diff
@@ -57,40 +59,80 @@ class UltraComposer : public ComposerBase {
...
-        size_t gate_index;
-        bool operator<(const MemoryRecord& other) const
+    struct RomRecord {
+        uint32_t index_witness = 0;
+        uint32_t value_column1_witness = 0;
+        uint32_t value_column2_witness = 0;
+        uint32_t index = 0;
+        uint32_t record_witness = 0;
+        size_t gate_index = 0;
+        bool operator<(const RomRecord& other) const { return index < other.index; }
```
adr1anh, line 73: Is this partial order enough? What would happen if we have two entries with the same `value_column1_witness`? Could we end up in a situation where the entries are in the wrong order in different places?

cpp/src/aztec/plonk/composer/ultra_composer.hpp
```diff
@@ -57,40 +59,80 @@ class UltraComposer : public ComposerBase {
...
+        // A vector of records, each of which contains:
+        // + The constant witness with the index
+        // + The value in the memory slot
+        // + The actual index value
+        std::vector<RamRecord> records;
+
+        // used for RAM records, to compute the timestamp when performing a read/write
+        size_t access_count = 0;
+    };
+
```
adr1anh: This structure is not super clear.

```cpp
// Represents the current state of the array. Elements are variable indices.
// Every update requires a new entry in the `records` vector below.
std::vector<uint32_t> state;
```
```cpp
// A vector of records, each of which contains:
// - Witnesses for [index, timestamp, value, record]
//   (record is initialized during the proof creation, and points to 0 until then)
// - Index of the element in the `state` vector
// - READ/WRITE flag
// - Real timestamp value, initialized to the current `access_count`
std::vector<RamRecord> records;

// used for RAM records, to compute the timestamp when performing a read/write
// Incremented at every init/read/write operation.
size_t access_count = 0;
```

cpp/src/aztec/plonk/composer/ultra_composer.hpp
```diff
@@ -378,9 +465,26 @@ class UltraComposer : public ComposerBase {
...
+    /**
+     * @brief Each entry in ram_arrays represents an independent RAM table.
+     * RamTranscript tracks the current table state,
+     * as well as the 'records' produced by each read and write operation.
+     * Used in `compute_proving_key` to generate consistency check gates required to validate the RAM read/write history
+     */
+    std::vector<RamTranscript> ram_arrays;
+
+    /**
+     * @brief Each entry in rim_arrays represents an independent ROM table.
```
adr1anh: typo in `rom_arrays`

adr1anh (commented):
Just a few comments regarding the data structures, though it became clearer after getting through all the code.

cpp/src/aztec/plonk/composer/ultra_composer.hpp
zac-williamson: I think this is working as intended. The purpose of `operator<` is to produce a set of ROM records that are sorted by index value as this forms part of the witness commitment.

We don't have a need to sort by value. We could add comments to the `<` operator explaining what it's used for as it's not sufficient to do a full "sort" of Rom records.

cpp/src/aztec/plonk/proof_system/widgets/transition_widgets/plookup_auxiliary_widget.hpp
zac-williamson: Yes that's a good point, would you mind updating the comment?

PR #112, opened 2023-02-01
Added dynamic array abstraction into standard library

zac-williamson:
# Description

Added classes to the stdlib that leverage the new ram gates

stdlib::ram_table is a fixed-size table of field elements that can be read from/written to at indexes not known at circuit-compile time

stdlib::dynamic_array emulates a variable-sized array using stdlib::ram_table. A maximum potential size must be defined that affects the number of constraints required to initialise/resize the array.

# Checklist:


cpp/src/aztec/plonk/composer/composer_base.cpp
```diff
@@ -154,6 +154,8 @@ template <size_t program_width, bool with_tags> void ComposerBase::compute_sigma
                 }
                 if (last_node) {
                     sigma_mappings[current_column][current_row].is_tag = true;
+
+                    // TODO: yikes, std::maps are expensive. Can we find a way to get rid of this?
```
Rumata888: If this is really slow, maybe we should create an issue to take care of this later

cpp/src/aztec/plonk/composer/ultra_composer.hpp
```diff
@@ -195,6 +245,32 @@ class UltraComposer : public ComposerBase {
         }
 
         constexpr size_t gate_width = ultra_settings::program_width;
+        // each RAM gate adds +2 extra gates due to the ram reads being copied to a sorted list set,
+        // as well as an extra gate to validate timestamps
+        for (size_t i = 0; i < ram_arrays.size(); ++i) {
+            for (size_t j = 0; j < ram_arrays[i].state.size(); ++j) {
+                if (ram_arrays[i].state[j] == UNINITIALIZED_MEMORY_RECORD) {
+                    ramcount += 2;
```
Rumata888: I'd replace these inlined constants with definitions, so that it is easier to understand the logic

cpp/src/aztec/plonk/composer/ultra_composer.hpp
```diff
@@ -195,6 +245,32 @@ class UltraComposer : public ComposerBase {
...
+            }
+            ramcount += (ram_arrays[i].records.size() * 2);
+            ramcount += 1; // we add an addition gate after procesing a ram array
+
+            // there will be 'max_timestamp' number of range checks, need to calculate.
+            const auto max_timestamp = ram_arrays[i].access_count - 1;
+
+            // TODO: if a range check of length `max_timestamp` already exists, this will be innacurate!
+            // TODO: fix this
+            size_t padding = (gate_width - (max_timestamp % gate_width)) % gate_width;
```
Rumata888: Should we leave this TODO? If we forget about it, there can be all kinds of weird cases. Especially, if we reintroduce check_circuit for testing

cpp/src/aztec/plonk/composer/ultra_composer.hpp
```diff
@@ -378,9 +467,26 @@ class UltraComposer : public ComposerBase {
...
+    /**
+     * @brief Each entry in ram_arrays represents an independent RAM table.
+     * RamTranscript tracks the current table state,
+     * as well as the 'records' produced by each read and write operation.
+     * Used in `compute_proving_key` to generate consistency check gates required to validate the RAM read/write history
+     */
+    std::vector<RamTranscript> ram_arrays;
+
+    /**
+     * @brief Each entry in rim_arrays represents an independent ROM table.
```
Rumata888: Typo: rim_arrays

cpp/src/aztec/plonk/composer/ultra_composer.test.cpp
```diff
@@ -582,6 +582,28 @@ TEST(ultra_composer, range_with_gates)
     bool result = verifier.verify_proof(proof);
     EXPECT_EQ(result, true);
 }
+
+TEST(ultra_composer, range_with_gates_where_range_is_not_a_power_of_two)
```
Rumata888: What is the purpose of this test? Why are we testing this separately? I think there should be a comment

cpp/src/aztec/plonk/composer/ultra_composer.test.cpp
```diff
@@ -764,4 +786,75 @@ TEST(ultra_composer, rom)
...
+    };
+
+    size_t ram_id = composer.create_RAM_array(8);
+
+    for (size_t i = 0; i < 8; ++i) {
+        composer.init_RAM_element(ram_id, i, ram_values[i]);
+    }
+
+    uint32_t a_idx = composer.read_RAM_array(ram_id, composer.add_variable(5));
+    EXPECT_EQ(a_idx != ram_values[5], true);
```
Rumata888: I think this needs a comment, explaining why this is the expected behavior

cpp/src/aztec/plonk/composer/ultra_composer.hpp
```diff
@@ -149,13 +191,15 @@ class UltraComposer : public ComposerBase {
...
     }
 
-    void create_new_range_constraint(const uint32_t variable_index, const uint64_t target_range);
-    void create_range_constraint(const uint32_t variable_index, const size_t num_bits, std::string const&)
+    void create_new_range_constraint(const uint32_t variable_index,
+                                     const uint64_t target_range,
+                                     std::string const msg = "create_new_range_constraint");
+    void create_range_constraint(const uint32_t variable_index, const size_t num_bits, std::string const& msg)
     {
         if (num_bits <= DEFAULT_PLOOKUP_RANGE_BITNUM) {
```
Rumata888: Why not just use decompose_into_default_range for both cases since it will default to create_new_range_constraint in the < DEFAULT case?

cpp/src/aztec/stdlib/primitives/memory/dynamic_array.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+namespace plonk {
+namespace stdlib {
+
+/**
+ * @brief Construct a new Dynamic Array< Composer>:: Dynamic Array object
+ *
+ * @details Dynamic arrays require a maximum size when created, that cannot be exceeded.
+ *          Read and write operations cost 3.25 UltraPlonk gates.
+ *          Each dynamic array requires an additional 3.25 * maximum_size number of gates
```
Rumata888: I'd add a hint that creating non power of 2 max-sized dynamic arrays can also be costly because of creating independent range constraint tables (I assume power of 2 ones will often be used in general for bit range constraints)

cpp/src/aztec/stdlib/primitives/memory/dynamic_array.cpp
```diff
@@ -0,0 +1,277 @@
...
+    _inner_table.write(index, value);
+}
+
+/**
+ * @brief Push a field element onto the dynamic array
+ *
+ * @tparam Composer
+ * @param value
+ */
+template <typename Composer> void DynamicArray<Composer>::push(const field_pt& value)
```
Rumata888: I think you forgot a range check for the new size

cpp/src/aztec/stdlib/primitives/memory/dynamic_array.cpp
```diff
@@ -0,0 +1,277 @@
...
+ * @brief Conditionally push a field element onto the dynamic array
+ *
+ * @tparam Composer
+ * @param predicate
+ * @param value
+ */
+template <typename Composer>
+void DynamicArray<Composer>::conditional_push(const bool_pt& predicate, const field_pt& value)
+{
+    if (native_size() >= _max_size) {
```
Rumata888: Also no bounds range constraint

cpp/src/aztec/stdlib/primitives/memory/ram_table.cpp
```diff
@@ -0,0 +1,257 @@
...
+
+    if (uint256_t(index.get_value()) >= _length) {
+        // TODO: what's best practise here? We are assuming that this action will generate failing constraints,
+        // and we set failure message here so that it better describes the point of failure.
+        // However, we are not *ensuring* that failing constraints are generated at the point that `failure()` is
+        // called. Is this ok?
+        _context->failure("ram_table: RAM array access out of bounds");
+    }
+
+    initialize_table();
```
Rumata888: This will fail if all operations were with constants. If we check, whether the context is a nullptr, we can simply return a constant value in the constant case

Rumata888 (approved):
The only thing I'm worried about here are the range constraints on pushes

cpp/src/aztec/stdlib/primitives/memory/dynamic_array.cpp
zac-williamson: The `RamTable::write` method should take care of this.

The `DynamicArray` class wraps a `RamTable` table object with a fixed maximum size. We apply an arithmetic constraint that checks the final 'index' value in our sorted list of RAM read/writes equals the maximum size (ultra_composer.cpp line 2486), if the index is out-of-bounds this constraint will be unsatisfied.

cpp/src/aztec/stdlib/primitives/memory/dynamic_array.cpp
zac-williamson: see previous comment

cpp/src/aztec/plonk/composer/ultra_composer.hpp
zac-williamson: I think it's worth making an issue about so we don't forget, but not delaying this PR. It's a very niche issue that's a pain to correctly handle.

cpp/src/aztec/plonk/composer/ultra_composer.hpp
zac-williamson: updated

cpp/src/aztec/plonk/composer/ultra_composer.hpp
zac-williamson: fixed

cpp/src/aztec/plonk/composer/ultra_composer.hpp
zac-williamson: fixed

cpp/src/aztec/plonk/composer/ultra_composer.test.cpp
zac-williamson: there was a bug where all range checks were being bumped to be a power of two. Added this test to verify this bug has been fixed

cpp/src/aztec/stdlib/primitives/memory/dynamic_array.cpp
zac-williamson: updated.

cpp/src/aztec/stdlib/primitives/memory/ram_table.cpp
zac-williamson: Checking the context pointer isn't sufficient to determine whether the operations are all over constants. Other methods in this class will simply break if the pointer is null (see the asserts in other methods)

It's a major pain to get the RAM table implementation to a state where it can correctly handle pure constant cases. I took the simpler approach where we just instantiate all constants as witnesses and constrain them to be equal to the required values. Will require a substantial rewrite to provide a constant-friendly code path that can accomodate a null context pointer.

PR #134, opened 2023-02-07
feat: Utilize globally installed leveldb if available

phated:
# Description

This allows leveldb to be provided globally in environments where there is no network access or where the sandbox cannot be changed (e.g. nix) during the building of the package. This is done through `FetchContent_MakeAvailable` which can be used in combination with `FIND_PACKAGE_ARGS` to start with a `find_package` search before trying to fetch from the network.

This changes requires a CMake minimum version upgrade because `FIND_PACKAGE_ARGS` wasn't added until v3.24 as documented at https://cmake.org/cmake/help/latest/module/FetchContent.html#command:fetchcontent_declare. Does anyone know which version of CMake is being used in the docker containers?

# Checklist:

### DIFF
--- a/cpp/src/aztec/stdlib/merkle_tree/CMakeLists.txt
+++ b/cpp/src/aztec/stdlib/merkle_tree/CMakeLists.txt
@@ -1,26 +1,49 @@
+barretenberg_module(stdlib_merkle_tree stdlib_primitives stdlib_blake3s stdlib_pedersen)
+
 if(NOT WASM)
     include(FetchContent)
     FetchContent_Declare(
         leveldb
         GIT_REPOSITORY https://github.com/google/leveldb.git
         GIT_TAG 1.22
+        FIND_PACKAGE_ARGS
     )
 
-    FetchContent_GetProperties(leveldb)
-    if(NOT leveldb_POPULATED)
-        FetchContent_Populate(leveldb)
-        set(LEVELDB_BUILD_TESTS OFF CACHE BOOL "LevelDB tests off")
-        add_subdirectory(${leveldb_SOURCE_DIR} ${leveldb_BINARY_DIR} EXCLUDE_FROM_ALL)
+    # Disable some leveldb targets before we call FetchContent_MakeAvailable
+    # so they are configured correctly if it needs to fetch
+    set(LEVELDB_BUILD_TESTS OFF CACHE BOOL "LevelDB tests off")
+    set(LEVELDB_BUILD_BENCHMARKS OFF CACHE BOOL "LevelDB benchmarks off")
+
+    FetchContent_MakeAvailable(leveldb)
+
+    if (leveldb_FOUND)
+        # Globally installed leveldb needs Threads available as Threads::Threads as discovered by `find_package`
+        find_package(Threads REQUIRED)
+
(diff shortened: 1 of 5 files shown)


phated:
It looks like the answer to "what cmake files in docker" is 3.16.3 due to the test failures.

cpp/src/aztec/stdlib/merkle_tree/CMakeLists.txt
```diff
@@ -1,26 +1,49 @@
...
 
-barretenberg_module(stdlib_merkle_tree stdlib_primitives stdlib_blake3s stdlib_pedersen)
\ No newline at end of file
+        target_link_libraries(stdlib_merkle_tree PRIVATE leveldb::leveldb)
+        target_link_libraries(stdlib_merkle_tree_objects PRIVATE leveldb::leveldb)
+    else()
+        # FetchContent_MakeAvailable calls FetchContent_Populate if `find_package` is unsuccessful
+        # so these variables will be available if we reach this case
+        set_property(DIRECTORY ${leveldb_SOURCE_DIR} PROPERTY EXCLUDE_FROM_ALL)
+        set_property(DIRECTORY ${leveldb_BINARY_DIR} PROPERTY EXCLUDE_FROM_ALL)
```
dbanks12, line 30: So here are we saying, "if leveldb needs to be compiled from scratch here, exclude its source files and binary targets from 'all'"? If so, I see why we would want to do that for binary dir, but why for source dir?

cpp/src/aztec/stdlib/merkle_tree/CMakeLists.txt
```diff
@@ -1,26 +1,49 @@
...
+        target_link_libraries(stdlib_merkle_tree PRIVATE leveldb)
+        target_link_libraries(stdlib_merkle_tree_objects PRIVATE leveldb)
+        if(TESTING)
+            target_link_libraries(stdlib_merkle_tree_tests PRIVATE leveldb)
+            target_link_libraries(stdlib_merkle_tree_test_objects PRIVATE leveldb)
+        endif()
+        if(BENCHMARKS)
+            target_link_libraries(stdlib_merkle_tree_bench PRIVATE leveldb)
+            target_link_libraries(stdlib_merkle_tree_bench_objects PRIVATE leveldb)
+        endif()
```
phated: This seems fragile. I was thinking that `barretenberg_module` could set some list value (maybe `stdlib_merkle_tree_TARGETS`) that can be iterated, which would allow it to contain targets based on the various flags.

dbanks12 (approved):
looks good to me

cpp/src/aztec/stdlib/merkle_tree/CMakeLists.txt
phated: I was just mirroring the original logic of `add_subdirectory(${leveldb_SOURCE_DIR} ${leveldb_BINARY_DIR} EXCLUDE_FROM_ALL)` which was marking both directories as excluded. I assumed both of these were "added" by default with `FetchContent_Populate`. Happy to change if you know otherwise!

dbanks12: Ah I didn't see the original code above. Got it, looks good.

PR #140, opened 2023-02-09
Ad/Remove `waffle`

### LINKED ISSUE #128: Deprecate `waffle`...
Maybe also move some shared stuff into `bonk` namespace. Maybe also give some better names to things (e.g., `waffle::StandardComposer` can become `plonk::StandardComposer` and `honk::StandardHonkComposer` can become `honk::StandardComposer`).

arijitdutta67:
# Description

This PR removes the usage of the namespace `waffle` which is an obsolete term. Instead, we use the namespace `bonk` (both `PLONK` and `HONK`, credits to Cody for coming up with the term) where the stuffs are shared between `HONK` and `PLONK`, and the namespace `plonk` where stuffs are related only to `PLONK`. 

# Checklist:


cpp/src/aztec/benchmark/plonk_bench/plonk.bench.cpp
```diff
@@ -28,15 +28,15 @@ void generate_test_plonk_circuit(waffle::StandardComposer& composer, size_t num_
     }
 }
 
-waffle::Prover provers[NUM_CIRCUITS];
-waffle::Verifier verifiers[NUM_CIRCUITS];
-waffle::plonk_proof proofs[NUM_CIRCUITS];
+plonk::Prover provers[NUM_CIRCUITS];
+plonk::Verifier verifiers[NUM_CIRCUITS];
+plonk::plonk_proof proofs[NUM_CIRCUITS];
```
codygunton: What do you think of making this `plonk::proof`?

cpp/src/aztec/honk/circuit_constructors/standard_circuit_constructor.hpp
```diff
@@ -14,7 +14,7 @@ inline std::vector<std::string> standard_selector_names()
 class StandardCircuitConstructor : public CircuitConstructorBase<STANDARD_HONK_WIDTH> {
   public:
     // TODO: replace this with Honk enums after we have a verifier and no longer depend on plonk prover/verifier
-    static constexpr waffle::ComposerType type = waffle::ComposerType::STANDARD_HONK;
+    static constexpr plonk::ComposerType type = plonk::ComposerType::STANDARD_HONK;
```
codygunton, line 17: This should be `honk`.

cpp/src/aztec/honk/proof_system/prover.cpp
```diff
@@ -291,7 +291,7 @@ template <typename settings> void Prover<settings>::execute_relation_check_round
 {
     // queue.flush_queue(); // NOTE: Don't remove; we may reinstate the queue
 
-    using Multivariates = sumcheck::Multivariates<barretenberg::fr, waffle::STANDARD_HONK_TOTAL_NUM_POLYS>;
+    using Multivariates = sumcheck::Multivariates<barretenberg::fr, bonk::STANDARD_HONK_TOTAL_NUM_POLYS>;
```
codygunton, line 292: Seems like this should be in `honk`, no?

cpp/src/aztec/plonk/composer/composer_base.cpp
```diff
@@ -2,7 +2,7 @@
 #include <proof_system/proving_key/proving_key.hpp>
 #include <plonk/proof_system/utils/permutation.hpp>
 
-namespace waffle {
+namespace bonk {
```
codygunton: Shouldn't this be `plonk`?

cpp/src/aztec/plonk/composer/standard_composer.test.cpp
```diff
@@ -13,65 +13,65 @@ auto& engine = numeric::random::get_debug_engine();
 
 TEST(standard_composer, base_case)
```
codygunton, line 14: This file seems like a good candidate for a `using namespace plonk` or otherwise putting these tests themselves in the `plonk` namespace. Then we don't have to qualify all of the names--it really does seem redundant here since we're testing stuff in the `plonk` namespace.

codygunton: Similarly for other test files that follow. But maybe I'm overlooking something?

cpp/src/aztec/plonk/proof_system/constants.hpp
```diff
@@ -1,13 +1,13 @@
...
 
-namespace waffle {
+namespace plonk {
 
 enum ComposerType {
     STANDARD,
     TURBO,
     PLOOKUP,
-    STANDARD_HONK,
+    STANDARD_HONK, // Todo(Arijit): We should replace STANDARD_HONK outside plonk or the namespace should be bonk
```
codygunton, line 10: This type is used to determine how the standard library constructs circuits using the given composer, and since we will be making a Honk composer that uses the `PLOOKUP` circuits, as well, it is better to make an enum `ComposerType` in `honk` and then move `STANDARD_HONK` into that enum and rename simply to `STANDARD`. That way we have `honk::ComposerType::Standard` and `plonk::ComposerType::Standard`.

cpp/src/aztec/honk/proof_system/prover.cpp
codygunton: Actually, I see your reasoning. There will be a different polynomial manifest used by Honk, and Plonk will probably continue to use the polynomial manifests, so we will probably put polynomial_manifest.hpp in the `plonk` namespace. For now I think it's fine to leave things as you've done them. But will you leave a TODO describing what I just wrote in polynomial_manifest.hpp?

codygunton: More context: `PolynomialIndex` is not used by Honk at all.

cpp/src/aztec/proof_system/types/polynomial_manifest.hpp
```diff
@@ -4,8 +4,8 @@
 #include <vector>
 #include <string>
 #include <plonk/proof_system/constants.hpp>
-
-namespace waffle {
+using namespace plonk;
```
codygunton: What is this `using` doing?

codygunton (dismissed):
Thanks for doing this! Most of it looks great. I just have a few comments.

cpp/src/aztec/benchmark/plonk_bench/plonk.bench.cpp
arijitdutta67: Yes, makes sense. Changed.

cpp/src/aztec/plonk/composer/standard_composer.test.cpp
arijitdutta67: A great observation. Done.

cpp/src/aztec/proof_system/types/polynomial_manifest.hpp
arijitdutta67: Nice spot. Removed.

cpp/src/aztec/honk/proof_system/prover.cpp
arijitdutta67: Added TODO

cpp/src/aztec/plonk/composer/composer_base.cpp
arijitdutta67: This is problematic because [here](https://github.com/AztecProtocol/barretenberg/blob/fdc97b6ff57d87161ed047a785ff98167c88deae/cpp/src/aztec/plonk/composer/composer_base.hpp#L14) we are declaring `proving_key` and `verification_key`, which are under the `bonk` namespace. If I change the namespace to `plonk`, conflicts arise which I could not resolve. For now, I am adding a TODO for this. Please let me know if you see a way out.

cpp/src/aztec/plonk/proof_system/constants.hpp
arijitdutta67: This makes sense. However, when I tried it, it seems like we need to compare between two different `enum` giving error ([-Wdeprecated-enum-compare](https://clang.llvm.org/docs/DiagnosticsReference.html#wdeprecated-enum-compare)). Instead, I created a constants file [here](https://github.com/AztecProtocol/barretenberg/blob/ad/remove-waffle/cpp/src/aztec/proof_system/constants.hpp#L1). Hope this resolves the issue.

cpp/src/aztec/honk/circuit_constructors/standard_circuit_constructor.hpp
arijitdutta67: Keeping it as it is as I could not separate out the two composer types.

cpp/src/aztec/plonk/composer/composer_base.cpp
ledwards2225: We do not need a forward declaration of `proving_key` and `verification_key` in the `.hpp`. You can simply delete those two lines then change the namespace in the both cpp/hpp to plonk.

cpp/src/aztec/plonk/composer/composer_base.hpp
```diff
@@ -7,8 +7,8 @@
 #include <srs/reference_string/file_reference_string.hpp>
 
 using namespace bonk;
```
ledwards2225: I know you didn't add this line but please delete it. This is bad form an in this case entirely unnecessary.

cpp/src/aztec/plonk/composer/composer_base.hpp
```diff
@@ -7,8 +7,8 @@
...
-
-namespace waffle {
+// TODO (Arijit): Change the below namespace to plonk
+namespace bonk {
```
ledwards2225: Change to plonk (after deleting forward declaration of structs).

cpp/src/aztec/plonk/proof_system/constants.hpp
ledwards2225: The ideal would be to make this work as Cody described above. If that's not possible (or if the scope of doing it is too large, etc.) I would prefer leaving `plonk/proof_system/constants.hpp` and `honk/proof_system/constants.hpp` in place (with `STANDARD_HONK ` inside of the of the plonk enum). That way we have the ugly `plonk::ComposerType::STANDARD_HONK` in one or two places rather than hundreds of e.g. `bonk::ComposerType::PLOOKUP` that will have to be reverted to `plonk::ComposerType::PLOOKUP` in the near future. I think this basically amounts to leaving things as you had them prior to the commit [4e37d5a](https://github.com/AztecProtocol/barretenberg/pull/140/commits/4e37d5ae33c1af9249fa426670c855594eb56830).

ledwards2225 (changes requested):
LGTM - Just a couple more things to help alleviate future headache before we get this in!

arijitdutta67:
> LGTM - Just a couple more things to help alleviate future headache before we get this in!

Thanks @ledwards2225 for the nice suggestions. I adopted all of them.

arijitdutta67:
> Thanks for doing this! Most of it looks great. I just have a few comments.

Thanks @codygunton for all the great suggestions. I have mostly adopted them.

cpp/src/aztec/plonk/composer/composer_base.cpp
arijitdutta67: Works. Thanks.

cpp/src/aztec/plonk/proof_system/constants.hpp
arijitdutta67: Done as suggested.

ledwards2225 (approved):
This represents a large step in the right direction. Namely, searching `waffle` now returns zero results :). There will necessarily be lots of minor follow ons to tweak the `plonk`/`honk`/`bonk` namespaces, particularly as Honk evolves. Also, as much as is possible I think we should to do away with `using namespace <>` statements at the top of headers which is generally considered bad practice.

PR #155, opened 2023-02-16
chore!: Remove most GCC targets

phated:
# Description

As discussed in slack, this begins to remove the GCC targets for the project. Trying to make things conform to GCC standards is slowing down the team and we don't actively use it because GCC & Clang-compiled versions can't co-mingle.

Currently a draft because I need some help understanding the cross-compiling stuff:
1. Where is it used?
2. Where is it built?
3. Is it part of CI?
4. Did I move the stuff around correctly?

# Checklist:


dbanks12:
1. Not sure
2. Only `Dockerfile.x86_64-linux-gcc` that I know of
3. We build `Dockerfile.x86_64-linux-gcc` in CI, but do not run any tests against it
4. Looks mostly right to me. Why not also remove `arm64-linux-gcc` (Dockerfile and .cmake)?

phated:
Thanks @dbanks12 - I've removed the cross-compile stuff since we'll probably want to cross-compile with Clang going forward. That also allowed me to remove some pragmas and comments related to gcc

cpp/src/aztec/ecc/curves/bn254/scalar_multiplication/scalar_multiplication.test.cpp
```diff
@@ -510,19 +510,10 @@ TEST(scalar_multiplication, endomorphism_split)
...
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Warray-bounds"
-#endif
+
     fr k1{ (*k1_t).data[0], (*k1_t).data[1], 0, 0 };
     fr k2{ (*k2_t).data[0], (*k2_t).data[1], 0, 0 };
-#if !defined(__clang__) && defined(__GNUC__)
-#pragma GCC diagnostic pop
-#endif
+
```
adr1anh, line 516: Can we just keep this in? I'm fine with not officially supporting GCC and removing it from the CI, but this error was occurring in the alpine1.17 gcc 12 build. Also, the thing we're doing here is very tricky and I think the textual warning should stay.

cpp/CMakeLists.txt
```diff
@@ -72,12 +72,8 @@ if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
...
         message(WARNING "Clang <10 is not supported")
     endif()
-elseif(CMAKE_CXX_COMPILER_ID MATCHES "GNU")
-    if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS "10")
-        message(WARNING "GCC <10 is not supported")
-    endif()
-else() 
-    message(WARNING "Unsuported compiler, use Clang >10 or GCC >10")
+else()
+    message(WARNING "Unsuported compiler, use Clang >10")
```
adr1anh, line 76: If we're going to update our supported compilers, we should test whether the code actually compiles with the clang >10 stdlib as well.

adr1anh (commented):
Not really sure we should be removing all the dockerfiles. Even if we aren't using them now, they were a pain to setup at the time and they can just stay there.

phated:
Going to replace this by dropping the forced compilers and setting `CC` and `CXX` in the gcc docker containers (I think).

PR #162, opened 2023-02-17
feat!: Remove TOOLCHAIN logic and replace with CMake presets

phated:
# Description

Before these changes, Barretenberg was forcing specific C and C++ Compilers through the `-DTOOLCHAIN` variable. This is a flawed design because it doesn't allow consumers to set the `CC` or `CXX` environment variables to pick their compilers. Being able to set those values is extremely important for cross-compiling.

This PR removes all of that `TOOLCHAIN` logic and instead leverages [CMake Presets](https://cmake.org/cmake/help/v3.24/manual/cmake-presets.7.html) to provide the defaults we expect (such as clang, clang-15, gcc, gcc-10 and clang from Homebrew). Shoutout to @adr1anh for sharing the idea for presets which **massively** helped with this transition plan.

It also updates the `bootstrap.sh` file to detect and print which preset should be used. If you use cmake commands directly, you'll need to start using the related preset commands.

I've also updated the dockerfiles and docs to all reference these preset commands.

The `cmake/toolchains` directory still exists, but the files have been re-done as proper CMake Cross-Compile [Toolchain files](https://cmake.org/cmake/help/book/mastering-cmake/chapter/Cross%20Compiling%20With%20CMake.html#toolchain-files) which are really just supposed to set the target name and architecture, but can include extras like the i386 file.

I've tested this quite extensively on my Arm Mac but I need feedback for Linux, the dockerfiles, and the compare_honk_bench stuff (because it was non-standard).

# Checklist:


README.md
```diff
@@ -42,21 +38,24 @@ If you've installed the C++ Vscode extension you should configure it to format o
 
 Each module has its own tests. e.g. To build and run `ecc` tests:
 
-```
-cmake --build . --parallel --target ecc_tests
```
dbanks12, line 46: Are we removing `--parallel` everywhere just because it shouldn't be listed as a recommended/default option?

cpp/dockerfiles/Dockerfile.arm64-linux-gcc
```diff
@@ -1,6 +1,8 @@
+# Unused & not maintained. This probably won't work without changes.
+
 FROM aztecprotocol/crosstool-ng-arm64:latest
 WORKDIR /usr/src/barretenberg/cpp
 COPY . .
-RUN mkdir build && cd build && cmake -DTOOLCHAIN=arm64-linux-gcc .. && cmake --build . --parallel
+RUN cmake --toolchain ./cmake/toolchains/aarch64-linux.cmake --preset gcc && cmake --build --preset gcc
```
dbanks12, line 6: Why not use a preset here?

cpp/CMakePresets.json
```diff
@@ -0,0 +1,213 @@
+{
```
dbanks12, line 1: Very clean. I like this a lot.

dbanks12:
Looks good to me. Just looking for clarification on a couple of things mentioned in my comments.

README.md
phated: `--parallel` doesn't work with presets. I'm pretty sure that's because it's a legacy option and `--jobs N` is supposed to be used; however I didn't try combining presets with `--jobs`. Instead, we set `jobs: 0` on all our presets so they are running at maximum concurrency.

cpp/dockerfiles/Dockerfile.arm64-linux-gcc
phated: It's using the gcc preset and adding the cross-compiling toolchain file. We could have presets for all cross-compilation, but that's usually a matrix, so it'd drastically increase the number of presets we have to maintain.

dbanks12 (approved):

README.md
phated: Slight correction, it seems that `--parallel` works with presets, but requires a very specific ordering (e.g. you need to specify `--build --preset default --parallel` otherwise the preset isn't picked up correctly). Since we set `jobs: 0` in all the presets, we shouldn't recommend the flag because it just makes the command invocation more complex.

charlielye (approved):
Skimmed. But looks like a great step in the right direction 👍

adr1anh (approved):
LGTM

adr1anh:
To address @codygunton's issue with GCC12, it seems like we can instead set the generator to Ninja in the CMakePresets.json rather than the VSCode workspace settings. 

```
    {
      "name": "default",
      "displayName": "Build with Clang",
      "description": "Build with globally installed Clang",
      "binaryDir": "build",
      "generator": "Ninja",
      "environment": {
        "CC": "clang",
        "CXX": "clang++"
      }
    },
```

When testing on the mainframe, my build were being parallelized after performing this change. 

On another note, switching configure preset requires the CMake cache to be rebuilt. The option is "CMake: Delete cache and reconfigure" in VSCode. This could be solved by having a separate build folder for each configuration, though that might be more annoying to setup. (i.e. `build-gcc` `build-gcc`, with `build` being the folder for the default configuration with clang.)

adr1anh:
I've added a commit which sets the generator in the `default` preset to `Ninja`. The second commit tells CMake to generate a `compile_commands.json` file that helps the `clangd` language server understand how the file being edited is compiled. It required an extra argument in the config to suppress a warning from a GCC-specific compile flag.

EDIT: I had to modify the docker files to download ninja as well.

codygunton:
VS Code integration is still kind of funky, e.g. building with wasm preset using CMake Tools is _extremely_ slow, but it works for most common workflow, so I'm going to merge.

PR #167, opened 2023-02-18
add missing to_nt converts

iAmMichaelConnor:
# Description

Adds a couple of conversion functions.

### DIFF
--- a/cpp/src/aztec/stdlib/types/convert.hpp
+++ b/cpp/src/aztec/stdlib/types/convert.hpp
@@ -152,7 +152,8 @@ template <typename Composer> typename NT::grumpkin_point to_nt(typename CT<Compo
 
 template <typename Composer> typename NT::bn254_point to_nt(typename CT<Composer>::bn254_point const& e)
 {
-    return NT::bn254_point{ e.x.get_value(), e.y.get_value() };
+    // TODO: cannot instantiate a field from a uint512_t?
+    return NT::bn254_point{ e.x.get_value().lo, e.y.get_value().lo };
 };
 
 template <typename Composer>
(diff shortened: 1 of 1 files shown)


cpp/src/aztec/stdlib/types/convert.hpp
```diff
@@ -152,7 +152,8 @@ template <typename Composer> typename NT::grumpkin_point to_nt(typename CT<Compo
 
 template <typename Composer> typename NT::bn254_point to_nt(typename CT<Composer>::bn254_point const& e)
 {
-    return NT::bn254_point{ e.x.get_value(), e.y.get_value() };
+    // TODO: cannot instantiate a field from a uint512_t?
+    return NT::bn254_point{ e.x.get_value().lo, e.y.get_value().lo };
```
iAmMichaelConnor: @suyash67 this is the first conversion function which has seemed strange to me. This is a conversion function for a bn254 point, from a circuit type to a native type. The circuit type is a non-native field (big field), and its underlying value (retrieved with `get_value()`) is a `uint512_t` (which itself has a `.hi` and `.lo` component; each of 256-bits). I thought it strange that a 254-bit field would be represented by a 512-bit number. Anyway, there isn't a constructor of the _native_ type (which is being called on line 156) which takes a `uint512_t`, so I've just taken the `lo` value, for now.

suyash67: The reason why `get_value()` outputs a `uint512_t` instead of `uint256_t` is because we want to able to support `biggroup`s with `x, y` values of upto 272 bits ($68 \times 4 = 272$). In case of BN254, since the `x, y` coordinates are just 254-bit numbers, it should be fine to just take the `lo` value.

suyash67: Actually, there's an inbuilt `get_value()` function on `CT<Composer>::bn254_point` that converts it to the native form. (Does exactly what you've done so we should probably use that).

suyash67 (approved):
looks good to me.

cpp/src/aztec/stdlib/types/convert.hpp
iAmMichaelConnor: Thanks for spotting that!

PR #172, opened 2023-02-21
feat: Add equality and serialization to poly_triple

phated:
# Description

Noir needs to be able to serialize and deserialize the poly_triple struct. It also needs to be able to check equality of these.

# Checklist:


cpp/src/aztec/proof_system/composer/composer_base.hpp
```diff
@@ -56,8 +57,49 @@ struct poly_triple {
...
+            lhs.a == rhs.a &&
+            lhs.b == rhs.b &&
+            lhs.c == rhs.c &&
+            lhs.q_m == rhs.q_m &&
+            lhs.q_l == rhs.q_l &&
+            lhs.q_r == rhs.q_r &&
+            lhs.q_o == rhs.q_o &&
+            lhs.q_c == rhs.q_c;
+        // clang-format on
+    }
```
adr1anh: Is it possible to replace this with `bool operator==(poly_triple const& lhs, poly_triple const& rhs) = default;`?

kevaundray: If the minimum version of CPP supported is C++20, then yep we can use default comparisons here instead

phated: Checking the CMake file, it looks like C++20 is required, so I can make that change.

adr1anh (approved):
LGTM

PR #173, opened 2023-02-21
feat: Add `signature_verification_result` to schnorr stdlib

phated:
# Description

Noir needs `verify_signature` in the schnorr stdlib to return a `bool_t` indicating whether the signature was valid or not.

Currently a draft because I tried to update @kevaundray's code to add a constraint but I have no idea if I did it correctly.

# Checklist:

### DIFF
--- a/cpp/src/aztec/stdlib/encryption/schnorr/schnorr.cpp
+++ b/cpp/src/aztec/stdlib/encryption/schnorr/schnorr.cpp
@@ -1,4 +1,5 @@
 #include "schnorr.hpp"
+#include <array>
 #include <crypto/pedersen/pedersen.hpp>
 #include <ecc/curves/grumpkin/grumpkin.hpp>
 #include <stdlib/hash/blake2s/blake2s.hpp>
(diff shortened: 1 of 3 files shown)


cpp/src/aztec/stdlib/encryption/schnorr/schnorr.cpp
```diff
@@ -293,8 +293,9 @@ void verify_signature(const byte_array<C>& message, const point<C>& pub_key, con
 
     field_t<C> output_hi(output.slice(0, 16));
     field_t<C> output_lo(output.slice(16, 16));
-    output_lo.assert_equal(sig.e_lo, "verify signature failed");
```
codygunton, line 296: `assert_equal` imposes a constraint (a copy constraint, which is free in the sense of adding no gates) that two things are equal. Your change breaks the model of this function. As it was before, the circuit would be supplied with non-satisfying witnesses when the signature verification fails, i.e., the prover would produce a proof that would fail to verify. With your change, the prover would supply a passing proof that the signature failed to verify. This is why some of the tests don't pass. To resolve, you could put the `assert_equal`'s back in place.

codygunton: @kevaundray Are you proposing to change the semantics here? Is there a use case for proving that a signature does _not_ verify?

codygunton (changes requested):

cpp/src/aztec/stdlib/encryption/schnorr/schnorr.cpp
zac-williamson: One thing to note is that if the current behaviour has a minor issue:

It is not currently possible to distinguish between the signature being invalid, and a Prover deliberately providing incorrect witness values.  

This isn't a problem for the Aztec Connect use case, but it may be an issue for a Noir app where proof computation is being delegated to a third party (i.e. you can't distinguish between "I can't make a valid proof because you gave me a bad signature" and "I *won't* make a valid proof because I'm dishonest"

phated: I've re-added the `assert_equal` calls, which I mistakenly removed when I made the change to the logic to add actual constraints (I was confused by the "equal" name and "==" (equal operator) and thought they were the same thing).

I'm pretty sure Noir just needs to be able to know the that something was valid/invalid via return type (and if I understand correctly, that needs to be constrained)

codygunton: @zac-williamson so perhaps remove the `assert_equal`s and rename to `verify_signature_unsafe` or `prove_signature_verification_result` or something like this?

zac-williamson (dismissed):
I can see how it would be useful to have a verification method that does not assert but instead returns a boolean. It gives more flexibility to developers and makes it easier to evaluate branching paths where one wishes to conditionally verify a signature.

We have a lot of tests and circuit code that relies on the signature verification algorithm to produce unsatisfiable constraints if the signature is invalid.

I think the fastest path forward is to create a new independent method that returns a bool_t, which does not interfere with existing tests.

codygunton:
Rebased and resolved conflicts with `waffle` removal PR.

codygunton (approved):

codygunton:
@zac-williamson looks good? Merging is blocked until you approve :innocent:

codygunton:
^ rebase

PR #175, opened 2023-02-22
fix: Check for wasm-opt during configure & run on post_build

phated:
# Description

This PR makes the wasm-opt dependency more explicit. It is now referenced as a dependency since the wasm build fails without it. I've also added a `find_program` call to ensure it is installed during the configure phase for a wasm build; otherwise you'd get to the very end and it would crash.

I've changed the executable target to be `barretenberg.wasm` and then used the [`TARGET` variant](https://cmake.org/cmake/help/latest/command/add_custom_command.html#build-events) of `add_custom_command` so the wasm-opt command isn't run if `barretenberg.wasm` is already built. The previous way of doing this created a target that was always stale.

This should enable adding the wasm binary as an installation target, which is not possible with custom targets.

# Checklist:

### DIFF
--- a/cpp/src/aztec/CMakeLists.txt
+++ b/cpp/src/aztec/CMakeLists.txt
@@ -62,7 +62,7 @@ if(WASM)
     # that as functions in 'env' should be implemented in JS itself.
     # It turns out that just explicitly telling the wasm module which object files to include was easiest.
     add_executable(
-        barretenberg-step1.wasm
+        barretenberg.wasm
         $<TARGET_OBJECTS:srs_objects>
         $<TARGET_OBJECTS:numeric_objects>
         $<TARGET_OBJECTS:crypto_sha256_objects>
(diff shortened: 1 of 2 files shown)


cpp/src/aztec/CMakeLists.txt
```diff
@@ -88,22 +88,23 @@ if(WASM)
...
+        TARGET barretenberg.wasm
+        POST_BUILD
+        COMMAND wasm-opt "$<TARGET_FILE:barretenberg.wasm>" -O2 --asyncify -o "$<TARGET_FILE:barretenberg.wasm>"
         VERBATIM
     )
 
-    add_custom_target(
-        barretenberg.wasm
-        DEPENDS ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/barretenberg.wasm
-    )
```
dbanks12, line 106: Here we reuse the same target `barretenberg.wasm`. Does this basically say the following?
1. Create an executable target `barretenberg.wasm` that depends on some objects
2. Give that same target some link options
3. Add a custom command for that same target that runs `wasm-opt` (or errors if absent)

dbanks12:
Left a comment. Otherwise looks good to me. This is good cleanup.

cpp/src/aztec/CMakeLists.txt
phated: Sorry for the delay.

You are mostly correct, with slight nuance.
1. Fail if wasm-opt doesn't exist; this now happens during configuration instead of at the end of the build.
2. Create an executable target `barretenberg.wasm` that depends on some objects. (This is the thing that can be installed).
3. Give that same target some link options
4. Add a custom command for that same target that runs wasm-opt on the artifact as a POST_BUILD, so we know the artifact exists and we just optimize and replace the artifact in place. This avoids re-computation if nothing changed and makes installation easier.

dbanks12: Ahh okay. So....
1. Since wasm-opt was part of the custom command previously, it wouldnt fail until build. Got it.
2. Got it.
3. Got it.
4. Ah, I didn't notice the `POST_BUILD`. This makes perfect sense, thanks :+1:

PR #177, opened 2023-02-23
Remove testing methods from Honk + clean up

adr1anh:
# Description

- Unify the interfaces for relations and rework tests to not use `_testing` methods anymore.
- Remove `#pragma GCC` and fixed errors.
- Fix `Univariate` operator interfaces to be `const` correct, and return references when required [ref](https://en.cppreference.com/w/cpp/language/operator_assignment).

# Checklist:

### DIFF
--- a/cpp/src/aztec/honk/composer/standard_honk_composer.test.cpp
+++ b/cpp/src/aztec/honk/composer/standard_honk_composer.test.cpp
@@ -358,27 +358,27 @@ TEST(StandardHonkComposer, SumcheckRelationCorrectness)
     // get the transposition.
     // Ex: polynomial_spans[3][i] returns the i-th coefficient of the third polynomial
     // in the list below
-    std::array<std::span<const fr>, num_polynomials> univariate_array;
+    std::array<std::span<const fr>, num_polynomials> evaluations_array;
 
     using POLYNOMIAL = bonk::StandardArithmetization::POLYNOMIAL;
-    univariate_array[POLYNOMIAL::W_L] = prover.wire_polynomials[0];
-    univariate_array[POLYNOMIAL::W_R] = prover.wire_polynomials[1];
-    univariate_array[POLYNOMIAL::W_O] = prover.wire_polynomials[2];
-    univariate_array[POLYNOMIAL::Z_PERM] = z_perm_poly;
-    univariate_array[POLYNOMIAL::Z_PERM_SHIFT] = z_perm_poly.shifted();
-    univariate_array[POLYNOMIAL::Q_M] = prover.key->polynomial_cache.get("q_m_lagrange");
-    univariate_array[POLYNOMIAL::Q_L] = prover.key->polynomial_cache.get("q_1_lagrange");
-    univariate_array[POLYNOMIAL::Q_R] = prover.key->polynomial_cache.get("q_2_lagrange");
-    univariate_array[POLYNOMIAL::Q_O] = prover.key->polynomial_cache.get("q_3_lagrange");
-    univariate_array[POLYNOMIAL::Q_C] = prover.key->polynomial_cache.get("q_c_lagrange");
-    univariate_array[POLYNOMIAL::SIGMA_1] = prover.key->polynomial_cache.get("sigma_1_lagrange");
-    univariate_array[POLYNOMIAL::SIGMA_2] = prover.key->polynomial_cache.get("sigma_2_lagrange");
-    univariate_array[POLYNOMIAL::SIGMA_3] = prover.key->polynomial_cache.get("sigma_3_lagrange");
-    univariate_array[POLYNOMIAL::ID_1] = prover.key->polynomial_cache.get("id_1_lagrange");
-    univariate_array[POLYNOMIAL::ID_2] = prover.key->polynomial_cache.get("id_2_lagrange");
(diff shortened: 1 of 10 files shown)


cpp/src/aztec/honk/composer/standard_honk_composer.test.cpp
```diff
@@ -336,136 +335,81 @@ TEST(standard_honk_composer, test_check_sumcheck_relations_correctness)
...
+
+    constexpr size_t num_polynomials = bonk::StandardArithmetization::NUM_POLYNOMIALS;
+    // Compute grand product polynomial (now all the necessary polynomials are inside the proving key)
+    polynomial z_perm_poly = prover.compute_grand_product_polynomial(beta, gamma);
+
+    // Create an array of spans to the underlying polynomials to more easily
+    // get the transposition.
+    // Ex: polynomial_spans[3][i] returns the i-th coefficient of the third polynomial
+    // in the list below
+    std::array<std::span<const fr>, num_polynomials> univariate_array;
```
ledwards2225: Any particular reason why we're calling these univariates? I suppose the relations are a bit representation agnostic but for sumcheck at large we think of them as multivariates

cpp/src/aztec/honk/sumcheck/relations/relation.test.cpp
```diff
@@ -19,11 +18,12 @@ template <class FF> class SumcheckRelation : public testing::Test {
   public:
     template <size_t t> using Univariate = Univariate<FF, t>;
     template <size_t t> using UnivariateView = UnivariateView<FF, t>;
+    static constexpr size_t FULL_RELATION_LENGH = 5;
```
ledwards2225, line 21: Definitely understand if you don't want to do this in this PR but just some thoughts: This `FULL_RELATION_LENGH` is necessarily tied to the actual full relation length so the hard coding is kind of brittle and has tripped me up on a few occasions. It would be nice to compute this programmatically instead then update`compute_mock_extended_edges` to create appropriately sized mock univariates. (We could do this by either directly generating the full sized univariate or just defining length 2 univariates then extending them appropriately).

cpp/src/aztec/honk/sumcheck/polynomials/univariate.hpp
```diff
@@ -309,4 +295,67 @@ template <class Fr, size_t view_length> class UnivariateView {
         return os;
     }
 };
+
```
ledwards2225, line 298: These methods don't seem to be used anywhere. You see some use for them in the future?

cpp/src/aztec/honk/composer/standard_honk_composer.test.cpp
```diff
@@ -336,136 +335,81 @@ TEST(standard_honk_composer, test_check_sumcheck_relations_correctness)
...
-    }
-    // Check all relations at all indices
-    for (size_t i = 0; i < prover.key->circuit_size; i++) {
-        round.accumulate_relation_univariates_testing(
-            sumcheck_typed_polynomial_vector[i], results, { beta, gamma, public_input_delta });
-        EXPECT_EQ(std::get<0>(results), ArithmeticUnivariate(0));
-        EXPECT_EQ(std::get<1>(results), GrandProductComputationUnivariate(0));
-        EXPECT_EQ(std::get<2>(results), GrandProductInitializationUnivariate(0));
+        // // Compute an array containing all the evaluations at a given row i
+        std::array<fr, num_polynomials> transposed;
```
ledwards2225: Isn't this operation that you created `transposed_univariate_array_at` for?

ledwards2225 (approved):
Overall looks excellent - lots of great cleanup. Just a few minor questions and comments in the spirit of nit picking :)

cpp/src/aztec/honk/sumcheck/polynomials/univariate.hpp
adr1anh: The main use case is to convert an array of `Univariate` into `UnivariateView`. The main use case would be to let Sumcheck decide the required degree of the relation evaluation, rather than hardcoding it inside the relation. The `_aux` version could also be used to create an array of only the polynomials required by the relation, and it could help us implement the optimization where we extend each edge only up to the maximum degree that is required over all relations (for example, `L_LAST` only needs degree 3). 

I'll save this comment as the context might help.

cpp/src/aztec/honk/sumcheck/relations/relation.test.cpp
adr1anh: Agreed that this would be nice. I think that we can get this done as part of Arijit's PR since he introduced a lot more functionality to the way the test is performed. Left a more detailed TODO for now

cpp/src/aztec/honk/composer/standard_honk_composer.test.cpp
adr1anh: The problem was that that function only worked over `Univariate` since it doesn't have the `operator[]` defined. I moved it to the test it was being used it. In retrospect it may have been a bit too generic and the following loop is simple enough.

cpp/src/aztec/honk/composer/standard_honk_composer.test.cpp
adr1anh: Accidentally copied the previous name without thinking. Will rename it to something better!

adr1anh:
Fixes #182  btw

PR #185, opened 2023-02-24
feat: Add installation targets for libbarretenberg, wasm & headers

### LINKED ISSUE #176: Should have install targets
This project currently only builds to a "binary dir" (as cmake calls them) but it should also have install targets so you can do `cmake --install . --prefix "/usr/local"`

phated:
# Description

Closes #176

This adds install targets for `libbarretenberg.a`, `barretenberg.wasm` and all the headers. It uses an "interface library" to collect headers as the subdirectories call the `barretenberg_module` function. I couldn't come up with a better way to collect the headers for installation because they exist in the subdirectories (instead of something like an `include/` directory where the public headers exist, such as https://github.com/google/benchmark/tree/main/include/benchmark), which means that we promote *all* header files instead of just public ones.

We should open a follow up issue to differentiate public from private headers.

With installation targets, we can now do `cmake --install build --prefix "/usr/local"` and it'll install like a normal C++ library.

I also added the `INSTALL_BARRETENBERG` option in emulation of googletest and benchmark because using `FetchContent_MakeAvailable` will add them as install targets. You can see how the `BENCHMARK_ENABLE_INSTALL` and `INSTALL_GTEST` are turned off now too.

This is the minimal setup we need to have installation targets, but many more improvements can be made so projects can consume this via other CMake builds, pkg-config, etc. These improvements are currently unnecessary for Noir to use barretenberg, but I can create follow up issues.

# Checklist:

### DIFF
--- a/cpp/src/aztec/CMakeLists.txt
+++ b/cpp/src/aztec/CMakeLists.txt
@@ -55,6 +55,8 @@ if(BENCHMARKS)
     add_subdirectory(benchmark)
 endif()
 
+include(GNUInstallDirs)
+
 if(WASM)
     # Well, this is awkward. We can't build a wasm module by just linking to the libraries as that produces, nothing.
     # There are a couple of other ways to avoiding listing all the object files here and leveraging the dependency
@@ -105,6 +107,10 @@ if(WASM)
         VERBATIM
     )
 
+    if(INSTALL_BARRETENBERG)
+        install(TARGETS barretenberg.wasm DESTINATION ${CMAKE_INSTALL_BINDIR})
+    endif()
+
     # For use when compiling dependent cpp projects for WASM
     message(STATUS "Compiling all-in-one barretenberg WASM archive")
     add_library(
@@ -131,7 +137,6 @@ if(WASM)
         $<TARGET_OBJECTS:stdlib_aes128_objects>
         $<TARGET_OBJECTS:stdlib_merkle_tree_objects>
     )
-
 else()
     # For use when compiling dependent cpp projects
     message(STATUS "Compiling all-in-one barretenberg archive")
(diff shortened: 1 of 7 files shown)


phated:
I found a couple more things that needed to change to install all headers got installed:
1. `*.tcc` files are included by headers
2. We needed to collect the common headers

codygunton (approved):
Looks good to me. Could you add something to the readme about how to install?

cpp/src/aztec/common/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+# Collect our common/*.hpp files and include in installation
+file(GLOB_RECURSE HEADER_FILES *.hpp)
+target_sources(
+    barretenberg_headers
+    INTERFACE
+    FILE_SET HEADERS
+    FILES ${HEADER_FILES}
+)
```
dbanks12, line 8: Is this only necessary here because common is not a `barretenberg_module`?

cpp/src/aztec/CMakeLists.txt
```diff
@@ -160,4 +166,13 @@ else()
...
-endif()
\ No newline at end of file
+
+    if(INSTALL_BARRETENBERG)
+        install(
+            TARGETS barretenberg barretenberg_headers
+            EXPORT barretenbergTargets
+            ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
+            FILE_SET HEADERS DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
+        )
```
dbanks12, line 176: Could you add a comment explaining what this is doing (especially for EXPORT and each DESTINATION arg). I had to look it up and I'm sure others will too. Still not 100% sure I know what this command is doing after doing a bit of digging.

cpp/src/aztec/common/CMakeLists.txt
phated: Correct!

cpp/src/aztec/CMakeLists.txt
phated: Will do in a follow up PR.

PR #200, opened 2023-03-06
fix: Ensure barretenberg provides headers that Noir needs

phated:
# Description

This adds headers for the `c_binds` files that Noir binds to. It also exposes `pedersen__commit` which is something we converted to a public function on Pedersen, and changes the `new_pippenger` "points" argument to a `uint8_t` type so it is easier to bind to.

# Checklist:


cpp/src/aztec/crypto/schnorr/c_bind.hpp
```diff
@@ -0,0 +1,39 @@
...
+WASM_EXPORT void compute_public_key(uint8_t const* private_key, uint8_t* public_key_buf);
+WASM_EXPORT void negate_public_key(uint8_t const* public_key_buffer, uint8_t* output);
+
+WASM_EXPORT void construct_signature(
+    uint8_t const* message, size_t msg_len, uint8_t const* private_key, uint8_t* s, uint8_t* e);
+
+WASM_EXPORT bool verify_signature(
+    uint8_t const* message, size_t msg_len, uint8_t const* pub_key, uint8_t const* sig_s, uint8_t const* sig_e);
+
+WASM_EXPORT void multisig_create_multisig_public_key(uint8_t const* private_key, uint8_t* multisig_pubkey_buf);
```
kevaundray, line 16: We can remove the functions prefixed with `multisig`

phated: That's unrelated to this PR though. I'm just adding headers for the functions that exist in `c_bind.cpp` (not sure why these headers were removed when moving from aztec-connect).

kevaundray: Those multisig functions are only used for aztec_connect and never by Noir

kevaundray: (Don't think they will ever be used by Aztec 3 either)

phated: Let's open an issue and get in front of people before we delete things from the codebase and break someone we might not be thinking of. This is just adding signatures for things that already exist.

kevaundray: I don't understand why its breaking since it doesn't exist in this codebase as of now and Noir should not be doing anything related to aztec-connect's multisig functionality.  CC @codygunton or @charlielye to see if these bindings are needed elsewhere, maybe for Aztec 3

phated: @kevaundray These functions **do** exist in the `c_bind.cpp` file. I am just adding headers so rust-bindgen can be used against the headers. By removing them from the header, they won't be visible to other code. We should consider removing the type signature **AND** the implementation in a PR that is specifically about doing that.

kevaundray: Hmm thats a good point about the .cpp --  though they are still not needed by Noir (I was reviewing this PR with the lens of anything we add will be needed by Noir) -- though since they already exist in the cpp, I don't feel as strongly about not including their header files

ledwards2225 (approved):
Looks fine to me. Can't speak to why these were removed originally. None of these c_binds are used by the Honk team (including, for future ref, the `multisig_*` functions).

Issue #192, opened 2023-02-27
Formatting git hook breaks `git add -p` workflows


phated:
Since the git commit hook to format code runs on and commits all files that are changed, if you try to use `git add -p` to stage chunks of a file, the entire file gets committed.

This is really bad for workflows that make a bunch of changes and then chop them up into digestible commits.

PR #201, opened 2023-03-06
feat: Improve barretenberg headers

phated:
# Description

This PR does a few things:
1. Renames `cpp/src/aztec` to `cpp/src/barretenberg`, which will align the installed headers to match the `libbarretenberg` library that will get installed by #185
2. Moves `cpp/src/aztec/CMakeLists.txt` to `cpp/src/CMakeLists.txt` which allows all header files to be prefixed with the `barretenberg` namespace. This ensures that header includes when building the project are the same as when installed.
3. Updates all the header includes to use that `barretenberg/` prefix.
4. Updates all local includes to use double-quotes instead of angle brackets. Generally, angle-brackets should only be used for system libraries and double-quotes should be used for user-defined headers. This also surfaces a lot of unnecessary or incorrect includes via clangd but I didn't want to change any of those in this PR.
5. Lots of clang-format changes since I touched files and it updates the entire file whenever one is staged.

# Checklist:


dbanks12:
Looks mostly good to me. We should ping the crypto team letting them know about this PR because it will likely break people's branches. But I do think this is a good change.

It might make sense to move some of the C++ formatting changes into a separate PR, but I don't feel strongly either way.

phated:
The formatting changes happened because I moved the files and the git hook applies to all files that are staged (see #192). I think it's fine to format them so all formatting is consistent as per the git hook.

PR #205, opened 2023-03-07
IPA uses srs

### LINKED ISSUE #163: IPA uses SRS
From https://github.com/AztecProtocol/barretenberg/pull/31#pullrequestreview-1240637105: the function `commit`, `ipa_prove` and `ipa_verify` should take in an SRS object instead of a vector of generators (and have this SRS class load the required points from disk like our KZG SRS)

### LINKED ISSUE #164: Share `pippenger_runtime_state`.
From https://github.com/AztecProtocol/barretenberg/pull/31#pullrequestreview-1240637105: "The SRS object contains a single `pippenger_runtime_state` object that is used to perform all pippenger computations (constructing `pippenger_runtime_state` is expensive)." 

The objective of this issue is to reduce the cost of using pippenger by using only a single `pippenger_runtime_state`.

arijitdutta67:
# Description

This PR enables the IPA scheme to use kzg commitment key. This helps the scheme to use kzg srs elements as the set of generators. It also enables the scheme to use a single `pippenger_runtime_state` for performing all multi-scalar-multiplications. In particular, it resolves the issues #163 , #164.

# Checklist:

### DIFF
--- a/cpp/src/barretenberg/honk/pcs/commitment_key.hpp
+++ b/cpp/src/barretenberg/honk/pcs/commitment_key.hpp
@@ -193,4 +193,95 @@ template <typename G> struct Params {
     using VK = VerificationKey<G>;
 };
 } // namespace fake
+
+namespace ipa {
+
+/**
+ * @brief CommitmentKey object over a group 𝔾₁, using a structured reference string (SRS).
+ * The SRS is given as a list of 𝔾₁ points
+ *  { [xʲ]₁ }ⱼ where 'x' is unknown.
+ *
+ * @todo This class should take ownership of the SRS, and handle reading the file from disk.
+ */
+class CommitmentKey {
+    using Fr = typename barretenberg::g1::Fr;
+    // C is a "raw commitment" resulting to be fed to the transcript.
+    using C = typename barretenberg::g1::affine_element;
+    // Commitment represent's a homomorphically computed group element.
+    using Commitment = barretenberg::g1::element;
+
+    using Polynomial = barretenberg::Polynomial<Fr>;
+
+  public:
+    CommitmentKey() = delete;
+
+    /**
+     * @brief Construct a new Kate Commitment Key object from existing SRS
+     *
+     * @param n
+     * @param path
+     *
+     * @todo change path to string_view
+     */
+    CommitmentKey(const size_t num_points, std::string_view path)
+        : pippenger_runtime_state(num_points)
+        , srs(num_points, std::string(path))
+    {}
+
+    /**
+     * @brief Uses the ProverSRS to create a commitment to p(X)
+     *
+     * @param polynomial a univariate polynomial p(X) = ∑ᵢ aᵢ⋅Xⁱ ()
+     * @return Commitment computed as C = [p(x)] = ∑ᵢ aᵢ⋅[xⁱ]₁
+     */
+    C commit(std::span<const Fr> polynomial)
+    {
+        const size_t degree = polynomial.size();
+        ASSERT(degree <= srs.get_monomial_size());
+        return barretenberg::scalar_multiplication::pippenger_without_endomorphism_basis_points(
+            const_cast<Fr*>(polynomial.data()), srs.get_monomial_points(), degree, pippenger_runtime_state);
+    };
+
+    barretenberg::scalar_multiplication::pippenger_runtime_state pippenger_runtime_state;
+    bonk::FileReferenceString srs;
+};
+
+class VerificationKey {
+    using Fr = typename barretenberg::g1::Fr;
+    using C = typename barretenberg::g1::affine_element;
+
+    using Commitment = barretenberg::g1::element;
+    using Polynomial = barretenberg::Polynomial<Fr>;
+
+  public:
+    VerificationKey() = delete;
+
+    /**
+     * @brief Construct a new Kate Commitment Key object from existing SRS
+     *
+     *
+     * @param verifier_srs verifier G2 point
+     */
+    VerificationKey(const size_t num_points, std::string_view path)
+        : pippenger_runtime_state(num_points)
+        , srs(num_points, std::string(path))
+    {}
+
+    barretenberg::scalar_multiplication::pippenger_runtime_state pippenger_runtime_state;
+    bonk::FileReferenceString srs;
+};
+
+struct Params {
+    using Fr = typename barretenberg::g1::Fr;
+    using C = typename barretenberg::g1::affine_element;
+
+    using Commitment = barretenberg::g1::element;
+    using Polynomial = barretenberg::Polynomial<Fr>;
+
+    using CK = CommitmentKey;
+    using VK = VerificationKey;
+};
+
+} // namespace ipa
+
 } // namespace honk::pcs
--- a/cpp/src/barretenberg/honk/pcs/commitment_key.test.hpp
+++ b/cpp/src/barretenberg/honk/pcs/commitment_key.test.hpp
@@ -31,6 +31,12 @@ template <> inline std::shared_ptr<kzg::CommitmentKey> CreateCommitmentKey<kzg::
     const size_t n = 128;
     return std::make_shared<kzg::CommitmentKey>(n, kzg_srs_path);
 }
+// For IPA
+template <> inline std::shared_ptr<ipa::CommitmentKey> CreateCommitmentKey<ipa::CommitmentKey>()
+{
+    const size_t n = 128;
+    return std::make_shared<ipa::CommitmentKey>(n, kzg_srs_path);
+}
 
 template <typename CK> inline std::shared_ptr<CK> CreateCommitmentKey()
 // requires std::default_initializable<CK>
@@ -44,13 +50,17 @@ template <> inline std::shared_ptr<kzg::VerificationKey> CreateVerificationKey<k
 {
     return std::make_shared<kzg::VerificationKey>(kzg_srs_path);
 }
-
+// For IPA
+template <> inline std::shared_ptr<ipa::VerificationKey> CreateVerificationKey<ipa::VerificationKey>()
+{
+    const size_t n = 128;
+    return std::make_shared<ipa::VerificationKey>(n, kzg_srs_path);
+}
 template <typename VK> inline std::shared_ptr<VK> CreateVerificationKey()
 // requires std::default_initializable<VK>
 {
     return std::make_shared<VK>();
 }
-
 template <typename Params> class CommitmentTest : public ::testing::Test {
     using CK = typename Params::CK;
     using VK = typename Params::VK;
@@ -187,6 +197,7 @@ template <typename Params>
 typename std::shared_ptr<typename Params::VK> CommitmentTest<Params>::verification_key = nullptr;
 
 using CommitmentSchemeParams = ::testing::Types<kzg::Params>;
+using IpaCommitmentSchemeParams = ::testing::Types<ipa::Params>;
 // IMPROVEMENT: reinstate typed-tests for multiple field types, i.e.:
 // using CommitmentSchemeParams =
 //     ::testing::Types<fake::Params<barretenberg::g1>, fake::Params<grumpkin::g1>, kzg::Params>;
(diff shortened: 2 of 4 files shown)


cpp/src/aztec/honk/pcs/commitment_key.hpp
```diff
@@ -80,7 +80,6 @@ class CommitmentKey {
             const_cast<Fr*>(polynomial.data()), srs.get_lagrange_points(), degree, pippenger_runtime_state);
     };
 
-  private:
```
codygunton: We should find a way to leave this as private. It seems to me this is a bit of a hack so that you can use the commitment key with kzg params for your IPA work. But it seems more natural (i.e., it respects the existing idiom) if you make IPA params for the commitment key. If you're worried about code duplication, we should find a solution that allows sharing code.

cpp/src/aztec/honk/pcs/ipa/ipa.hpp
```diff
@@ -68,9 +69,7 @@ template <typename Fr, typename Fq, typename G1> class InnerProductArgument {
      * @return an IpaProof, containing information required to verify whether the commitment is computed correctly and
      * the polynomial evaluation is correct in the given challenge point.
      */
-    static IpaProof ipa_prove(const IpaPubInput& ipa_pub_input,
-                              const std::vector<Fr>& polynomial,
-                              const std::vector<affine_element>& G_vector)
+    static IpaProof ipa_prove(std::shared_ptr<CK> ck, const IpaPubInput& ipa_pub_input, const Polynomial& polynomial)
```
codygunton: Even though I don't like the KZG interface very much, I think you should rename this function to `reduce_prove` for consistency with that.

cpp/src/aztec/honk/pcs/ipa/ipa.test.cpp
```diff
@@ -4,50 +4,51 @@
...
-    std::vector<barretenberg::g1::affine_element> G_vec(n);
-    for (size_t i = 0; i < n; ++i) {
-        auto scalar = fr::random_element();
-        G_vec[i] = barretenberg::g1::affine_element(barretenberg::g1::one * scalar);
-    }
-    InnerProductArgument<barretenberg::fr, barretenberg::fq, barretenberg::g1> newIpa;
-    auto C = newIpa.commit(coeffs, n, G_vec);
-    InnerProductArgument<barretenberg::fr, barretenberg::fq, barretenberg::g1>::IpaPubInput pub_input;
-    pub_input.commitment = C;
+    using newIpa = InnerProductArgument<TypeParam>;
```
codygunton: We don't really use this `new*` convention anywhere except, I see, in commitment_scheme.test.cpp. Will you just call it `IPA` or something like this? Similarly, we don't use camelCase, but CamelCase or snake_case.

cpp/src/aztec/honk/pcs/commitment_key.hpp
codygunton: This will require instantiating your IPA tests not with `CommitmentSchemeParams` (probably this should go away and we should instantiate the typed tests in each file with appropriate parameters) but with whatever new parameter struct you create.

codygunton (changes requested):

cpp/src/aztec/honk/pcs/ipa/ipa.hpp
codygunton: It's kind of ugly to have `IPA::ipa_prove`, right?

codygunton:
Regarding your points: 
>     I had to make srs and pippenger_runtime_state public inside commitment_key.hpp to use them inside IPA. Would like to have a discussion if that is concerning.
I left a suggestion for this in my review.
>     The srs is not generic. For now it seems ok to me as we are not going to use the grumpkin srs immediately. But I wish to know if some improvement on this is possible with minimal changes.
The aforementioned suggestion should also handle this. The commitment key is a flexible parameter that can contain/point to whatever sort of SRS you like.
>     I have kept the commit function inside IPA unlike other pcs as it uses pippenger_without_endomorphism_basis_points method instead of pippengen_unsafe method.
I would prefer that we keep the same structure as in honk/pcs/kzg, unless there's a good reason for departing from what's there. I think your point is that we are somehow not obligated to insulate the user from a call to an unsafe Pippenger method? This to me doesn't seem like a good reason to use different interfaces.

cpp/src/aztec/honk/pcs/commitment_key.hpp
arijitdutta67: Thanks for the suggestion. I have reverted the change inside `kzg` namespace. Instead, I have added separate `commitment_key` and `verification_key` classes inside `ipa` namespace. Here I have kept `srs` and `pippenger_runtime_state` as public member variables. This seems reasonable to me because we are actively using these member variables inside `reduce_prove` and `reduce_verify` methods in `ipa` unlike other commitment schemes. Please let me know if you have any thoughts on this.

cpp/src/aztec/honk/pcs/ipa/ipa.hpp
arijitdutta67: Thanks. Done.

cpp/src/aztec/honk/pcs/ipa/ipa.test.cpp
arijitdutta67: Thanks. Done.

arijitdutta67:
> Regarding your points:
> 
> > ```
> > I had to make srs and pippenger_runtime_state public inside commitment_key.hpp to use them inside IPA. Would like to have a discussion if that is concerning.
> > ```
> 
> I left a suggestion for this in my review.
> 
> > ```
> > The srs is not generic. For now it seems ok to me as we are not going to use the grumpkin srs immediately. But I wish to know if some improvement on this is possible with minimal changes.
> > ```
> 
> The aforementioned suggestion should also handle this. The commitment key is a flexible parameter that can contain/point to whatever sort of SRS you like.
> 
> > ```
> > I have kept the commit function inside IPA unlike other pcs as it uses pippenger_without_endomorphism_basis_points method instead of pippengen_unsafe method.
> > ```
> 
> I would prefer that we keep the same structure as in honk/pcs/kzg, unless there's a good reason for departing from what's there. I think your point is that we are somehow not obligated to insulate the user from a call to an unsafe Pippenger method? This to me doesn't seem like a good reason to use different interfaces.

Thanks Cody for you timely suggestions which helped me a lot. I have tried my best to adhere to your suggestions. Please let me know if you have any more concerns.

cpp/src/aztec/honk/pcs/commitment_key.hpp
```diff
@@ -194,4 +194,95 @@ template <typename G> struct Params {
...
+        , srs(num_points, std::string(path))
+    {}
+
+    /**
+     * @brief Uses the ProverSRS to create a commitment to p(X)
+     *
+     * @param polynomial a univariate polynomial p(X) = ∑ᵢ aᵢ⋅Xⁱ ()
+     * @return Commitment computed as C = [p(x)] = ∑ᵢ aᵢ⋅[xⁱ]₁
+     */
+    C commit(std::span<const Fr> polynomial)
```
codygunton: This function is not used by the IPA commitment scheme.

codygunton: This diverges from what is done in KZG. We need to make a decision about which approach to take and uniformize the code, but let's just get your work in and leave that for a future PR.

codygunton: This also applies to the vk; in KZG, the vk has a function to do verification work.

cpp/src/aztec/honk/pcs/ipa/ipa.test.cpp
```diff
@@ -4,50 +4,48 @@
...
-        points[i] = barretenberg::g1::affine_element(barretenberg::g1::element::random_element());
-    }
+TYPED_TEST_SUITE(IpaCommitmentTest, IpaCommitmentSchemeParams);
 
-    barretenberg::g1::element expected = points[0] * scalars[0];
+TYPED_TEST(IpaCommitmentTest, commit)
+{
+    constexpr size_t n = 128;
+    auto poly = this->random_polynomial(n);
+    auto CK = this->ck();
```
codygunton: `CK` is used for the type of the commitment key, so it would be better to call this `ck` or `key`.

cpp/src/aztec/honk/pcs/ipa/ipa.test.cpp
```diff
@@ -4,50 +4,48 @@
...
-    }
-    InnerProductArgument<barretenberg::fr, barretenberg::fq, barretenberg::g1> newIpa;
-    auto C = newIpa.commit(coeffs, n, G_vec);
-    InnerProductArgument<barretenberg::fr, barretenberg::fq, barretenberg::g1>::IpaPubInput pub_input;
-    pub_input.commitment = C;
+    using IPA = InnerProductArgument<TypeParam>;
+    using PubInput = typename IPA::PubInput;
+    // generate a random polynomial, degree needs to be a power of two
+    size_t n = 128;
+    // auto [claim, witness] = this->random_claim(n);
```
codygunton: Extraneous comment line can go away.

codygunton (dismissed):
You raised the following issues in our review call:
 - You were unsure whether it is secure to use unsafe Pippenger methods here.
 - You were concerned that it may be inefficient to use the same Pippenger runtime state for different sets of generators and different MSM lengths.

Will you briefly look into these questions, if you haven't, and make github issues as appropriate?

cpp/src/aztec/honk/pcs/commitment_key.hpp
codygunton: ...but, as we said, this is an issue to be handled later by a PR that makes the PCS a swappable component of Honk. Would you please make an issue recording this?

cpp/src/aztec/honk/pcs/ipa/ipa.test.cpp
arijitdutta67: Thanks for pointing this out. Removed this and directly using `srs` now e.g. `auto srs_elements = this->ck()->srs.get_monomial_points();`

cpp/src/aztec/honk/pcs/commitment_key.hpp
arijitdutta67: Thanks. Created the issue #236 for this.

arijitdutta67:
> You raised the following issues in our review call:
> 
> * You were unsure whether it is secure to use unsafe Pippenger methods here.
> * You were concerned that it may be inefficient to use the same Pippenger runtime state for different sets of generators and different MSM lengths.
> 
> Will you briefly look into these questions, if you haven't, and make github issues as appropriate?

Thanks. Created the issue #237 for this.

cpp/src/barretenberg/honk/pcs/commitment_key.hpp
```diff
@@ -193,4 +193,95 @@ template <typename G> struct Params {
...
+    // Commitment represent's a homomorphically computed group element.
+    using Commitment = barretenberg::g1::element;
+
+    using Polynomial = barretenberg::Polynomial<Fr>;
+
+  public:
+    CommitmentKey() = delete;
+
+    /**
+     * @brief Construct a new Kate Commitment Key object from existing SRS
```
codygunton, line 219: IPA, you mean.

cpp/src/barretenberg/honk/pcs/commitment_key.hpp
```diff
@@ -193,4 +193,95 @@ template <typename G> struct Params {
     using VK = VerificationKey<G>;
 };
 } // namespace fake
+
+namespace ipa {
+
+/**
+ * @brief CommitmentKey object over a group 𝔾₁, using a structured reference string (SRS).
+ * The SRS is given as a list of 𝔾₁ points
```
codygunton, line 201: Slightly weird line break here.

cpp/src/barretenberg/honk/pcs/commitment_key.hpp
arijitdutta67: Nice catch! Resolved in PR #244.

cpp/src/barretenberg/honk/pcs/commitment_key.hpp
arijitdutta67: Thanks for noticing. Resolved in PR #244.

PR #235, opened 2023-03-09
Lde/deprecate polynomial cache

### LINKED ISSUE #215: Nix PolynomialCache
The PolynomialCache (coupled with a PolynomialMemoryStore) was originally added to facilitate a low memory prover. The primary motivation was construction of UP proofs in the browser for AC. Since this is no longer required, the PolynomialCache can be swapped out for a simpler class that essentially just wraps a std::map with minor additional functionality.

ledwards2225:
# Description

Replace the `PolynomialCache` with a much simpler container called `PolynomialStore`. The former had complicated functionality to facilitate memory reduction. This was really only relevant for Plonk variants (in particular, the computation of UP proofs in the browser) and was never fully switched on. The `PolynomialStore` is a simple wrapper around `std::map`.

Note: I still prefer a custom class to simply using `std::map` directly because 1) std::map has some undesirable behavior that I want to restrict, for example, `operator[]` when called on a key that does not exist will return an empty `value`, and 2) this makes it easy to provide some nice bonus methods like `print()` and `get_size_in_bytes()`.

# Checklist:


kevaundray:
@ledwards2225 out of scope for this PR, one thing you can do is extend the `put` and `get` methods with method like:

- `put_lagrange` 
- `put_extended_lagrange` / `put_fft` 
- `put_monomial`

Where `put_lagrange` for example calls `put` and appends "_lagrange" to the key.

The benefit being that the codebase no longer has "_fft" and "_lagrange" string literals all over the place; they would be localized to the `polynomial_store` object class.

The disadvantage is that API of the `polynomial_store` method becomes more than just `put` and `get` which is quite simple

ledwards2225:
@kevaundray It's a good point. I would hate to pollute Honk (where there is only one representation) with `put_lagrange` (and `get_lagrange`) but perhaps we could compromise by letting `put` imply `put_lagrange` and only using the more verbose `put_monomial` and `put_extended_lagrange` for those cases. This is a tad icky too. Your point about the strings is well taken. I don't think that pattern has any defenders around here.

kevaundray:
> @kevaundray It's a good point. I would hate to pollute Honk (where there is only one representation) with `put_lagrange` (and `get_lagrange`) but perhaps we could compromise by letting `put` imply `put_lagrange` and only using the more verbose `put_monomial` and `put_extended_lagrange` for those cases. This is a tad icky too. Your point about the strings is well taken. I don't think that pattern has any defenders around here.

Oh my bad, didn't know honk was only using Lagrange form. That's a good point against using this, especially if the standard/turbo/ultra code will eventually calcify and changes will be made mostly/only to honk

cpp/src/barretenberg/honk/composer/composer_helper/composer_helper_lib.cpp
```diff
@@ -97,11 +100,11 @@ void compute_monomial_and_coset_selector_forms(bonk::proving_key* circuit_provin
         barretenberg::polynomial selector_poly_fft(selector_poly, circuit_proving_key->circuit_size * 4 + 4);
         selector_poly_fft.coset_fft(circuit_proving_key->large_domain);
 
-        // TODO(#215)(Luke/Kesha): Lagrange polynomials could be deleted from cache here since they are no longer
-        // needed.
+        // Remove the selector lagrange forms since they will not be needed beyond this point
+        circuit_proving_key->polynomial_store.remove(selector_properties[i].name + "_lagrange");
```
Rumata888, line 104: Sweet. Thanks for this

cpp/src/barretenberg/proof_system/polynomial_store/polynomial_store.test.cpp
```diff
@@ -0,0 +1,114 @@
...
+    Polynomial poly(1024);
+    for (auto& coeff : poly) {
+        coeff = Fr::random_element();
+    }
+
+    // Make a copy for comparison after original is moved into container
+    Polynomial poly_copy(poly);
+
+    // Move the poly into the container
+    polynomial_store.put("id", std::move(poly));
```
Rumata888, line 28: Is there a need for std::move here? You are doing it inside anyway

cpp/src/barretenberg/proof_system/work_queue/work_queue.cpp
```diff
@@ -111,12 +111,15 @@ void work_queue::put_fft_data(barretenberg::fr* result, const size_t work_item_n
...
                 const size_t n = key->circuit_size;
-                barretenberg::polynomial& wire_fft = key->polynomial_cache.get(item.tag + "_fft", 4 * n + 4);
+                barretenberg::polynomial wire_fft(4 * n + 4);
 
                 for (size_t i = 0; i < n; ++i) {
                     wire_fft[4 * i + item.index] = result[i];
                 }
                 wire_fft[4 * n + item.index] = result[0];
+
+                key->polynomial_store.put(item.tag + "_fft", std::move(wire_fft));
```
Rumata888, line 121: Do we no longer have this polynomial in the store before the computation? Because if we do, then we'll be using extra memory

Rumata888 (approved):
Everything looks great. Please, just look at the small fft in work queue. There is probably a reason why you did it this way, but if there is still a polynomial in store before the computation, then we can save some resources

cpp/src/barretenberg/proof_system/work_queue/work_queue.cpp
ledwards2225: In this case, the polynomial does not exist in the cache when we call `get`. There used to be an overload of `get` of the form `get(key, size)` that would construct a polynomial of the given `size` in place if it did not exist in the cache. So previously, we were constructing a size `4n+4` polynomial of zeros in the cache then setting its values in place. (This was of course not clear from the code which is why I removed that overload). Now I'm constructing that same polynomial locally, then moving it. In either case, we're only storing the memory once at any given time.

cpp/src/barretenberg/proof_system/polynomial_store/polynomial_store.test.cpp
ledwards2225: Since this is just a test I suppose there is no real need, but since most of these tests are meant to be a simple illustration of how the store is to be used, I think it makes sense to demonstrate the correct pattern.

PR #247, opened 2023-03-15
Native workalike to `compute_tree_root` in `plonk::stlib::merkle_tree`

dbanks12:
# Description

Native workalike to `compute_tree_root` in `plonk::stlib::merkle_tree`

# Checklist:

### DIFF
--- a/cpp/src/barretenberg/stdlib/merkle_tree/hash.hpp
+++ b/cpp/src/barretenberg/stdlib/merkle_tree/hash.hpp
@@ -32,12 +32,11 @@ inline barretenberg::fr compute_tree_root_native(std::vector<barretenberg::fr> c
 {
     // Check if the input vector size is a power of 2.
     ASSERT(input.size() > 0);
-    ASSERT(!(input.size() & (input.size() - 1)) == true);
+    ASSERT(numeric::is_power_of_two(input.size()));
     auto layer = input;
     while (layer.size() > 1) {
         std::vector<barretenberg::fr> next_layer(layer.size() / 2);
         for (size_t i = 0; i < next_layer.size(); ++i) {
-            next_layer[i] = crypto::pedersen_hash::hash_multiple({ layer[i * 2], layer[i * 2 + 1] });
             if (plonk::SYSTEM_COMPOSER == plonk::PLOOKUP) {
                 next_layer[i] = crypto::pedersen_hash::lookup::hash_multiple({ layer[i * 2], layer[i * 2 + 1] });
             } else {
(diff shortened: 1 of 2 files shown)


suyash67:
We need to call another function if we're using UltraPlonk. So you'll have to add conditional logic like:
```c++
if (plonk::SYSTEM_COMPOSER == plonk::PLOOKUP) {
    return crypto::pedersen_hash::lookup::hash_multiple({ lhs, rhs });
} else {
    return crypto::pedersen_hash::hash_multiple({ lhs, rhs });
}
```

dbanks12:
> We need to call another function if we're using UltraPlonk. So you'll have to add conditional logic like:

Thanks for the feedback @suyash67! Why do we need to use different pedersen implementations here when these are native functions? What makes a **native** pedersen hash different when done in via ultraplonk?

dbanks12:
Fixed here: https://github.com/AztecProtocol/barretenberg/pull/247/commits/c379a74ad0d80ea36b3738ecb1a76b5d78b61bdd

suyash67:
> > We need to call another function if we're using UltraPlonk. So you'll have to add conditional logic like:
> 
> Thanks for the feedback @suyash67! Why do we need to use different pedersen implementations here when these are native functions? What makes a **native** pedersen hash different when done in via ultraplonk?

Sorry for a late reply, I missed this comment. Basically, ultraplonk uses a very efficient lookup based Pedersen hash for merkle tree hashing. On the other hand, Pedersen hash using turbo plonk uses a custom fixed base gate. The output of the Pedersen when using turbo and ultra are different (because the underlying designs are totally different). That's the reason we want the conditional logic.

cpp/src/barretenberg/stdlib/merkle_tree/hash.hpp
```diff
@@ -22,6 +22,34 @@ inline barretenberg::fr compress_native(barretenberg::fr const& lhs, barretenber
...
+inline barretenberg::fr compute_tree_root_native(std::vector<barretenberg::fr> const& input)
+{
+    // Check if the input vector size is a power of 2.
+    ASSERT(input.size() > 0);
+    ASSERT(!(input.size() & (input.size() - 1)) == true);
+    auto layer = input;
+    while (layer.size() > 1) {
+        std::vector<barretenberg::fr> next_layer(layer.size() / 2);
+        for (size_t i = 0; i < next_layer.size(); ++i) {
+            next_layer[i] = crypto::pedersen_hash::hash_multiple({ layer[i * 2], layer[i * 2 + 1] });
```
suyash67: Do we need this line?

PR #248, opened 2023-03-16
Lde/transcript

### LINKED ISSUE #161: New manifest paradigm for Honk
The manifests are currently a primary source of loose coupling in Barretenberg. Resolving this issue would give a more flexible transcript (e.g., simplifying the tests that now require a manifest (sometimes a partial one)) and would follow the paradigm of having "a single source of information". Namely, the prover's algorithm and the verifier's algorithm will be specified in code, and a manifest recording their interactions will be documented in a test. In-order execution of the algorithms will be enforced via the properties of the hash function that is used to generate the challenges; out-of-order execution will lead to differing challenges.

### LINKED ISSUE #226: TODO: Remove zeta and alpha from RelationParameters
The class RelationParameters contains zeta and alpha which are not used by the relations. They are however used in Sumcheck so this is not as simple as deleting them. It's possible that the right thing to do is simply rename/reorganize the class.

ledwards2225:
# Description

Integrate Adrian's new transcript class.

High level changes:
- There is now a `ProverTranscript` and a `VerifierTranscript`, which implement specific methods on top of basic functionality inherited from `BaseTranscript`
- The prover and verifier both dynamically construct a `TranscriptManifest` object as they interact with their respective transcripts. The transcript classes themselves do not constrain the user to any predefined manifest, however we now have tests that check consistency between the expected manifest and the one produced in the course of proof construction/verification.
- Sumcheck now returns the data needed subsequently in the prover (multivariate challenge and multivariate evaluations), rather than adding it to the transcript then re-extracting it. (Note: in the new model, a `ProverTranscript` has no method to retrieve data from the transcript once it has been written).

# Checklist:


cpp/src/barretenberg/honk/pcs/claim.hpp
```diff
@@ -81,19 +77,15 @@ template <typename Params> class OpeningClaim {
  *
  * @tparam CommitmentKey
  */
-template <typename Params> struct MLEOpeningClaim {
-    using Commitment = typename Params::Commitment;
+template <typename Params> class MLEOpeningClaim {
```
codygunton, line 80: `class` because it has a method and you're thinking `struct` sounds more like a container?

cpp/src/barretenberg/honk/pcs/gemini/gemini.hpp
```diff
@@ -165,19 +166,17 @@ template <typename Params> class MultilinearReductionScheme {
...
-                                    static_cast<CommitmentAffine>(commitments[l]).to_buffer());
+            transcript.send_to_verifier("Gemini:FOLD_" + std::to_string(l + 1), commitments[l]);
         }
 
         /*
          * Generate evaluation challenge r, and compute rₗ = r^{2ˡ} for l = 0, 1, ..., m-1
          */
-        transcript->apply_fiat_shamir("r");
-        const Fr r_challenge = Fr::serialize_from_buffer(transcript->get_challenge("r").begin());
+        const Fr r_challenge = transcript.get_challenge("Gemini:r");
```
codygunton, line 179: Nice.

cpp/src/barretenberg/honk/pcs/gemini/gemini.hpp
```diff
@@ -259,17 +258,33 @@ template <typename Params> class MultilinearReductionScheme {
...
+                                                           VerifierTranscript<Fr>& transcript)
     {
         const size_t num_variables = mle_opening_point.size();
 
+        // Get polynomials Fold_i, i = 1,...,m-1 from transcript
+        std::vector<CommitmentAffine> commitments;
+        commitments.reserve(num_variables - 1);
+        for (size_t i = 0; i < num_variables - 1; ++i) {
+            auto commitment =
+                transcript.template receive_from_prover<CommitmentAffine>("Gemini:FOLD_" + std::to_string(i + 1));
```
codygunton, line 270: Yuck. Ultimately, I think the verifier transcript is written this way in order to: avoid supplying the transcript with the type of a commitment; use a single deserialization method for all types. But we only ever receive field elements and commitments, so what do you think about 2x overloading `receive_from_prover` manually and supplying the commitment type?

cpp/src/barretenberg/honk/pcs/gemini/gemini.hpp
```diff
@@ -259,17 +258,33 @@ template <typename Params> class MultilinearReductionScheme {
...
+        // Get evaluations a_i, i = 0,...,m-1 from transcript
+        std::vector<Fr> evaluations;
+        evaluations.reserve(num_variables);
+        for (size_t i = 0; i < num_variables; ++i) {
+            auto eval = transcript.template receive_from_prover<Fr>("Gemini:a_" + std::to_string(i + 1));
+            evaluations.emplace_back(eval);
+        }
+
         // Compute evaluation A₀(r)
-        auto a_0_pos = compute_eval_pos(batched_evaluation, mle_opening_point, r_squares, proof.evaluations);
```
codygunton, line 272: The point here is that the transcript already contains the proof, right?

cpp/src/barretenberg/honk/transcript/transcript.hpp
```diff
@@ -0,0 +1,267 @@
...
+        transcript.send_to_verifier("Init", init);
+        return transcript;
+    };
+};
+
+template <class FF> class VerifierTranscript : public BaseTranscript<FF> {
+
+    /// Contains the raw data sent by the prover.
+    const std::vector<uint8_t> proof_data_;
+    size_t num_bytes_read_ = 0;
```
codygunton, line 231: The main cost of using a single class between prover and verifier is that we have to allocate this? Seems unlikely the cost of instantiating the functions matters. Is there a good reason to have three classes in this file instead of one?

cpp/src/barretenberg/honk/transcript/transcript.hpp
```diff
@@ -0,0 +1,267 @@
...
+        // TODO(Adrian): Do we want to use a domain separator as the initial challenge buffer?
+        // We could be cheeky and use the hash of the manifest as domain separator, which would prevent us from having
+        // to domain separate all the data. (See https://safe-hash.dev)
+        std::vector<uint8_t> full_buffer;
+        if (round_number > 0) {
+            full_buffer.insert(full_buffer.end(), previous_challenge_buffer.begin(), previous_challenge_buffer.end());
+        }
+        full_buffer.insert(full_buffer.end(), current_round_data.begin(), current_round_data.end());
+
+        // Optionally pre-hash the full buffer to minimize the amount of data passed to the cryptographic hash function.
```
codygunton: It's not really optional here.

cpp/src/barretenberg/honk/transcript/transcript.hpp
```diff
@@ -0,0 +1,267 @@
...
+     * @param element Serializable object that will be added to the transcript
+     *
+     * @todo Use a concept to only allow certain types to be passed. Requirements are that the object should be
+     * serializable.
+     *
+     */
+    template <class T> void send_to_verifier(const std::string& label, const T& element)
+    {
+        using serialize::write;
+        // DANGER: When serializing an affine_element, we write the x and y coordinates
```
codygunton: Can we elaborate here? Also "annowing" ~> "annoying".

codygunton: It this note stays in we should add a "todo" label somewhere because we might not remember to search for `DANGER`.

cpp/src/barretenberg/honk/transcript/transcript.test.cpp
```diff
@@ -0,0 +1,213 @@
...
+using namespace honk;
+
+template <typename FF> class TranscriptTest : public testing::Test {
+  public:
+    /**
+     * @brief Construct a manifest for a standard Honk proof
+     *
+     * @return TranscriptManifest
+     */
+    TranscriptManifest construct_standard_honk_manifest(size_t circuit_size)
```
codygunton, line 26: I wasn't totally sure whether this manifest functionality was worth implementing, but looking at it here, I think it's natural and very nice to read in this form.

cpp/src/barretenberg/honk/transcript/transcript.hpp
```diff
@@ -0,0 +1,267 @@
...
+ *
+ * @tparam FF Field from which we sample challenges.
+ */
+template <typename FF> class BaseTranscript {
+    // TODO(Adrian): Make these tweakable
+    static constexpr size_t HASH_OUTPUT_SIZE = 32;
+    static constexpr size_t MIN_BYTES_PER_CHALLENGE = 128 / 8; // 128 bit challenges
+
+    size_t round_number = 0;
+    std::array<uint8_t, HASH_OUTPUT_SIZE> previous_challenge_buffer{};
```
codygunton: Can we explicitly `std::fill` this with `0`s?

codygunton: The  fact that we use the `{}` means that in the present case it is filled with zeros, but I prefer to be explicit rather than to rely on the behavior of the underlying type (I don't know what guarantees exist).

cpp/src/barretenberg/honk/sumcheck/sumcheck.test.cpp
```diff
@@ -74,68 +75,11 @@ std::array<std::span<FF>, NUM_POLYNOMIALS> construct_full_polynomials(std::array
     return full_polynomials;
 }
 
-Transcript produce_mocked_transcript(size_t multivariate_d, size_t num_public_inputs)
```
codygunton, line 77: Yay.

cpp/src/barretenberg/honk/sumcheck/sumcheck.test.cpp
```diff
@@ -196,19 +140,19 @@ TEST(Sumcheck, PolynomialNormalization)
...
 
     auto sumcheck = Sumcheck<FF,
-                             Transcript,
+                             ProverTranscript<FF>,
                              ArithmeticRelation,
                              GrandProductComputationRelation,
                              GrandProductInitializationRelation>(multivariate_n, transcript);
 
-    sumcheck.execute_prover(full_polynomials);
+    auto [multivariate_query, evaluations] = sumcheck.execute_prover(full_polynomials, {});
```
codygunton: "query"?

cpp/src/barretenberg/honk/sumcheck/sumcheck_output.hpp
```diff
@@ -0,0 +1,19 @@
...
+#include <vector>
+namespace honk::sumcheck {
+
+/**
+ * @brief Contains the multi-linear `evaluations` of the polynomials at the `evaluation_point`.
+ * These are computed by the prover and need to be checked using a multi-linear PCS like Gemini.
+ */
+template <typename FF> struct SumcheckOutput {
+    // u = (u_0, ..., u_{d-1})
+    std::vector<FF> multivariate_query;
```
codygunton: Yeah can we call this something else?

codygunton: It's a challenge point.

cpp/src/barretenberg/honk/sumcheck/sumcheck.hpp
```diff
@@ -67,70 +74,35 @@ template <typename FF, class Transcript, template <class> class... Relations> cl
...
-            polynomial.resize(multivariate_n >> 1);
-        }
-    };
-
-    /**
-     * @brief Get all the challenges and computed parameters used in sumcheck in a convenient format
-     *
-     * @return RelationParameters<FF>
-     */
-    RelationParameters<FF> retrieve_proof_parameters()
```
codygunton, line 90: Yay.

cpp/src/barretenberg/honk/proof_system/prover.cpp
```diff
@@ -66,6 +64,13 @@ Prover<settings>::Prover(std::vector<barretenberg::polynomial>&& wire_polys,
     prover_polynomials[POLYNOMIAL::W_L] = wire_polynomials[0];
     prover_polynomials[POLYNOMIAL::W_R] = wire_polynomials[1];
     prover_polynomials[POLYNOMIAL::W_O] = wire_polynomials[2];
+
+    // Add public inputs to transcript from the second wire polynomial
```
codygunton, line 68: Err, why is this public inputs stuff happening here now?

cpp/src/barretenberg/honk/proof_system/prover.cpp
```diff
@@ -191,57 +196,39 @@ template <typename settings> Polynomial Prover<settings>::compute_grand_product_
...
-    // Add some initial data to transcript (circuit size and PI size)
-
     // queue.flush_queue(); // NOTE: Don't remove; we may reinstate the queue
 
-    transcript.add_element("circuit_size",
-                           { static_cast<uint8_t>(key->circuit_size >> 24),
-                             static_cast<uint8_t>(key->circuit_size >> 16),
-                             static_cast<uint8_t>(key->circuit_size >> 8),
-                             static_cast<uint8_t>(key->circuit_size) });
+    const auto circuit_size = static_cast<uint32_t>(key->circuit_size);
```
codygunton, line 206: This new interface is obviously better.

codygunton (commented):
This transcript upgrade is a big improvement. I have a couple of thoughts on structure and some more minor quibbles.

cpp/src/barretenberg/honk/pcs/claim.hpp
ledwards2225: As far as I'm concerned there's no practical reason to use `struct` ever. I think it's sometimes used as a stylistic shorthand to say "this is a simple class that only contains data members, not methods". That's fine. But IMO it's much better to use `class` and be explicit with access specifiers.

cpp/src/barretenberg/honk/pcs/gemini/gemini.hpp
ledwards2225: This change is reflecting an overall change in how the prover is allowed to use the transcript. In the old model, the prover would use the transcript as a sort of container for the data it was producing. So in this case, Sumcheck was computing evaluations and adding them to the transcript. Then, in the prover, we would extract those evals from the transcript and place them into a `proof` object which could be passed along to Gemini. With this new transcript paradigm, the prover never extracts anything from the transcript once it has been added. If data is computed in one module then needed in another, the first module has to return it explicitly, which is why Sumcheck now returns data.

cpp/src/barretenberg/honk/pcs/gemini/gemini.hpp
ledwards2225: So we would have a function like call like `transcript.receive_from_prover(label, data_type)` where `data_type` is some kind of flag to indicate the data type? I count at least 5 different types that we need to retrieve: `uint_32`, `FF`, `Commitment`, `Univariate`, `array<FF,...>`. I agree the `.template ` thing is icky. Maybe what you suggest is preferable but I do like the flexibility of the existing method. Let's discuss

cpp/src/barretenberg/honk/proof_system/prover.cpp
ledwards2225: This was something Adrian did that I think makes a lot of sense. First of all, the public inputs are needed in two places (add to transcript in some early round, then use to compute PI delta in grand product computation) so it makes sense to have a member variable `public_inputs`. Also, they're stored in the wires and we initialize the prover owned wires in this `Prover` constructor so I think it's natural to do the same for the PI. The final change (which was not necessary but makes good sense to me) was to add the PI to the transcript in the same round that we add `public_input_size`, rather than (seemingly arbitrarily) in `execute_wire_commitments_round`. I'm assuming there was originally some reason for this which has since gone away. Or maybe I'm missing a reason that still exists? Either way let's discuss

cpp/src/barretenberg/honk/sumcheck/sumcheck.test.cpp
ledwards2225: Yeah I've started to use `query` and `evaluation` for respectively, the point at which a polynomial is evaluated and the result of that evaluation. I'm open to other terminology I mostly just want to develop a clear convention.

cpp/src/barretenberg/honk/sumcheck/sumcheck_output.hpp
ledwards2225: Ok, so if we have v = f(z) (for either univariate or multilinear) we would have z = `challenge` and v = `evaluation`?

cpp/src/barretenberg/honk/transcript/transcript.hpp
ledwards2225: Sure, we can do that

cpp/src/barretenberg/honk/transcript/transcript.hpp
ledwards2225: This was Adrian's comment. I'll get some more details from him then update to TODO / clarify / remove as appropriate

cpp/src/barretenberg/honk/transcript/transcript.hpp
ledwards2225: I think the splitting makes sense given the sort of "interactive protocol" language that Adrian has adopted. The prover and verifier only share the common functionality in `BaseTranscript`. Aside from that, the prover can only "send to verifier" and the verifier can only "receive from prover". I don't think combining into one class makes the usage more clear and it opens up the developer to misuse by say, calling "receive from prover" in the prover..

cpp/src/barretenberg/honk/pcs/gemini/gemini.hpp
codygunton: Oh, I was just thinking overloading name with different types, no flag needed, but  was only thinking of field elements and commitments. But yeah let's chat.

cpp/src/barretenberg/honk/sumcheck/sumcheck_output.hpp
codygunton: I mean, if you like, I think it makes more sense to call it a query in the context of a PCS, since the verifier makes a query (though I'm used to evaluating functions at "points" and not "queries"...). But here it read as a bit jarring to me since the context is just sumcheck.

cpp/src/barretenberg/honk/transcript/transcript.hpp
codygunton: Yeah, but in the interactive protocol there is a single thing called at transcript that must be the same for the prover and the verifier. Like splitting the transcript _breaks_ the semantics if anything IMO.

cpp/src/barretenberg/honk/proof_system/prover.cpp
codygunton: Yeah, sounds reasonable, I just wasn't clear on why it's happening in this PR, if it was necessary I wanted to understand why.

cpp/src/barretenberg/honk/pcs/gemini/gemini.hpp
ledwards2225: We've decided that there is no (clear) superior alternative to the templated method used here.

cpp/src/barretenberg/honk/transcript/transcript.hpp
ledwards2225: We've agreed this is OK for now.

cpp/src/barretenberg/honk/sumcheck/sumcheck_output.hpp
ledwards2225: Changed to `challenge_point`

cpp/src/barretenberg/honk/transcript/transcript.hpp
ledwards2225: It seems like we could alternatively just pass the full buffer to blake3 but we use the (cheaper?) pedersen to compress the potentially large buffer to the size of the pedersen output before doing blake3. I think this is what he meant by "optional". What do you think?

cpp/src/barretenberg/honk/transcript/transcript.hpp
ledwards2225: On second thought, to do this I need to add a constructor where there was not one before:
`  public:
    BaseTranscript ()
    {
        previous_challenge_buffer.fill(0);
    }`

As you pointed out, zero initialization is guaranteed in this case so filling with zeros would really only serve to clarify. I'm not sure this even achieves that purpose but I'll default to your preference.

ledwards2225: Update: after doing some digging, I'm having a hard time convincing myself beyond any doubt that zero initialization is indeed guaranteed in this case. I asked @adr1anh if he has confidence on this and if he could point us to a reference. We can of course just go with the extra safe option but now I'm just curious.

ledwards2225: Ok, I'm convinced that the line in question is guaranteed to default initialize the elements of the array (in this case  `uint_8` but more generally `T`) and since the the default value for `uint_8` is `0U`, the array is guaranteed to be "filled" with zeros. We can still add the `fill` if you'd like but I think it is unnecessary and could lead to confusion.

cpp/src/barretenberg/honk/transcript/transcript.hpp
ledwards2225: TODOs added (with attribution to Adrian) with more details

cpp/src/barretenberg/honk/transcript/transcript.hpp
ledwards2225: Adrian referenced this [post](https://stackoverflow.com/questions/18295302/default-initialization-of-stdarray). [This](https://en.cppreference.com/w/cpp/language/zero_initialization) is also helpful.

cpp/src/barretenberg/honk/sumcheck/sumcheck.hpp
```diff
@@ -12,17 +14,22 @@
 #include <string>
 #include <vector>
 #include "barretenberg/honk/proof_system/prover.hpp"
+#include "barretenberg/honk/sumcheck/sumcheck_output.hpp"
+#include <optional>
+
 namespace honk::sumcheck {
+
 template <typename FF, class Transcript, template <class> class... Relations> class Sumcheck {
```
adr1anh, line 22: I think we can remove the `Transcript` template parameter now.

cpp/src/barretenberg/honk/transcript/transcript.hpp
adr1anh: Compression is useful when the verifier can efficiently compute group operations. This is not the case with a smart contract verifier where it's cheaper to pass the whole buffer to Keccak. 
It may be optional later on, either with cycle-of-curve recursion, or due to the new hash function we end up using.

adr1anh (approved):
Looks great!

cpp/src/barretenberg/honk/sumcheck/sumcheck.hpp
ledwards2225: Definitely. I got carried away with the "minimal changes" mantra..

cpp/src/barretenberg/honk/transcript/transcript.hpp
codygunton: Thanks for explaining. I updated the comment to this effect.

cpp/src/barretenberg/honk/sumcheck/sumcheck.hpp
codygunton: `Transcript` currently specializes to `ProverTranscript` and `VerifierTranscript`. If someone wants to make argument for splitting the whole library in to prover and verifier versions of everything, I'll listen, but I'm not in favor without a compelling argument. I'm really not even in favor of this for the `Transcript`, where I think the mental model of a single transcript between two parties reflects better the mental model of zk protocols, has essentially no implementation cost (+ we get to work with one class instead of three), and agrees with what is done in `plonk`.

PR #310, opened 2023-04-04
test: more vk tests to compare circuit/native/vk_data

dbanks12:
# Description

Add a test for recursion VK that makes sure its compressions match for circuit-types, native-types and vanilla vk_data. Some other minor fixes in other test.

# Checklist:


suyash67 (approved):
lgtm (except the circle ci comment below)

.circleci/config.yml
```diff
@@ -209,6 +209,18 @@ jobs:
           command: cond_spot_run_tests barretenberg-x86_64-linux-clang-assert 3 stdlib_recursion_tests --gtest_filter=-*turbo*
       - *save_logs
 
+  stdlib-verification-key-tests:
```
suyash67: I suppose you've added a separate circleci trigger so that `ultra-recursion`doesn't exceed 20 minute limit? But this thing now would run twice: once while running `{ stdlib_recursion_tests } - { *turbo* }` and other would be `stdlib-verification-key-tests`.

dbanks12: Correct itll run in its own job so it shouldn't exceed the limit. Ah you're right about it running twice. Should I filter the ultra tests to omit the vk tests too? Could we filter on the word "ultra" instead?

PR #311, opened 2023-04-04
Build a3crypto.wasm

spalladino:
Builds a smaller wasm than barretenberg with crypto primitives to be used from aztec3-packages. Goal is to have access to pedersen (and others) from small packages like p2p, that don't require the full-blown barretenberg or circuits wasm.

See [this thread](https://aztecprotocol.slack.com/archives/C03P17YHVK8/p1680533771535709) for more info.


ludamad:
Does this need asyncify? (=do any of these rely on the CRS?) 
We can try building with -Os (not necessary for this PR)
LGTM

ludamad (approved):

ludamad:
Not sure about the name however, maybe barretenberg_core_primitives or similar could be a thought

spalladino:
> Does this need asyncify? (=do any of these rely on the CRS?)

That's required for calling back to ts land, right? I believe not for Pedersen at least. Still, I'm not sure how to check exhaustively.

ludamad (approved):

PR #321, opened 2023-04-05
Split shplonk in prep for work queue

ledwards2225:
# Description

TLDR: Split Shplonk (just like we did for Gemini) to facilitate computation of commitments in work_queue.

The Shplonk prover has two main operations: 1) compute batched quotient polynomial Q, and 2) compute polynomial Q - Q_z. The challenge z at which Q is partially evaluated depends on the commitment [Q]. In order to compute this commitment in the work_queue, operations (2) must be separated from task (1). As part of this separation, all transcript operations have been moved to the `Prover` itself.

# Checklist:

### DIFF
--- a/cpp/src/barretenberg/honk/proof_system/prover.hpp
+++ b/cpp/src/barretenberg/honk/proof_system/prover.hpp
@@ -44,7 +44,6 @@ template <typename settings> class Prover {
     void execute_pcs_evaluation_round();
     void execute_shplonk_batched_quotient_round();
     void execute_shplonk_partial_evaluation_round();
-    // void execute_shplonk_round();
     void execute_kzg_round();
 
     void compute_wire_commitments();
--- a/cpp/src/barretenberg/honk/proof_system/prover.hpp
+++ b/cpp/src/barretenberg/honk/proof_system/prover.hpp
@@ -42,7 +42,9 @@ template <typename settings> class Prover {
     void execute_relation_check_rounds();
     void execute_univariatization_round();
     void execute_pcs_evaluation_round();
-    void execute_shplonk_round();
+    void execute_shplonk_batched_quotient_round();
+    void execute_shplonk_partial_evaluation_round();
+    // void execute_shplonk_round();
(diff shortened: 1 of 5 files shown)


cpp/src/barretenberg/honk/proof_system/prover.hpp
```diff
@@ -71,6 +73,9 @@ template <typename settings> class Prover {
     // Container for d + 1 Fold polynomials produced by Gemini
     std::vector<Polynomial> fold_polynomials;
 
+    Polynomial batched_quotient_Q;
```
ledwards2225, line 75: I wonder if a more proper way to handle this would be to have a `std::shared_ptr<Polynomial>`, rather than moving it and leaving an empty poly behind. Not a huge deal but curious about best practice here. Thoughts @adr1anh?

adr1anh: Maybe an optional here could more explicitly indicate whether the entry is used?

adr1anh: Shared pointer wouldn't make much sense since it's not actually shared anywhere. It also has a bit of overhead from having to do reference counting.

adr1anh: A unique pointer might make sense, but the main point it in checking that the element is null or not, in which case an optional is more explicit

adr1anh:
LGTM

adr1anh (approved):

cpp/src/barretenberg/honk/proof_system/prover.hpp
ledwards2225: @adr1anh I'm sure I'm clear on what you mean -- I haven't seen `optional` applied to a class member before, only an input/output to a function. Is this an option or am I misunderstanding you? Either way, I'll leave it as is for now since it's hardly dangerous and I'm guessing this gets refactored soon anyway :)

PR #324, opened 2023-04-05
Lde/reinstate work queue

### LINKED ISSUE #214: Re-introduce work queue
Update Honk so that all commitments are computed via the work queue.

ledwards2225:
# Description

Add a simplified `work_queue` for Honk. Use it to compute commitments throughout StandardHonk prover.

The new work queue for Honk is largely a copy-paste then simplify. The only real changes are
- work queue now owns a commitment key instead of calling pedersen directly. (This is instead of the Prover owning the proving key which now no longer makes sense since it is not responsible for commitments).
- simplified interface for adding things to the work queue, i.e. `add_commitment` instead of `add_to_queue(Lots of ugly details)`
- Remove some objects from `WorkItem` like the vague and no longer needed `index` and `constant` which were needed only for FFT operations

This PR also includes a simple change to the original `work_queue` that allows the MSM size to be set directly and flexibly rather than based on an enum. I made this change when I was still planning on using the original work queue for Honk (to accommodate Fold poly commitments). It's no longer necessary but seems like a nice simplifying change so I've left it in.

# Checklist:

### DIFF
--- a/cpp/src/barretenberg/honk/composer/composer_helper/standard_honk_composer_helper.cpp
+++ b/cpp/src/barretenberg/honk/composer/composer_helper/standard_honk_composer_helper.cpp
@@ -168,12 +168,6 @@ StandardProver StandardHonkComposerHelper<CircuitConstructor>::create_prover(
     auto manifest = Flavor::create_manifest(circuit_constructor.public_inputs.size(), num_sumcheck_rounds);
     StandardProver output_state(std::move(wire_polynomials), circuit_proving_key);
 
-    // TODO(Cody): This should be more generic
-    std::unique_ptr<pcs::kzg::CommitmentKey> kate_commitment_key =
-        std::make_unique<pcs::kzg::CommitmentKey>(circuit_proving_key->circuit_size, "../srs_db/ignition");
-
-    output_state.commitment_key = std::move(kate_commitment_key);
-
     return output_state;
 }
 template class StandardHonkComposerHelper<StandardCircuitConstructor>;
--- a/cpp/src/barretenberg/honk/pcs/kzg/kzg.test.cpp
+++ b/cpp/src/barretenberg/honk/pcs/kzg/kzg.test.cpp
@@ -40,7 +40,8 @@ TYPED_TEST(BilinearAccumulationTest, single)
 
     auto prover_transcript = ProverTranscript<Fr>::init_empty();
 
-    KZG::reduce_prove(this->ck(), opening_pair, witness, prover_transcript);
+    auto quotient_W = KZG::compute_opening_proof_polynomial(opening_pair, witness);
+    prover_transcript.send_to_verifier("KZG:W", this->commit(quotient_W));
 
     auto verifier_transcript = VerifierTranscript<Fr>::init_empty(prover_transcript);
     auto kzg_claim = KZG::reduce_verify(opening_claim, verifier_transcript);
@@ -148,7 +149,8 @@ TYPED_TEST(BilinearAccumulationTest, GeminiShplonkKzgWithShift)
 
     // KZG prover:
     // - Adds commitment [W] to transcript
-    KZG::reduce_prove(this->ck(), shplonk_opening_pair, shplonk_witness, prover_transcript);
+    auto quotient_W = KZG::compute_opening_proof_polynomial(shplonk_opening_pair, shplonk_witness);
+    prover_transcript.send_to_verifier("KZG:W", this->commit(quotient_W));
 
     // Run the full verifier PCS protocol with genuine opening claims (genuine commitment, genuine evaluation)
 
--- a/cpp/src/barretenberg/honk/proof_system/work_queue.hpp
+++ b/cpp/src/barretenberg/honk/proof_system/work_queue.hpp
@@ -16,7 +16,6 @@ template <typename Params> class work_queue {
     using FF = typename Params::Fr;
     using Commitment = typename Params::C;
 
-  public:
     struct work_item_info {
         uint32_t num_scalar_multiplications;
     };
@@ -27,12 +26,19 @@ template <typename Params> class work_queue {
         std::string label;
     };
 
+  private:
+    std::shared_ptr<proof_system::plonk::proving_key> key;
+    // TODO(luke): Consider handling all transcript interactions in the prover rather than embedding them in the queue.
+    proof_system::honk::ProverTranscript<FF>& transcript;
+    CommitmentKey commitment_key;
+    std::vector<work_item> work_item_queue;
+
+  public:
     explicit work_queue(std::shared_ptr<proof_system::plonk::proving_key>& proving_key,
-                        proof_system::honk::ProverTranscript<FF>& prover_transcript,
-                        std::shared_ptr<CommitmentKey>& commitment_key)
+                        proof_system::honk::ProverTranscript<FF>& prover_transcript)
         : key(proving_key)
         , transcript(prover_transcript)
-        , commitment_key(commitment_key){};
+        , commitment_key(key->circuit_size, "../srs_db/ignition"){}; // TODO(luke): make this properly parameterized
 
     work_queue(const work_queue& other) = default;
     work_queue(work_queue&& other) noexcept = default;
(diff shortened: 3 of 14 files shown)


cpp/src/barretenberg/honk/proof_system/work_queue.hpp
```diff
@@ -0,0 +1,143 @@
+#pragma once
```
codygunton, line 1: It seems like you're deciding (rightly IMO) that the honk work queue is going to not share code with the plonk work queue. In that case, will you move the work queue from `proof_system` to `plonk`? Otherwise we create tech debt.

cpp/src/barretenberg/plonk/proof_system/prover/prover.cpp
```diff
@@ -79,7 +79,7 @@ template <typename settings> void ProverBase<settings>::compute_wire_commitments
         barretenberg::fr* coefficients = key->polynomial_store.get(wire_tag).get_coefficients();
 
         // This automatically saves the computed point to the transcript
-        fr domain_size_flag = i > 2 ? work_queue::MSMType::MONOMIAL_N : work_queue::MSMType::MONOMIAL_N_PLUS_ONE;
+        fr domain_size_flag = i > 2 ? key->circuit_size : (key->circuit_size + 1);
```
codygunton, line 82: Right, this is a nice simplification.

cpp/src/barretenberg/honk/proof_system/work_queue.hpp
```diff
@@ -0,0 +1,143 @@
...
+#include "barretenberg/plonk/proof_system/proving_key/proving_key.hpp"
+#include <cstddef>
+#include <memory>
+
+namespace proof_system::honk {
+
+// Currently only one type of work queue operation but there will likely be others related to Sumcheck
+enum WorkType { SCALAR_MULTIPLICATION };
+
+template <typename Params> class work_queue {
```
codygunton, line 16: Is this the same sort of `Params` expected by some other class? Would be helpful to document that here if so.

cpp/src/barretenberg/honk/proof_system/work_queue.hpp
```diff
@@ -0,0 +1,143 @@
...
+    };
+
+    struct work_item {
+        WorkType work_type = SCALAR_MULTIPLICATION;
+        std::span<FF> mul_scalars;
+        std::string label;
+    };
+
+  private:
+    std::shared_ptr<proof_system::plonk::proving_key> key;
```
codygunton: Can we call this `proving_key` since there's another type of key floating around in this class?

cpp/src/barretenberg/honk/proof_system/work_queue.hpp
```diff
@@ -0,0 +1,143 @@
...
+        return work_item_info{ scalar_mul_count };
+    };
+
+    [[nodiscard]] FF* get_scalar_multiplication_data(size_t work_item_number) const
+    {
+        size_t count = 0;
+        for (const auto& item : work_item_queue) {
+            if (item.work_type == WorkType::SCALAR_MULTIPLICATION) {
+                if (count == work_item_number) {
+                    return const_cast<FF*>(item.mul_scalars.data());
```
codygunton, line 70: Why is this needed? I think it's easy to get undefined behavior when using `const_cast`. Can't we just return the span?

cpp/src/barretenberg/honk/proof_system/work_queue.hpp
```diff
@@ -0,0 +1,143 @@
...
+                if (count == work_item_number) {
+                    return item.mul_scalars.size();
+                }
+                ++count;
+            }
+        }
+        return 0;
+    };
+
+    void put_scalar_multiplication_data(const Commitment& result, size_t work_item_number)
```
codygunton, line 92: plonk passes by value. Are we sure we want a difference here?

cpp/src/barretenberg/honk/pcs/kzg/kzg.test.cpp
```diff
@@ -40,7 +40,8 @@ TYPED_TEST(BilinearAccumulationTest, single)
 
     auto prover_transcript = ProverTranscript<Fr>::init_empty();
 
-    KZG::reduce_prove(this->ck(), opening_pair, witness, prover_transcript);
+    auto quotient_W = KZG::compute_opening_proof_polynomial(opening_pair, witness);
```
codygunton, line 43: Really prefer this because I don't have to think about what the name means.

cpp/src/barretenberg/honk/composer/composer_helper/standard_honk_composer_helper.cpp
```diff
@@ -168,12 +168,6 @@ StandardProver StandardHonkComposerHelper<CircuitConstructor>::create_prover(
     auto manifest = Flavor::create_manifest(circuit_constructor.public_inputs.size(), num_sumcheck_rounds);
     StandardProver output_state(std::move(wire_polynomials), circuit_proving_key);
 
-    // TODO(Cody): This should be more generic
```
codygunton, line 171: ...becuase now the commitment key lives in the work queue ✔️

codygunton (approved):
Nice work. Asking for a discussion and for you to resolve some tech debt that this work would create.

cpp/src/barretenberg/honk/pcs/kzg/kzg.test.cpp
ledwards2225: You mean you prefer my change or the original?

cpp/src/barretenberg/honk/proof_system/work_queue.hpp
ledwards2225: That's right. Comment/TODO added for a bit more clarity

cpp/src/barretenberg/honk/proof_system/work_queue.hpp
ledwards2225: good call, done

cpp/src/barretenberg/honk/proof_system/work_queue.hpp
ledwards2225: Yeah I agree but two things: 1) this function is only used in the WASM/c_bind context where it expects an `fr*` so I didn't want to mess with it too much. 2) It may actually be necessary (at least for now) because the pedersen implementation makes heavy use of pointer arithmetic. (You'll notice that the CommitmentKey is also extracting a raw pointer from a span behind the scenes before passing to perdersen). I'm not sure how hard it would be to update pedersen to use a span somehow but needless to say this is not the time/place for that.

cpp/src/barretenberg/honk/proof_system/work_queue.hpp
ledwards2225: It's pretty small so probably not a big deal either way but I did this intentionally because all we're doing is serializing the commitment to the transcript. Its scope ends with the function itself so I think no need to pass by value.

cpp/src/barretenberg/honk/proof_system/work_queue.hpp
ledwards2225: good call, done

codygunton: Great, thanks.

cpp/src/barretenberg/honk/proof_system/work_queue.hpp
codygunton: Thanks.

cpp/src/barretenberg/honk/proof_system/work_queue.hpp
codygunton: Fair enough, but we should keep this in mind.

cpp/src/barretenberg/honk/pcs/kzg/kzg.test.cpp
codygunton: Yes, it requires much less thinking to  what your code is doing IMO

PR #331, opened 2023-04-07
`get_limbs` methods on bigfield, biggroup.

suyash67:
# Description

We need the limbs of bigfield elements for hashing in aztec3-circuits. This PR adds simple methods to get limb values on bigfield and biggroup elements. Need to add tests.

# Checklist:

### DIFF
--- a/cpp/src/barretenberg/ecc/groups/affine_element.hpp
+++ b/cpp/src/barretenberg/ecc/groups/affine_element.hpp
@@ -43,6 +43,11 @@ template <typename Fq, typename Fr, typename Params> class alignas(64) affine_el
               typename CompileTimeEnabled = std::enable_if_t<(BaseField::modulus >> 255) == uint256_t(0), void>>
     constexpr uint256_t compress() const noexcept;
 
+    /**
+     * @brief Get the limbs of the x and y coordinates of the group element.
+     *
+     * @return vector of the limbs of x, y coordinates (size 8)
+     */
     std::vector<barretenberg::fr> get_coordinate_limbs() const
     {
         constexpr size_t num_limb_bits = proof_system::plonk::NUM_LIMB_BITS_IN_FIELD_SIMULATION;
--- a/cpp/src/barretenberg/ecc/groups/affine_element.hpp
+++ b/cpp/src/barretenberg/ecc/groups/affine_element.hpp
@@ -45,7 +45,7 @@ template <typename Fq, typename Fr, typename Params> class alignas(64) affine_el
 
     std::vector<barretenberg::fr> get_coordinate_limbs() const
     {
-        constexpr size_t num_limb_bits = plonk::NUM_LIMB_BITS_IN_FIELD_SIMULATION;
+        constexpr size_t num_limb_bits = proof_system::plonk::NUM_LIMB_BITS_IN_FIELD_SIMULATION;
         const auto split_bigfield_limbs = [](const uint256_t& element) {
             std::vector<barretenberg::fr> limbs;
             limbs.push_back(element.slice(0, num_limb_bits));
--- a/cpp/src/barretenberg/ecc/groups/affine_element.hpp
+++ b/cpp/src/barretenberg/ecc/groups/affine_element.hpp
@@ -3,6 +3,8 @@
 #include <vector>
 #include <type_traits>
 #include "barretenberg/ecc/curves/bn254/fq2.hpp"
+#include "barretenberg/plonk/proof_system/constants.hpp"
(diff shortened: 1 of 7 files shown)


iAmMichaelConnor (approved):

cpp/src/barretenberg/stdlib/primitives/bigfield/bigfield.hpp
```diff
@@ -160,7 +160,7 @@ template <typename Composer, typename T> class bigfield {
         field_t<Composer> lo = binary_basis_limbs[0].element + (binary_basis_limbs[1].element * shift_1);
         field_t<Composer> hi = binary_basis_limbs[2].element + (binary_basis_limbs[3].element * shift_1);
         // n.b. this only works if NUM_LIMB_BITS * 2 is divisible by 8
-        ASSERT((NUM_LIMB_BITS / 8) * 8 == NUM_LIMB_BITS);
+        ASSERT((NUM_LIMB_BITS * 2 / 8) * 8 == NUM_LIMB_BITS * 2);
```
suyash67, line 163: @Rumata888 pls check this.

cpp/src/barretenberg/ecc/groups/affine_element.hpp
```diff
@@ -41,6 +43,31 @@ template <typename Fq, typename Fr, typename Params> class alignas(64) affine_el
               typename CompileTimeEnabled = std::enable_if_t<(BaseField::modulus >> 255) == uint256_t(0), void>>
     constexpr uint256_t compress() const noexcept;
 
+    /**
+     * @brief Get the limbs of the x and y coordinates of the group element.
+     *
+     * @return vector of the limbs of x, y coordinates (size 8)
+     */
+    std::vector<barretenberg::fr> get_coordinate_limbs() const
```
suyash67, line 51: @Rumata888 question: we might not end up using these methods in the immediate term in aztec3. do you think we should still keep these? Would they be useful anyway?

Rumata888: I think they are misplaced here. I'd add them to the bigfield files as a templated function. If they are there, then they'd stay useful

cpp/src/barretenberg/ecc/groups/affine_element.hpp
```diff
@@ -41,6 +43,31 @@ template <typename Fq, typename Fr, typename Params> class alignas(64) affine_el
...
+        const auto split_bigfield_limbs = [](const uint256_t& element) {
+            std::vector<barretenberg::fr> limbs;
+            limbs.push_back(element.slice(0, num_limb_bits));
+            limbs.push_back(element.slice(num_limb_bits, num_limb_bits * 2));
+            limbs.push_back(element.slice(num_limb_bits * 2, num_limb_bits * 3));
+            limbs.push_back(element.slice(num_limb_bits * 3, num_limb_bits * 4));
+            return limbs;
+        };
+
+        std::vector<barretenberg::fr> x_limbs = split_bigfield_limbs(this->x);
```
Rumata888, line 63: This could become an issue if the point is a point at infinity because of the way we encode this case. I'd recommend failing in that case

cpp/src/barretenberg/stdlib/primitives/bigfield/bigfield.hpp
Rumata888: Yes, it seems that this was a typo. It even caused the fuzzer to fail.

cpp/src/barretenberg/stdlib/primitives/bigfield/bigfield.hpp
```diff
@@ -312,6 +312,20 @@ template <typename Composer, typename T> class bigfield {
         prime_basis_limb.fix_witness();
     }
 
+    /**
+     * @brief Get the binary basis limbs of a bigfield scalar.
+     *
+     * @return vector of the binary basis limbs (circuit type)
+     */
+    std::vector<field_t<Composer>> get_limbs() const
```
Rumata888, line 320: If you intend to use these limbs in some way, you need to forcefully reduce the element. Otherwise there could be several limb values representing the same underlying value.

Rumata888 (changes requested):
There are three key points that it would be good to address:
1) Moving splitting logic into bigfield or biggroup from affine_element
2) Adding an edge condition for the infinity case
3) Reducing the bigfield element and using assert_is_in_field to ensure limb uniqueness

suyash67:
This work is not needed for aztec3 for now. If we do in future, we'll revisit this later.

PR #332, opened 2023-04-07
fix: Change to get_num_gates() inside get_total_circuit_size()

vezenovm:
# Description

In the DSL package we expose a method `get_total_circuit_size` in the `acir_proofs` folder. In Noir's `aztec_backend` we use this method to estimate the number of G1 points we should take from CRS to set up input to the bberg API such as Pippenger.  This circuit size should match the `total_num_gates` variable at the top of the `compute_proving_key_base` method. This estimation is currently off causing a mismatch between the size of the CRS used for proving and the one computed in the proving key. 

`6_array` in the Noir integrations was having a seg fault due to this mismatch. The minimum circuit size used to set the subgroup size in the proving key was greater than the original number of gates in the circuit (`size_t num_gates` field in the composer). `get_total_circuit_size` should have used `get_num_gates()` internally to fetch the same number of gates used as the proving key. In `compute_proving_key` the method first adds the gates related to ROM arrays and range lists before calling `compute_proving_key_base`. The circuit has not yet been finalized when we call `get_total_circuit_size` and only returns the original number of gates in the circuit. 

# Checklist:

- [] I have [linked](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) this pull request to the issue(s) that it resolves.
- [X] There are no circuit changes, OR specifications in `/markdown/specs` have been updated.
- [X] There are no circuit changes, OR a cryptographer has been assigned for review.

### DIFF
--- a/cpp/src/barretenberg/plonk/composer/ultra_composer.hpp
+++ b/cpp/src/barretenberg/plonk/composer/ultra_composer.hpp
@@ -353,7 +353,7 @@ class UltraComposer : public ComposerBase {
         }
 
         auto minimum_circuit_size = tables_size + lookups_size;
-        auto num_filled_gates = num_gates + public_inputs.size();
+        auto num_filled_gates = get_num_gates() + public_inputs.size();
         return std::max(minimum_circuit_size, num_filled_gates);
     }
 


phated (approved):

phated:
Tested this in noir integration tests and it is working for us

ledwards2225 (approved):
Yup, I think you've accurately described the problem and this is the right fix. This aspect of Ultra is a bit brittle and not as clear as it could be -- something I'd like to address soon. Good catch!

Issue #342, opened 2023-04-10
New Relations for UltraHonk


ledwards2225:
Add the new relations required for construction/verification of Ultra Honk proofs. Test each relation independent of proof construction.

PR #347, opened 2023-04-12
Permutation grand product relation for Ultra Honk

ledwards2225:
# Description

Add the permutation grand product construction relation for Ultra Honk. Part of issue #342 

This work includes two standalone tests (similar to what we have for all other relations), one for "consistency", i.e. checking that the relation arithmetic matches a simplified unoptimized version (useful once we actually optimize the relations), and another for checking "correctness", i.e. is the constraint encoded by the relation satisfied for a real circuit. 

I've done some reorganization of the relation tests to reflect these two different types of tests. We now have a `RelationConsistency` test suite and a `RelationCorrectness` test suite.

Note: We should be able to combine the two different permutation grand product construction relations (width 3 and width 4) into a single templated relation later on if desired. This is tricky now because of the rigid way we access polynomials but should be made easier by Cody's flavor updates.

# Checklist:


cpp/src/barretenberg/honk/composer/standard_honk_composer.test.cpp
```diff
@@ -304,109 +304,6 @@ TEST(StandardHonkComposer, VerificationKeyCreation)
               composer.circuit_constructor.selectors.size() + composer.num_wires * 2 + 2);
 }
 
-/**
```
ledwards2225, line 307: Simply moved this test to the RelationCorrectness suite

cpp/src/barretenberg/honk/composer/ultra_honk_composer.test.cpp
```diff
@@ -41,12 +41,13 @@ std::vector<uint32_t> add_variables(auto& composer, std::vector<fr> variables)
  */
 void verify_consistency(honk::UltraProver& honk_prover, plonk::UltraProver& plonk_prover)
 {
-    // Check that all lagrange polys agree
+    // Check that all lagrange polys agree (except for sigmas/ids which are constructed differently for Honk)
```
ledwards2225, line 44: I'm keeping these plonk-honk consistency checks around for now but they're slowly becoming less relevant (and will eventually be replaced with the normal composer tests). Here for example I've removed the consistency check on the permutation polys since we construct them differently for plonk and honk.

cpp/src/barretenberg/honk/flavor/flavor.hpp
```diff
@@ -187,4 +187,56 @@ struct StandardHonk {
         return output;
     }
 };
+
```
ledwards2225, line 190: This will be replaced by some equivalent with Cody's impending overhaul.

PR #348, opened 2023-04-12
Address potential security bug and improve testing for the logic constraint function

maramihali:
# Description
Ensures the result of the `create_logic_constraint` function is indeed produced using the left and right operand that are passed as arguments. 

Adds test to highlight a potential issue with the range constraints in the ultra composer.
# Checklist:


cpp/src/barretenberg/plonk/composer/ultra_composer.test.cpp
```diff
@@ -793,4 +793,25 @@ TYPED_TEST(ultra_composer, range_checks_on_duplicates)
     TestFixture::prove_and_verify(composer, /*expected_result=*/true);
 }
 
+TEST(ultra_composer, range_constraint_problem)
```
maramihali: - waiting to get more insight into this with help from Zac; could be removed from this PR

cpp/src/barretenberg/stdlib/primitives/logic/logic.test.cpp
```diff
@@ -19,81 +31,139 @@ template <class T> void ignore_unused(T&) {} // use to ignore unused variables i
...
+        run_test(8, composer);
+    }
+    auto prover = composer.create_prover();
+    plonk::proof proof = prover.construct_proof();
+    auto verifier = composer.create_verifier();
+    bool result = verifier.verify_proof(proof);
+    EXPECT_EQ(result, true);
+}
+
+// Tests the constraints will fail if the operands are larger than expected even though the result contains the correct
```
maramihali, line 98: I would actually like some opinions on what the behaviour should be if we provide `a` and `b` to be larger than `num_bits` wdyt?

cpp/src/barretenberg/stdlib/primitives/logic/logic.test.cpp
```diff
@@ -1,13 +1,25 @@
...
 #include "logic.hpp"
-#include "barretenberg/plonk/proof_system/constants.hpp"
 #include <gtest/gtest.h>
 #include "barretenberg/honk/composer/standard_honk_composer.hpp"
 #include "barretenberg/plonk/composer/standard_composer.hpp"
 #include "barretenberg/plonk/composer/ultra_composer.hpp"
 #include "barretenberg/plonk/composer/turbo_composer.hpp"
 #include "barretenberg/numeric/random/engine.hpp"
 
+#pragma GCC diagnostic ignored "-Wimplicitly-unsigned-literal"
```
codygunton: I suggested ignoring the unused local typedefs warning because it gave a systematic simplification and there is no risk when creating aliases that are not used, but in general we should remove these sorts of warnings before merging. this one in particular is breaking gcc compilation, so it'll have to go.

cpp/src/barretenberg/stdlib/primitives/logic/logic.test.cpp
codygunton: I think the XOR/AND should be the truncation of the XOR/AND of the two inputs, and the and the proof should pass. Thoughts?

cpp/src/barretenberg/plonk/composer/ultra_composer.test.cpp
codygunton: Tests are often the best documentation, so I'd say leave it in and add some comments explaining the issue. One step further, I would implement the test using a lambda `run_test(bool add_a_to_execution_trace, bool add_b_to_execution_trace, bool add_c_to_execution_trace)` and illustrate what happens when various combinations of gates are created.

cpp/src/barretenberg/stdlib/primitives/logic/logic.test.cpp
maramihali: Yes, thanks for pointing that out, it can be removed.

cpp/src/barretenberg/stdlib/primitives/logic/logic.test.cpp
maramihali: Ok, so we should  shorten `a` and `b` to `num_bits` and then applying the range constraint?

cpp/src/barretenberg/plonk/composer/ultra_composer.test.cpp
maramihali: I added a TODO because some of the details of the issue are still unclear, is that fine?

maramihali: issue fixed

cpp/src/barretenberg/stdlib/primitives/logic/logic.test.cpp
```diff
@@ -19,81 +29,137 @@ template <class T> void ignore_unused(T&) {} // use to ignore unused variables i
...
+
+        EXPECT_EQ(uint256_t(xor_result.get_value()), xor_expected);
+        EXPECT_EQ(uint256_t(xor_result_left_constant.get_value()), xor_expected);
+        EXPECT_EQ(uint256_t(xor_result_right_constant.get_value()), xor_expected);
+        EXPECT_EQ(uint256_t(xor_result_both_constant.get_value()), xor_expected);
+    };
+
+    auto composer = Composer();
+    for (size_t i = 8; i < 248; i += 8) {
+        run_test(8, composer);
```
codygunton, line 89: Looks like this test is only being run for one value! :crossed_fingers:

codygunton: Resolved in #432

Issue #217, opened 2023-03-09
TODO: Evaluate method for ensuring no zero polynomial in Honk


ledwards2225:
During Honk development, we added a dummy gate in the circuit constructor to ensure we could not encounter a zero polynomial. Determine whether this is acceptable or if there is a better option.

codygunton:
Per Adrian's comment, perhaps we should alter this gate so that all selector values are distinct?

codygunton:
Spawned #423

PR #351, opened 2023-04-12
Ultra Honk arithmetic and grand product relations

ledwards2225:
# Description

Add the arithmetic and permutation grand product relations for Ultra Honk. (Total of 4 relations) (Part of issue https://github.com/AztecProtocol/barretenberg/issues/342)

New relations (all names are fully flexible at this point):
- `UltraArithmeticRelation`: Primary arithmetic relation
- `UltraArithmeticRelationSecondary`: Secondary arithmetic relation for non-native field op applications
- `UltraGrandProductInitializationRelation`: Equivalent to Standard version
- `UltraGrandProductComputationRelation`: Equivalent to Standard version

This work also adds tests for all of these relations, following the format of the existing relations tests. I've reorganized all relations tests into two suites: 
- `RelationConsistency`: Check that arithmetic in the relation is equivalent to some simplified implementation of the same arithmetic. (Useful for when we start optimizing).
- `RelationCorrectness`: Check that the constraint encoded in each relation is actually satisfied for the polynomials produced by the composer for a real circuit.

Other changes:
- Some templatizing of test fixtures in `RelationConsistency` to account for different flavor-esque parameters
- I've narrowed the scope of the Plonk-Honk composer consistency tests to account for some correct and expected differences between the polynomials produced by the respective composers. They're still useful for checking the witnesses and table polys but are close to the end of their useful life.

Notes:
- We may be able to combine the permutation grand product relations for Standard/Ultra via templating later on if desired. This is tricky now because of the rigid way we access polynomials but should be made easier by Cody's flavor updates. On the other hand, a `Relation` is a pretty simple object and it might not hurt to have some code duplication there for simplicity.
- All of the naming in this PR is up for debate. I anticipate making some changes and am open to suggestions

# Checklist:


cpp/src/barretenberg/honk/composer/composer_helper/ultra_honk_composer_helper.cpp
```diff
@@ -201,11 +201,15 @@ std::shared_ptr<plonk::proving_key> UltraHonkComposerHelper<CircuitConstructor>:
 
     construct_lagrange_selector_forms(circuit_constructor, circuit_proving_key.get());
 
-    enforce_nonzero_polynomial_selectors(circuit_constructor, circuit_proving_key.get());
+    // TODO(#217)(luke): Naively enforcing non-zero selectors for Honk will result in some relations not being
+    // satisfied.
+    // enforce_nonzero_polynomial_selectors(circuit_constructor, circuit_proving_key.get());
```
Rumata888, line 206: If we do this, we'll have to add the zero knowledge relation before we can create a valid proof, since selectors could easily be zero.

cpp/src/barretenberg/honk/composer/ultra_honk_composer.test.cpp
```diff
@@ -39,14 +39,19 @@ std::vector<uint32_t> add_variables(auto& composer, std::vector<fr> variables)
...
 {
     // Check that all lagrange polys agree
     auto& honk_store = honk_prover.key->polynomial_store;
     auto& plonk_store = plonk_prover.key->polynomial_store;
     for (auto& entry : honk_store) {
         std::string key = entry.first;
-        if (plonk_store.contains(key)) {
+        bool is_sorted_table = (key.find("s_") != std::string::npos);
+        bool is_table = (key.find("table_value_") != std::string::npos);
+        if (plonk_store.contains(key) && (is_sorted_table || is_table)) {
```
Rumata888, line 54: Why are you not checking wires any more?

cpp/src/barretenberg/honk/flavor/flavor.hpp
```diff
@@ -187,4 +187,58 @@ struct StandardHonk {
...
 };
+
+struct UltraArithmetization {
+    /**
+     * @brief All of the multivariate polynomials used by the Ultra Honk Prover.
+     * @details The polynomials are broken into three categories: precomputed, witness, and shifted.
+     * This separation must be maintained to allow for programmatic access, but the ordering of the
+     * polynomials can be permuted within each category if necessary. Polynomials can also be added
+     * or removed (assuming consistency with the prover algorithm) but the constants describing the
+     * number of poynomials in each category must be manually updated.
```
Rumata888: Typo "poynomials"

cpp/src/barretenberg/honk/flavor/flavor.hpp
```diff
@@ -187,4 +187,58 @@ struct StandardHonk {
...
+        W_O,
+        W_4,
+        S_1,
+        S_2,
+        S_3,
+        S_4,
+        Z_PERM,
+        Z_LOOKUP,
+        /* --- SHIFTED POLYNOMIALS --- */
+        W_1_SHIFT,
```
Rumata888, line 236: W_L_SHIFT?

Rumata888 (approved):
I am a bit worried about the verify_consistency function, otherwise everything looks great

cpp/src/barretenberg/honk/flavor/flavor.hpp
ledwards2225: `W_1_SHIFT` was intentional since I prefer the numerical index but you're right I'm not being consistent here. I'll make them all numeric in an immediate follow on, but regardless, all of this Flavor stuff I'm adding will go away entirely with Cody's incoming flavor work.

cpp/src/barretenberg/honk/flavor/flavor.hpp
ledwards2225: Fixed, thanks

cpp/src/barretenberg/honk/composer/ultra_honk_composer.test.cpp
ledwards2225: The wires are checked just below this. It has to be done in a separate loop since the wires are owned directly by the prover in Honk but they are still stored in the proving_key for Plonk. I've added a comment to clarify.

cpp/src/barretenberg/honk/composer/composer_helper/ultra_honk_composer_helper.cpp
ledwards2225: Its a very good point. We need to make sure our tests are not passing trivially with zero selectors until we decide how to proceed. Maybe we should prioritize issue #217 in the near future.

PR #350, opened 2023-04-12
32-Byte Keccak256 challenges for UltraPlonK

codygunton:
# Description
Adds prover and verifier settings to generate verifier 32-byte verifier challenges using Keccak256. We follow the model set by @iAmMichaelConnor and @suyash67 in their work on standard plonk verification of ultra plonk proofs.

# Checklist:

- [x] I have reviewed my diff in github, line by line.
- [x] Every change is related to the PR description.
- [ ] I have [linked](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) this pull request to the issue(s) that it resolves.
- [x] There are no unexpected formatting changes, superfluous debug logs, or commented-out code.
- [x] There are no circuit changes, OR specifications in `/markdown/specs` have been updated.
- [x] There are no circuit changes, OR a cryptographer has been assigned for review.
- [ ] I've updated any terraform that needs updating (e.g. environment variables) for deployment.
- [x] The branch has been rebased against the head of its merge target.
- [x] I'm happy for the PR to be merged at the reviewer's next convenience.
- [ ] New functions, classes, etc. have been documented according to the doxygen comment format. Classes and structs must have `@brief` describing the intended functionality.
- [x] If existing code has been modified, such documentation has been added or updated.

PR #352, opened 2023-04-12
No `SYSTEM_COMPOSER`

### LINKED ISSUE #190: Remove `SYSTEM_COMPOSER`
This `SYSTEM_COMPOSER` variable was introduced by @suyash67 to ease the process of changing the underlying proof system of the Aztec Connect rollup (pre-root verifier circuit). It seems to me this feature is now just noise, so I propose to remove it from Barretenberg. It's possible that we should reinstate it in some other form in Aztec 3, though there we will be blending various proof systems and curves, so something more robust will be needed. Thoughts @suyash67 @iAmMichaelConnor?

codygunton:
# Description

This PR resolves https://github.com/AztecProtocol/barretenberg/issues/190. As explained there, the `SYSTEM_COMPOSER` global variable is a piece of tech debt. It complicated the work of #350, leading to more tech debt (use of special Keccak prover and verifier in undesired places), so I decided to take on the aforementioned issue.

`SYSTEM_COMPOSER` determined the composer that was used in the following places:
 - The DSL module.
 - The plonk prover cbinds.
 - The join_split_example.
 - Several standard library test suites.
 - Some of our benchmarking suites.

The refactor takes the following approach:
 - Delete stdlib/types/types.hpp and replace by two similar, smaller files, one specific to the DSL module, one specific to the join_split_example.
 - Manually set `UltraComposer` to be the default inthe plonk prover cbinds. Delete the unrolled cbinds--there is no linearized/unrolled distinction anymore. c_bind_unrolled.cpp was a strict subset of what was in c_bind.cpp.
 - Manually set `UltraComposer` to be the default in tests and benchmark suites where previously `SYSTEM_COMPOSER` was used.

 It was natural to do some cleanup along the way. Certain `using namespace` invocations lead to clashes with existing namespaces in the standard library, so I removed those. I removed lots of unneeded `include`s. 
 
 I had to add some includes in places where previously we were over-`include`ing due to our use of `stdlib/types/types.hpp`, which itself `include`d nearly all of the standard library. 

I cleaned up the use of `UltraWithKeccak` prover and verifier in places outside of the DSL model where it it was temporarily in use for expediency. It is still tested in the `UltraComposer` test suites and in the DSL module.

 If it is in fact desirable to couple the DSL composer and the composer of the plonk cbindings, I'm happy to tweak this PR or have someone else make necessary adjustments. My primary goal was to get rid of the global variable so we wouldn't have to work around it join-split example, stdlib tests, and benchmarks. 

# Checklist:


ludamad (approved):
Looks good to my eyes

cpp/src/barretenberg/stdlib/hash/sha256/sha256.bench.cpp
```diff
@@ -2,10 +2,15 @@
...
+#include "barretenberg/plonk/proof_system/prover/prover.hpp"
+#include "barretenberg/stdlib/primitives/packed_byte_array/packed_byte_array.hpp"
 
 using namespace benchmark;
-using namespace proof_system::plonk::stdlib::types;
+using namespace proof_system::plonk;
+
+using LocalComposer = UltraComposer;
+using LocalProver = UltraProver;
+using LocalVerifier = UltraVerifier;
```
dbanks12: What does "LocalComposer" mean here versus just "Composer" in `sha256.test.cpp`?

cpp/src/barretenberg/stdlib/hash/sha256/sha256.test.cpp
```diff
@@ -169,15 +179,15 @@ TEST(stdlib_sha256, test_duplicate_proving_key)
...
 {
-    typedef stdlib::field_t<plonk::UltraComposer> field_pt;
-    typedef stdlib::packed_byte_array<plonk::UltraComposer> packed_byte_array_pt;
+    typedef plonk::stdlib::field_t<plonk::UltraComposer> field_pt;
+    typedef plonk::stdlib::packed_byte_array<plonk::UltraComposer> packed_byte_array_pt;
 
     // 55 bytes is the largest number of bytes that can be hashed in a single block,
     // accounting for the single padding bit, and the 64 size bits required by the SHA-256 standard.
-    plonk::UltraComposer composer = UltraComposer();
+    plonk::UltraComposer composer = plonk::UltraComposer();
```
dbanks12: Could consider just `plonk::UltraComposer composer;` or `auto composer = plonk::UltraComposer();`. Doesn't really matter, just good to be consistent across bberg.

cpp/src/barretenberg/stdlib/merkle_tree/hash.hpp
```diff
@@ -69,11 +57,7 @@ inline std::vector<barretenberg::fr> compute_tree_native(std::vector<barretenber
     while (layer.size() > 1) {
         std::vector<barretenberg::fr> next_layer(layer.size() / 2);
         for (size_t i = 0; i < next_layer.size(); ++i) {
-            if (plonk::SYSTEM_COMPOSER == ComposerType::PLOOKUP) {
-                next_layer[i] = crypto::pedersen_hash::lookup::hash_multiple({ layer[i * 2], layer[i * 2 + 1] });
-            } else {
-                next_layer[i] = crypto::pedersen_hash::hash_multiple({ layer[i * 2], layer[i * 2 + 1] });
-            }
+            next_layer[i] = crypto::pedersen_hash::lookup::hash_multiple({ layer[i * 2], layer[i * 2 + 1] });
```
dbanks12, line 60: We are using the non-lookup implementations of these functions in a3-circuits (and therefore in a3-packages TS too) as we are _not_ using UltraPlonk. Maybe @suyash67 can chime in here.

dbanks12:
Looks mostly good to me. Just have a few comments. Might be good to have @suyash67 review some of the ecdsa related changes since he was working on that recently. And I believe we still need turbo versions of the merkle hashing functions, at least until the A3 codebase moves to UltraPlonk.

cpp/src/barretenberg/plonk/proof_system/prover/c_bind.cpp
```diff
@@ -48,8 +47,8 @@ WASM_EXPORT void* test_env_load_prover_crs(size_t num_points)
 {
     return env_load_prover_crs(num_points);
 }
-typedef std::conditional_t<plonk::SYSTEM_COMPOSER == ComposerType::TURBO, plonk::TurboProver, plonk::UltraProver>
-    WasmProver;
+
+using WasmProver = plonk::UltraProver;
```
dbanks12, line 51: We are not yet using the WasmProver in A3, but we might in the future. Does this hardcode the WasmProver to UltraPlonk? Do we still want the ability to use it with Turbo? Does the Noir team need the ability to use these bindings with Turbo if a user chooses to use Turbo instead of Ultra @phated?

phated: We currently don't use this file/interface at all, but it looks like it's hardcoding the prover

cpp/src/barretenberg/stdlib/hash/sha256/sha256.test.cpp
codygunton: Thanks, I agree. I went with `auto` because that's what we use in the stdlib primitives tests.

cpp/src/barretenberg/stdlib/hash/sha256/sha256.bench.cpp
codygunton: Well, I couldn't use alias `Prover` because that clashes with something in the namespace `proof_system::plonk` so I called that `LocalProver` and it propagated that way. I was getting some weird linker error from Google benchmark but I nixed the `using` invocation now and added some qualifiers to achieve more consistent aliasing.

cpp/src/barretenberg/stdlib/merkle_tree/hash.hpp
codygunton: Oh, ok, but then I guess you're not using these funtions? Or do you reset `SYSTEM_COMPOSER` in your branch? It seems to me the simplest thing here if you want easy switching between the two would be `template <bool use_lookup>` on these functions and then to set some some a3-circuits global.

cpp/src/barretenberg/plonk/proof_system/prover/c_bind.cpp
codygunton: OK, good to know, so if @ludamad is happy then it's probably fine to fix this. As for Noir, there is no Turbo verifier smart contract atm, so I think there is no short-term need to switch to Turbo.

cpp/src/barretenberg/stdlib/merkle_tree/hash.hpp
dbanks12: We set USE_TURBO and then make calls to these from our C++ and then a3-packages makes calls to pedersen cbinds directly. I guess we should duplicate the cbinds and have ultra+turbo versions?

phated: This work should eliminate my USE_TURBO hack, right? 🤞

cpp/src/barretenberg/plonk/proof_system/prover/c_bind.cpp
ludamad: I'm not clear when we might want to use WasmProver so not sure I have context

cpp/src/barretenberg/stdlib/merkle_tree/hash.hpp
dbanks12: I believe it can. @codygunton if that is the intention then it should be removed from our CMakeFiles too. 

But for the time being we would need to template the functions on composer type like Cody mentioned and duplicate the pedersen cbinds that currently change functionality/output based on composer type.

cpp/src/barretenberg/plonk/proof_system/prover/c_bind.cpp
dbanks12: If the Noir team isn't using it and doesn't plan to, then my intuition is that it won't be important to us either. I think we should try to use a similar proving process as the Noir team from a3 TS.

So this diff is probably fine.

cpp/src/barretenberg/dsl/acir_format/sha256_constraint.cpp
```diff
@@ -1,11 +1,12 @@
 #include "sha256_constraint.hpp"
 #include "round.hpp"
 #include "barretenberg/stdlib/hash/sha256/sha256.hpp"
-
-using namespace proof_system::plonk::stdlib::types;
+#include "barretenberg/dsl/types.hpp"
 
 namespace acir_format {
 
+// using namespace ;
```
suyash67: Commented out code can be removed?

cpp/src/barretenberg/dsl/types.hpp
```diff
@@ -0,0 +1,63 @@
...
+#include "barretenberg/stdlib/merkle_tree/hash_path.hpp"
+#include "barretenberg/stdlib/encryption/schnorr/schnorr.hpp"
+#include "barretenberg/stdlib/primitives/curves/bn254.hpp"
+#include "barretenberg/stdlib/primitives/curves/secp256k1.hpp"
+#include "barretenberg/stdlib/primitives/memory/rom_table.hpp"
+namespace acir_format {
+
+using Composer = plonk::UltraComposer;
+
+using Prover = std::conditional_t<
```
vezenovm, line 26: Thinking on adding recursion later, how simple would it be to substitute `UltraWithKeccakProver` with `UltraProver` to generate a proof using pedersen-blake3, while still maintaining an `UltraWithKeccakProver` in order to be able to verify using the solidity verifier? Would we need two different `Prover` types?

vezenovm (approved):
This passes Noir integration tests. Just one comment on possible future challenges with the DSL package

cpp/src/barretenberg/plonk/proof_system/prover/c_bind.cpp
vezenovm: We don't use it at the moment but @TomAFrench might have some more insight on this as there were discussions about passing a pointer to the prover in order to enable multithreading with wasm. For now though we do not need the ability to use these bindings with Turbo and I think the diff is fine as is.

cpp/src/barretenberg/stdlib/merkle_tree/hash.hpp
suyash67: The conditional logic here was done so that the tests would pass if you switch composer from ultra to turbo or vice-versa. Otherwise, we'd have to manually fix the failing tests by calling the correct function.

Therefore, although we use both versions of Pedersen hash in A3, the stdlib handles the abstraction of when to use which one depending on a constant in composers. This isn't related to stdlib pedersen hash and commitment gadgets.

For now, this looks fine to me. We'll have to manually change these if/when we test things with different composers. As for Cody's suggestion of templated functions, that sounds good to me (whenever we implement it). Minor correction would be, instead of a boolean template parameter, we should parametrise it using an enum: `PedersenType::Lookup/FixedBase/SomethingElse`

suyash67:
Looks good to me. Important work this, been meaning to do this since a couple of weeks, thanks for taking this on @codygunton. 

A few comments (not necessarily related to the PR changes but just recording my thoughts).
1. We will have lot of code-duplication if we have `types.hpp` file for each module in bb. I was thinking if there's a way to templatise a typedef to avoid duplication. Not sure if C++ allows that plus we can do that in a future PR.
2. Ideally, we should test all stdlib functions with all composers. This also has been long due and good to see @maramihali looking into improving our test stdlib test suite.

cpp/src/barretenberg/dsl/types.hpp
suyash67: Yes I think we should declare two provers:
`IntermediateProver = UltraProver`
`FinalProver = UltraWithKeccakProver`
Better to use explicit prover type when testing too.

cpp/src/barretenberg/stdlib/merkle_tree/hash.hpp
codygunton: Thanks all, I also got rid of the various `USE_TURBO` macro invocations.

@suyash67 yes, ease of swapping out composers (or perhaps circuit constructors, once those are introduced, depending on the application) in many places is definitely a high priority for us, and I think templating (already our idiom in most places) should be embraced over macros and global switches in preprocessing. We get into a little trouble sometimes with placing things in namespaces rather than in classes (you can't template a namespace), but there are solutions, e.g., the sort of templating you suggest. I agree an enum is better than a bool, but IMO templating by some class and then using `std::is_same_v` is preferable to avoid loose coupling. E.g., compile-time checking of the actual composer rather than relying on the loosely-coupled structure `ComposerType` seems generally superior, IMO. (Possible notable exception: I think we now put `ComposerType` in the vk? `ComposerType` might need to stick around for that purpose).

cpp/src/barretenberg/dsl/acir_format/sha256_constraint.cpp
codygunton: Indeed, thanks.

cpp/src/barretenberg/dsl/types.hpp
codygunton: Agree

codygunton:
@suyash67 
> Looks good to me. Important work this, been meaning to do this since a couple of weeks, thanks for taking this on @codygunton.
> 
> A few comments (not necessarily related to the PR changes but just recording my thoughts).
> 
> 1. We will have lot of code-duplication if we have `types.hpp` file for each module in bb. I was thinking if there's a way to templatise a typedef to avoid duplication. Not sure if C++ allows that plus we can do that in a future PR.

Yeah, I agree we want to avoid code duplication is many cases, but for me it's acceptable here for two reasons: (1) the code doesn't actually contain any logic (just type alises); (2) we shouldn't avoid code duplication at the expense of coupling parts of the system that are essentially unrelated (indeed, dsl module and join-split example are only related in the sense that they use the standard library). Perhaps there is a smart solution that embeds aliases in the standard library? You can ofc define a struct like 

    namespace proof_system::plonk::stdlib {
    template <typename Composer> struct TypeAliases {
        using witness = witness_t<Composer>;
        using field = witness_t<Composer>;
    }
    } // namespace proof_system::plonk::stdlib

and then later you could have, for instance,

    namespace acir_format {
        // trying to reduce verbosity with weird alias
        using ct_ = TypeAliases<Composer>;
        ...
        typename ct_::field(ctx, typename ct_::witness(val));
        ...
    } // namespace acir_format

but I think there's no way to get rid of the `typename`s, so it's pointless since the goal is to reduce verbosity (we save 9 keystrokes by typing `c` in `_ct` instead of `_t<Composer>`). If one-letter identifiers weren't so bad, we could maybe make a concept called `Composer`, then we could have `template <Composer C>` and type `field_t<C>(ctx, witness_t<C>(val))`, or maybe commit to a short name like `CC` or `CT`. Anyway, yeah, idk, I've rambled enough :)

> 
> 2. Ideally, we should test all stdlib functions with all composers. This also has been long due and good to see @maramihali looking into improving our test stdlib test suite.

Totally agree, also glad @maramihali has improved that situation.

codygunton:
^rebase

PR #366, opened 2023-04-19
Ultraplonk check_circuit

Rumata888:
# Implements the ultra_circuit_constructor check_circuit method

The check_circuit method allows to check ultra circuits without using the proving system. The method tests:
1) Gate identities
2) Lookups
3) Permutations (sets, too!!)
The check involves computation of all finalized gates, but they are discarded right after the check. This should make iteratively debugging nasty ultra circuits much easier.

One of the major changes is the addition of the "in-the-head" circuit structure. It is being used in all methods that are called from check_circuit. The information about gate and witness changes is put there instead of the body of circuit_constructor, which allows us to discard this updated state. For this we use switched_* mechanic, where the location of actual selector and witness data is decided on the go.

One of the changes is that plookup tables have been moved to proof_system. The logic is that linking plonk to proof system caused gcc link issues and it's nature for the tables to be in the same place as the circuits.

# Checklist:


cpp/src/barretenberg/plonk/composer/ultra_composer.cpp
```diff
@@ -1845,6 +1845,7 @@ std::array<uint32_t, 2> UltraComposer::decompose_non_native_field_double_width_l
     const uint256_t value = get_variable(limb_idx);
     const uint256_t low = value & LIMB_MASK;
     const uint256_t hi = value >> DEFAULT_NON_NATIVE_FIELD_LIMB_BITS;
+    // WTF(kesha): What is this supposed to do? Unless uint256_t has failed, this should always work
```
codygunton, line 1848: Ha, harder to find this if it doesn't contain the word TODO somewhere...

cpp/src/barretenberg/plonk/composer/splitting_tmp/composer_helper/ultra_plonk_composer_helper.cpp
```diff
@@ -50,7 +50,7 @@ void UltraPlonkComposerHelper<CircuitConstructor>::compute_witness(CircuitConstr
 
     // TODO(luke): subgroup size was already computed above but compute_witness_base computes it again. If we pass in
     // NUM_RANDOMIZED_GATES (as in the other split composers) the resulting sizes can differ. Reconcile this.
-    auto wire_polynomial_evaluations = compute_witness_base(circuit_constructor, total_num_gates, NUM_RANDOMIZED_GATES);
+    auto wire_polynomial_evaluations = compute_witness_base(circuit_constructor, total_num_gates, NUM_RESERVED_GATES);
```
codygunton, line 53: Corresponding change in the definition of `compute_witness_base`? If it's really should be called `NUM_RESERVED_GATES` then we should change the name of the function argument, no? Your goal is just uniformity of naming, right, there's no difference?

codygunton: But below we have `num_randomized_gates = NUM_RESERVED_GATES`. Are we just using the fact that presently we have `NUM_RESERVED_GATES==NUM_RANDOMIZED_GATES`? Both of these could be specified statically in the flavor, so it's probably best to keep the roles distinct.

cpp/src/barretenberg/proof_system/circuit_constructors/ultra_circuit_constructor.hpp
```diff
@@ -6,208 +6,446 @@
...
+                            return true;
+                        }
+                    }
                 }
             }
+            return false;
+        }
+    };
+    /**
+     * @brief Circuit-in-the-head is a structure we use to store all the information about the circuit which should be
```
codygunton: I find this name pretty unintuitive and I suspect others will as well. You said you chose this name as an alternative to using the word "virtual" since C++ has a the keyword `virtual`. What do you think about something like `TemporaryConstructor` or `ScratchConstructor`? 

Or, is it not possible to reduce duplication by doign something like this?
 - Derive `UltraCircuitConstructorData : public CircuitConstructorBase<arithmetization::Ultra>`
 - Derive `UltraCircuitConstructor : public UltraCircuitConstructorData`
 - Make a member `UltraCircuitConstructorData scratch_data` in `UltraCircuitConstructor`

 Do you not like this suggestion? Feels cleaner to me, what do you think?

cpp/src/barretenberg/proof_system/circuit_constructors/ultra_circuit_constructor.hpp
```diff
@@ -6,208 +6,446 @@
...
+     * used during check_circuit method call, but needs to be discarded later
+     *
+     * @details In check_circuit method in UltraCircuitConstructor we want to check that the whole circuit works, but
```
codygunton, line 235: Would you add another sentence like explaining further how we accomplish the goal? Maybe something like: 'This class lets us pause circuit construction and temporarily generate only the finalization gates needed for a partial circuit check. When the check is complete, we discard this data before continuing with main thread of circuit construction.'

cpp/src/barretenberg/proof_system/circuit_constructors/ultra_circuit_constructor.hpp
```diff
@@ -6,208 +6,446 @@
...
+            stored_state.memory_write_records = circuit_constructor.memory_write_records;
+            stored_state.range_lists = circuit_constructor.range_lists;
+            stored_state.circuit_finalised = circuit_constructor.circuit_finalised;
+            stored_state.num_gates = circuit_constructor.num_gates;
+            stored_state.cached_non_native_field_multiplications =
+                circuit_constructor.cached_non_native_field_multiplications;
+            return stored_state;
+        }
+        /**
+         * @brief CHecks that the circuit states is the same as the stored circuit's one
```
codygunton: Typos

cpp/src/barretenberg/proof_system/circuit_constructors/ultra_circuit_constructor.hpp
```diff
@@ -6,208 +6,446 @@
...
+         *
+         * @param circuit_constructor
+         * @return true
+         * @return false
+         */
+        template <typename CircuitConstructor> bool is_same_state(const CircuitConstructor& circuit_constructor)
```
codygunton, line 426: Why is this a template?

cpp/src/barretenberg/proof_system/circuit_constructors/ultra_circuit_constructor.hpp
```diff
@@ -609,5 +939,136 @@ class UltraCircuitConstructor : public CircuitConstructorBase<arithmetization::U
...
+    template <typename T> inline T& choose_virtual_or_real(T& virtual_member, T& real_member, bool in_the_head)
+    {
+        if (in_the_head) {
+            return virtual_member;
+        } else {
+            return real_member;
+        }
+    }
+
+#define PARENS ()
```
codygunton: Oh god, why is this all done with macros?

codygunton: This is pretty hard to read. I'm spending a while trying to figure out how and where we specify that we're constructing 'in the head' or not.

cpp/src/barretenberg/proof_system/circuit_constructors/ultra_circuit_constructor.cpp
```diff
@@ -2368,4 +2443,943 @@ void UltraCircuitConstructor::process_RAM_arrays(const size_t gate_offset_from_p
...
+}
+
+/**
+ * @brief Reset the circuit-in-the-head construction that we use for checking the correctness of the circuit to the
+ * values in the circuit
+ *
+ */
+void UltraCircuitConstructor::update_circuit_in_the_head()
+{
+    // Unfortunately we have to copy all variable structures
```
codygunton: Why?

cpp/src/barretenberg/proof_system/circuit_constructors/ultra_circuit_constructor.cpp
```diff
@@ -2368,4 +2443,943 @@ void UltraCircuitConstructor::process_RAM_arrays(const size_t gate_offset_from_p
...
+ *
+ * @return true
+ * @return false
+ */
+bool UltraCircuitConstructor::check_circuit()
+{
+    bool result = true;
+    // Put the circuit in the head
+    update_circuit_in_the_head();
+    in_the_head = true;
```
codygunton: Do I have it right that this is the only place where we switch to in-the-head mode?

cpp/src/barretenberg/proof_system/circuit_constructors/ultra_circuit_constructor.hpp
```diff
@@ -609,5 +939,136 @@ class UltraCircuitConstructor : public CircuitConstructorBase<arithmetization::U
...
+#define HEAD(x, ...) x
+#define TAIL(x, ...) __VA_ARGS__
+#define IN_THE_HEAD_MEMBER_NAME(member) circuit_in_the_head.member
+#define ASSIGN_VARIABLE_TO_VIRTUAL_OR_REAL(variable_prefix, switch_name, member)                                       \
+    auto& variable_prefix##member = choose_virtual_or_real(IN_THE_HEAD_MEMBER_NAME(member), member, switch_name);      \
+    (void)variable_prefix##member;
+#define CHOOSE_VIRTUAL_OR_REAL_MULTIPLE(variable_prefix, switch_name, member_name, ...)                                \
+    FOR_EACH(ASSIGN_VARIABLE_TO_VIRTUAL_OR_REAL, variable_prefix, switch_name, member_name, __VA_ARGS__)
+#define ENABLE_ALL_IN_THE_HEAD_SWITCHES                                                                                \
+    CHOOSE_VIRTUAL_OR_REAL_MULTIPLE(switched_,                                                                         \
```
codygunton: Like what does this do? Why is this called `MULTIPLE`?

cpp/src/barretenberg/proof_system/circuit_constructors/ultra_circuit_constructor.cpp
```diff
@@ -2368,4 +2443,943 @@ void UltraCircuitConstructor::process_RAM_arrays(const size_t gate_offset_from_p
     }
 }
 
+/**
+ * @brief Various methods relating to circuit evaluation
```
codygunton: I think this will eventually confuse doxygen, can you write the comment withou tags? This is just like a big divider, right?

cpp/src/barretenberg/proof_system/circuit_constructors/ultra_circuit_constructor.cpp
codygunton: I guess this is for simplicity because you just want one global boolean switch?

cpp/src/barretenberg/proof_system/circuit_constructors/ultra_circuit_constructor.cpp
```diff
@@ -2368,4 +2443,943 @@ void UltraCircuitConstructor::process_RAM_arrays(const size_t gate_offset_from_p
...
+ *
+ */
+
+/**
+ * @brief Arithmetic gate-related methods
```
codygunton, line 2420: I think we need to have some tests that connect the check circuit functions that verify they are correct. Am I missing that this is already implicit somwhere? I think presently we are relying on someone thorougly auditing the code below and making sure to keep it updated with any changes that may happen (hopefully don't happen, but may) with how circuits are Ultra constructed. Do you have any thoughts about how/when to do that?

codygunton (changes requested):
This is really impressive work and a huge win for us. I have some concerns about the complexity of its implementation though. Our code is already pretty complex (for newcomers, but even for people on adjacent teams), and I fear that this sets a concerning precedent (embeding a sort of sub-constructor in each constructor resulting in duplication; using hard-to-read macros), so I'm asking you to spend a bit of time thinking about potential simplifications and/or ways to reuse code, maybe avoid calling everything `switched` (also not a very intuitive term IMO), and possibly renaming some some of what you have done here becuase I really think this `in_the_head` motif is not intuitive and will trip many people up. Let's discuss more tomorrow and find a way to get this in soon.

cpp/src/barretenberg/proof_system/circuit_constructors/ultra_circuit_constructor.cpp
codygunton: Namely, for each type of gate, there should be a unit test of that gate creation function that runs `check_circuit` and checks that returns `true` if and only if the circuit assignment is valid. The hope is that this is sufficient, though I feel we need an argument that composition is then valid, and we probably need to mix in funny things like genperm range constraints and `assert_equal`s. And of course we should still test some higher-level circuits thoroughly in the full construction-verification flow.

codygunton (approved):

codygunton:
This is awesome, thanks.

PR #368, opened 2023-04-20
[Temporary] Pending bb work for aztec3

suyash67:
# Description

The aztec3 circuits were supposed to use `barretenberg/master` as a submodule, but there are issues in using the current master. The main issue is the number of generators for Fixed-base Pedersen was reduced by the Noir team since larger number of generators broke their integration tests with Ultraplonk (the PR is [here](https://github.com/AztecProtocol/barretenberg/pull/338)). But aztec3 needs lot more generators so I decided to create a temporary branch `aztec3-temporary` off master to support larger number of generators.

Therefore, until we figure out the root cause of why the noir tests were failing with more number of generators, aztec3 will keep using this temporary branch as a submodule. One consequence of this is that subsequent bb work (needed for aztec3) will be merged in this branch. We will try to keep the PRs very short and regularly rebase this branch to avoid major conflicts.

# Checklist:


ludamad:
Thanks for this, was also hitting the generator issue trying to work off of barretenberg master. I feel like we should ASAP make a CI test (just treated as additional info) that runs a bunch of aztec3 checks off of barretenberg commits

dbanks12 (commented):
LGTM but I'd change `GEN_PARAMS` for `__wasm__` to `{64, 32, 16, 64}` as that is what it was pre-`GEN_PARAMS`.

Edit: Nevermind. @ludamad made sure the numbers use in this PR match master and therefore match noir needs.

dbanks12:
Should get @suyash67 to approve this before we merge I think

cpp/src/barretenberg/stdlib/primitives/bigfield/bigfield.hpp
```diff
@@ -160,7 +160,7 @@ template <typename Composer, typename T> class bigfield {
         field_t<Composer> lo = binary_basis_limbs[0].element + (binary_basis_limbs[1].element * shift_1);
         field_t<Composer> hi = binary_basis_limbs[2].element + (binary_basis_limbs[3].element * shift_1);
         // n.b. this only works if NUM_LIMB_BITS * 2 is divisible by 8
-        ASSERT((NUM_LIMB_BITS / 8) * 8 == NUM_LIMB_BITS);
+        ASSERT((NUM_LIMB_BITS * 2 / 8) * 8 == NUM_LIMB_BITS * 2);
```
codygunton, line 169: I see the comment above. Weird that this wasn't violated before. Was this assert triggering? Probably we should leave a comment here explaining the before-and-after so that someone can double check during a future audit when this logic is valid.

cpp/src/barretenberg/crypto/generators/generator_data.cpp
```diff
@@ -4,18 +4,30 @@ namespace crypto {
 namespace generators {
 namespace {
 
-// The number of unique base points with default main index with precomputed ladders
+// Parameters for generator table construction
```
codygunton: Already reviewed in #409

cpp/src/barretenberg/ecc/groups/affine_element_impl.hpp
```diff
@@ -229,5 +229,38 @@ affine_element<Fq, Fr, T> affine_element<Fq, Fr, T>::hash_to_curve(const uint64_
...
+    Fq x;
+    Fq y;
+    while (!found_one) {
+        // Sample a random x-coordinate and check if it satisfies curve equation.
+        x = Fq::random_element(engine);
+        yy = x.sqr() * x + T::b;
+        if constexpr (T::has_a) {
+            yy += (x * T::a);
+        }
+        auto [found_root, y1] = yy.sqrt();
```
codygunton, line 251: When `!found_root` it makes sense to skip the work below here (sampling a random byte and doing a field negation), but doesn't matter unless we plan to generate a lot of points with this function.

cpp/src/barretenberg/stdlib/encryption/ecdsa/ecdsa_impl.hpp
```diff
@@ -13,6 +13,11 @@ bool_t<Composer> verify_signature(const stdlib::byte_array<Composer>& message,
 {
     Composer* ctx = message.get_context() ? message.get_context() : public_key.x.context;
 
+    // Check if revovery id v is either 27 ot 28.
```
codygunton: Typo

cpp/src/barretenberg/stdlib/encryption/ecdsa/ecdsa_impl.hpp
```diff
@@ -13,6 +13,11 @@ bool_t<Composer> verify_signature(const stdlib::byte_array<Composer>& message,
...
+    // TODO(Suyash): check with Zac/Kesha/Cody.
+    field_t<Composer>(sig.v).assert_is_in_set({ field_t<Composer>(27), field_t<Composer>(28) },
```
codygunton, line 43: What is the meaning of this condition? Is it documented somewhere?

codygunton (approved):
Looks good, I just have some small suggestions and requests for clarification. NB: I only skimmed the c_bind stuff, but I assume there are some tests somewhere that verify it works.

cpp/src/barretenberg/stdlib/primitives/bigfield/bigfield.hpp
suyash67: Yes, this was triggering because the method `to_byte_array()` was being used to convert an ECDSA public key to `byte_array`. I think we never really used the conversion of a bigfield scalar to `byte_array` (we still don't actually, the ECDSA conversion I mentioned is no longer required and has been removed). But I still kept this fix.

Apparently, Guido had bumped into this a while back in his fuzzing: https://github.com/AztecProtocol/aztec2-internal/issues/1496

I have also added a comment explaining the need for this assertion. I've removed your TODO.

cpp/src/barretenberg/ecc/groups/affine_element_impl.hpp
suyash67: This was Kesha's suggestion and I agree with him. If we don't sample a bit and negate the y-coordinate, we're sampling a random point from only "upper" half of the curve points. To really get a "random" point on the curve, we need to randomly change the sign of the `y` coordinate.

cpp/src/barretenberg/stdlib/encryption/ecdsa/ecdsa_impl.hpp
suyash67: Thanks for fixing.

cpp/src/barretenberg/stdlib/encryption/ecdsa/ecdsa_impl.hpp
suyash67: I have added detailed comment explaining this (also added references). Resolving this.

suyash67:
> Looks good, I just have some small suggestions and requests for clarification. NB: I only skimmed the c_bind stuff, but I assume there are some tests somewhere that verify it works.

Thanks for reviewing, I'll merge it now. About the c_binds, yes they are being tested in aztec3-packages' `barretenberg.js`. Ideally at some point, we should add local tests for all c_binds in barretenberg (@dbanks12 did this for the aztec3 circuit c_binds). I'll open an issue for that but we can do it later.

cpp/src/barretenberg/ecc/groups/affine_element_impl.hpp
codygunton: Yes, this make sense, but my point is that if `found_root` is `false`, then you don't need to bother sampling a random byte and negating a coordinate, so it would be more efficient to to wrap everything below line 251 in an `if (found_root)` block. About half the time the `while` loop will execute more than once, meaning that if you're generating 2 billion random points of the curve, then you will sample about 1 billion unneeded bytes and do about 1 billion unneeded 256-bit field element negations.

PR #377, opened 2023-04-23
Zw/recursion constraint reduction

zac-williamson:
# Description

Reduced the cost of recursively verifying an UltraPlonk proof to 247,119 constraints. Goal is to be able to recursively verify 2 proofs in a circuit with size 2^19. 2 key changes:

1. Reduced duplicate creation of non-native field multiplication constraints

UltraComposer now caches non-native scalar multiplications in `evaluate_non_native_field_multiplication` . When finalising the circuit, duplicate constraints (i.e. same witnesses) are removed. This occurs when a duplicate non-native field multiplication is created over the same witnesses. Savedsapprox 35,000 constraints 

2. Removed blake3s hash

The UltraPlonk recursive prover/verifier now only uses a Pedersen hash when applying the Fiat-Shamir heuristic. Note that we will need to change this longer term! The intention is to use a Poseidon hash once we have implemented a Poseidon gadget. Saves approx. 35,000 constraints
 
# Checklist:


cpp/src/barretenberg/plonk/composer/ultra_composer.cpp
```diff
@@ -1939,82 +1938,127 @@ std::array<uint32_t, 2> UltraComposer::evaluate_non_native_field_multiplication(
...
-        input.q[0],
-        input.q[1],
-        hi_2_idx,
-        -1,
-        input.neg_modulus[3] * LIMB_RSHIFT + input.neg_modulus[2] * LIMB_RSHIFT_2,
-        input.neg_modulus[2] * LIMB_RSHIFT + input.neg_modulus[1] * LIMB_RSHIFT_2,
-        LIMB_RSHIFT_2,
-        0,
-    });
+    auto last =
```
codygunton, line 1969: Eliminate duplicates by shifting unique entries back to toward the beginning of the vector, then returns a pointer to the first element of the vector past the last unique element :heavy_check_mark:.

cpp/src/barretenberg/plonk/composer/ultra_composer.cpp
```diff
@@ -1939,82 +1938,127 @@ std::array<uint32_t, 2> UltraComposer::evaluate_non_native_field_multiplication(
...
-            hi_1_idx,
-            -input.neg_modulus[1] * LIMB_SHIFT - input.neg_modulus[0],
-            -input.neg_modulus[0] * LIMB_SHIFT,
-            -1,
-            -1,
-            0,
-        },
-        true);
+/**
+ * @brief Called in `compute_proving_key` when finalizing circuit.
```
codygunton, line 1961: TODO: name and description of `evaluate_non_native_field_multiplication` are no longer correct.

cpp/src/barretenberg/plonk/composer/ultra_composer.cpp
```diff
@@ -1939,82 +1938,127 @@ std::array<uint32_t, 2> UltraComposer::evaluate_non_native_field_multiplication(
...
-                          lo_1_idx,
-                          input.neg_modulus[0] + input.neg_modulus[1] * LIMB_SHIFT,
-                          input.neg_modulus[0] * LIMB_SHIFT,
-                          -LIMB_SHIFT,
-                          -LIMB_SHIFT.sqr(),
-                          0 },
-                        true);
+    // Add witnesses into the multiplication cache
+    // (when finalising the circuit, we will remove duplicates; several dups produced by biggroup.hpp methods)
+    cached_non_native_field_multiplication cache{
```
codygunton: TODO: probably better to name this something like `cache_entry`.

cpp/src/barretenberg/plonk/composer/ultra_composer.cpp
```diff
@@ -1939,82 +1938,127 @@ std::array<uint32_t, 2> UltraComposer::evaluate_non_native_field_multiplication(
...
+        .a = input.a,
+        .b = input.b,
+        .q = input.q,
+        .r = input.r,
+        .cross_terms = { lo_0_idx, lo_1_idx, hi_0_idx, hi_1_idx, hi_2_idx, hi_3_idx },
```
codygunton, line 1952: TODO: clearer to make this a struct.

codygunton: Sanity check: these are witness indices, so the multiplications being removed really are duplicates. :heavy_check_mark:.

cpp/src/barretenberg/plonk/composer/ultra_composer.hpp
```diff
@@ -39,6 +40,57 @@ class UltraComposer : public ComposerBase {
...
+            bool valid = true;
+            for (size_t i = 0; i < 5; ++i) {
+                valid = valid && (a[i] == other.a[i]);
+                valid = valid && (b[i] == other.b[i]);
+                valid = valid && (q[i] == other.q[i]);
+                valid = valid && (r[i] == other.r[i]);
+            }
+            return valid;
+        }
+        bool operator<(const cached_non_native_field_multiplication& other) const
```
codygunton, line 84: Each `return` is a way in which we can see that `this` and `other` are different. `cross_terms` entry is determined by a static formula in `a, b, q, r`, hence there is no need to compare these. In case the formula were computed in a dynamic way, we'd have an issue here, but it's not worth worrying about that now.

cpp/src/barretenberg/stdlib/recursion/transcript/transcript.hpp
```diff
@@ -253,8 +253,23 @@ template <typename Composer> class Transcript {
...
 
-        byte_array<Composer> base_hash = stdlib::blake3s(compressed_buffer);
-
+        // TODO(@zac-williamson) make this a Poseidon hash
+        byte_array<Composer> base_hash;
+        if constexpr (Composer::type == ComposerType::PLOOKUP) {
+            std::vector<field_pt> compression_buffer;
+            field_pt working_element(context);
+            size_t byte_counter = 0;
+            split(working_element, compression_buffer, field_pt(compressed_buffer), byte_counter, 32);
```
codygunton, line 262: TODO: comment describing `split` is incorrect. The function splits `element` and inserts it into `element_buffer`, using `working_element` as an auxiliary field element to accomplish this.

cpp/src/barretenberg/stdlib/recursion/transcript/transcript.hpp
```diff
@@ -267,9 +282,25 @@ template <typename Composer> class Transcript {
```
codygunton, line 282: This block of code only executes for `num_challenges > 2`, which (currently) only happens in the `nu` round when we need to generate short scalars. In this case, we generate 32-byte challenges and split them in half to get the relevant challenges.

codygunton (commented):
Looked it over and made some comments. Going to address TODOs then bring the same logic in to Honk.

PR #381, opened 2023-04-25
chore: Fetch master for DSL recursion

vezenovm:
# Description

The noir-recursion branch requires more changes from master that Noir also requires

There was a change in the recursion work that removed the reference string from the std lib verification key which required usage of the env crs. However, any usage of `env_load_prover_crs` or `env_load_verifier_crs` fails on mainframe due to `undefined reference to `bbmalloc'` (oddly enough it all works on my local macbook). This branch also includes a workaround to get EnvReferenceString compiling by just calling aligned_alloc which bbmalloc calls internally. However, I don't know if this possibly breaks things elsewhere in wasm. Otherwise we may have to go back to storing the reference string inside the stdlib verification key, or nail down why we are getting an undefined reference when linking on linux.

# Checklist:

- [ ] I have reviewed my diff in github, line by line.
- [ ] Every change is related to the PR description.
- [ ] I have [linked](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) this pull request to the issue(s) that it resolves.
- [ ] There are no unexpected formatting changes, superfluous debug logs, or commented-out code.
- [ ] There are no circuit changes, OR specifications in `/markdown/specs` have been updated.
- [ ] There are no circuit changes, OR a cryptographer has been assigned for review.
- [ ] I've updated any terraform that needs updating (e.g. environment variables) for deployment.
- [ ] The branch has been rebased against the head of its merge target.
- [ ] I'm happy for the PR to be merged at the reviewer's next convenience.
- [ ] New functions, classes, etc. have been documented according to the doxygen comment format. Classes and structs must have `@brief` describing the intended functionality.
- [ ] If existing code has been modified, such documentation has been added or updated.

PR #383, opened 2023-04-25
feat: DSL recursion changes

vezenovm:
# Description

This PR enables the starting skeleton for recursion in Noir. 

There were heavy merge conflicts between `zw/noir-recursion` and master. I originally made this branch `mv/recursion-master-merge` in this PR: https://github.com/AztecProtocol/barretenberg/pull/381, but it included all the changes from master in the diff. This branches checks out `zw/noir-recursion` and brought it up-to-date with master. The diff should now represent only the recursion changes. 

There was a change in the recursion work that removed the reference string from the std lib verification key which required usage of the env crs. All the stdlib recursion tests now use the env crs. However, in the solidity tests using the env crs resulted in an error that we were trying to invert zero in the field. As a quick workaround I added back the reference string field to the stdlib verification key. 

# Checklist:


cpp/src/barretenberg/env/CMakeLists.txt
```diff
@@ -1 +1 @@
-barretenberg_module(env)
```
codygunton: Err, this seems kinda bad: we have env depends on ecc depends on numeric depends on env. Not the CMake wizard, but this is not great, right? Where did this come from?

cpp/src/barretenberg/plonk/proof_system/verification_key/verification_key.cpp
```diff
@@ -178,4 +178,102 @@ sha256::hash verification_key::sha256_hash()
     return sha256::sha256(to_buffer(vk_data));
 }
 
+/**
+ * @brief When recursively verifying proofs, we represent the verification key using field elements.
+ *        This method exports the key formatted in the manner our recursive verifier expects.
+ *        NOTE: only used by the dsl at the moment. Might be cleaner to make this a dsl function?
```
codygunton: If "recursion format" refers to a DSL-specific format, then I'd rather make this function in the dsl module. Otherwise it seems fine. But that change can wait if it's inconvenient to handle it now.

cpp/src/barretenberg/stdlib/encryption/ecdsa/ecdsa.hpp
```diff
@@ -18,6 +18,11 @@ bool_t<Composer> verify_signature(const stdlib::byte_array<Composer>& message,
                                   const G1& public_key,
                                   const signature<Composer>& sig);
 
+template <typename Composer, typename Curve, typename Fq, typename Fr, typename G1>
```
codygunton: Right, as discussed a while back, this is an ECDSA circuit that lets a prover demonstrate that a signature does not verify.

cpp/src/barretenberg/stdlib/recursion/transcript/transcript.hpp
```diff
@@ -48,6 +48,54 @@ template <typename Composer> class Transcript {
...
+        , transcript_base(input_manifest, transcript::HashType::PedersenBlake3s, 16)
+        , current_challenge(in_context)
+    {
+        size_t count = 0;
+
+        const auto num_rounds = input_manifest.get_num_rounds();
+        for (size_t i = 0; i < num_rounds; ++i) {
+            for (auto manifest_element : input_manifest.get_round_manifest(i).elements) {
+                if (!manifest_element.derived_by_verifier) {
+                    if (manifest_element.num_bytes == 32 && manifest_element.name != "public_inputs") {
```
codygunton, line 76: Right the cases are, in order: polynomial evaluation; commitment; public inputs. FYI this will become more explicit in Honk.

cpp/src/barretenberg/stdlib/recursion/verification_key/verification_key.hpp
```diff
@@ -105,22 +119,70 @@ template <typename Composer> struct evaluation_domain {
  */
```
codygunton: The comment above this function is for `from_witness`, so it should move down.

cpp/src/barretenberg/stdlib/recursion/verification_key/verification_key.hpp
```diff
@@ -105,22 +119,70 @@ template <typename Composer> struct evaluation_domain {
...
 template <typename Curve> struct verification_key {
     using Composer = typename Curve::Composer;
+
+    template <bool inner_proof_contains_recursive_proof = false>
+    static std::shared_ptr<verification_key> from_field_pt_vector(
```
codygunton: Naming this `from_field_pt_vector` is a little funny because it refers to an alias `field_pt` that is not universal (in fact we use `field_ct` more often). Morever, it would improve readability a lot IMO to name this something like `from_field_elements` and have that create a struct like `verification_key_initializer_t<Composer>` to avoid the magic numbers below (would help readability and make us less dependent on order). Buuut that's a bigger refactor and it means doing more work below. Maybe for this and related issues could you make a github issue?

cpp/src/barretenberg/stdlib/recursion/verification_key/verification_key.hpp
```diff
@@ -25,6 +25,20 @@ namespace stdlib {
 namespace recursion {
 
 template <typename Composer> struct evaluation_domain {
+    static evaluation_domain from_field_pt_vector(const std::vector<field_t<Composer>>& fields)
+    {
+        evaluation_domain domain;
+        domain.root = fields[0];
```
codygunton, line 184: This line is effectively _defining_ the structure of the initializer vector, right?

cpp/src/barretenberg/stdlib/recursion/verification_key/verification_key.hpp
```diff
@@ -105,22 +119,70 @@ template <typename Composer> struct evaluation_domain {
...
+            const uint32_t idx = static_cast<uint32_t>(uint256_t(fields[6 + i].get_value()));
+            key->recursive_proof_public_input_indices.emplace_back(idx);
+        }
+        // Apply constraints to force the recursive proof information to be circuit constants
+        fields[5].assert_equal(inner_proof_contains_recursive_proof);
+        for (size_t i = 0; i < 16; ++i) {
+            fields[6 + i].assert_equal(recursive_proof_public_input_indices[i]);
+        }
+
+        size_t count = 22;
```
codygunton, line 257: 6 + 15 == 22 :heavy_check_mark:

cpp/src/barretenberg/stdlib/recursion/verification_key/verification_key.hpp
```diff
@@ -264,23 +325,17 @@ template <typename Curve> struct verification_key {
...
     std::shared_ptr<VerifierReferenceString> reference_string;
-
     PolynomialManifest polynomial_manifest;
-
+    // Used to check in the circuit if a proof contains any aggregated state.
+    bool contains_recursive_proof = false;
+    std::vector<uint32_t> recursive_proof_public_input_indices;
     size_t program_width = 4;
-
-    std::shared_ptr<plonk::verification_key> base_key;
```
codygunton, line 399: I suppose the reasoning for getting rid of this was: dsl already produced circuit types, so the options were to either: instantiate a vk from this and use existing code paths OR refactor the recursive verifier a bit. You noted that the base key could pretty easily go away, so you opted for the latter. Is that right?

cpp/src/barretenberg/stdlib/recursion/verifier/verifier.hpp
```diff
@@ -325,18 +343,14 @@ aggregation_state<Curve> verify_proof(typename Curve::Composer* context,
         rhs_scalars.push_back(random_separator);
     }
 
-    // Check if recursive proof information is correctly set.
-    key->contains_recursive_proof.assert_equal(key->base_key->contains_recursive_proof,
```
codygunton, line 315: ⚠️ it looks like this logic was moved into `from_field_pt_vector`, but if we use the `from_witness` of a verification constructor of a verification key we will be missing this constraint!

codygunton (approved):
OK, I've reviewed everything not in the directories dsl or solidity_helpers. Most of what I have to say are little comments for clarifiation / disucssion and thoughts for improvements that could be pursued later. 

I think I noticed one constraint that goes away, and it would be good to fix this by adding the appropriate constraint in the `from_witness` verification key constructor. It's debatable whether that's actually a good idea -- I think it's probably ideal to keep as much of the recursive verifier logic as possible in one place (for security analysis we don't want branching in the space of circuits we create--hard to keep track of all of the possibilities). But, on that note, 

> **UltraPlonK must not be considered secure we audit it** 

and our goal is to get something performant into developers' hands now, so it's certainly reasonable to move things around as you have and address the missing constraint in some way.

cpp/src/barretenberg/env/CMakeLists.txt
vezenovm: Without including the ecc module I would get undefined reference errors for `bbmalloc` inside the `env_load_prover_crs` and `env_load_verifier_crs` methods. I was originally able to work around this by using `aligned_alloc` inside those methods (`aligned_alloc` is just the internal method to `bbmalloc`). I could try going back to that.

The usage of the env crs came about because @zac-williamson realized it could be used instead of having to store the reference string from the native verification key in the stdlib verification key. This was before the solidity tests were added though. I reverted back to storing the reference string as I kept getting errors when trying to use the env crs with the solidity helpers. So theoretically I could revert this module change and remove all usage of the env crs in the stdlib recursive verifier tests, but it still leaves this issue with the env crs (I will make an issue on the repo if we go with this strategy).

Perhaps there is someone more familiar with cmake I can request takes a look at this?

cpp/src/barretenberg/plonk/proof_system/verification_key/verification_key.cpp
vezenovm: I will look at moving this into the DSL module

codygunton:
@suyash67 I recommend having a look here to see if any of the recursive verifier changes will affect you. But note that we will get some feedback from aztec tests once @vezenovm rebases :innocent:

cpp/src/barretenberg/stdlib/encryption/ecdsa/ecdsa.hpp
vezenovm: ECDSA was fixed in this PR: https://github.com/AztecProtocol/barretenberg/pull/435 and a couple subsequent ones that were fixed. So any ECDSA changes should no longer be a part of this PR

cpp/src/barretenberg/plonk/proof_system/verification_key/verification_key.cpp
vezenovm: Moved

cpp/src/barretenberg/stdlib/recursion/verification_key/verification_key.hpp
vezenovm: I renamed the method as I agree it is more of a universal name. I will open a separate issue to move to an initializer struct

cpp/src/barretenberg/stdlib/recursion/verification_key/verification_key.hpp
vezenovm: moved

cpp/src/barretenberg/stdlib/recursion/verification_key/verification_key.hpp
```diff
@@ -98,29 +112,72 @@ template <typename Composer> struct evaluation_domain {
...
         key->reference_string = input_key->reference_string;
         key->polynomial_manifest = input_key->polynomial_manifest;
 
         // Circuit types:
         key->n = witness_t<Composer>(ctx, barretenberg::fr(input_key->circuit_size));
         key->num_public_inputs = witness_t<Composer>(ctx, input_key->num_public_inputs);
         key->domain = evaluation_domain<Composer>::from_witness(ctx, input_key->domain);
-        key->contains_recursive_proof = witness_t<Composer>(ctx, input_key->contains_recursive_proof);
-
+        key->contains_recursive_proof = input_key->contains_recursive_proof;
```
vezenovm, line 294: @zac-williamson I imagine we did this switch as we need the recursive proof indices to be circuit constants either way so we should not be fetching it from a witness?

cpp/src/barretenberg/stdlib/recursion/verifier/verifier.hpp
vezenovm: @zac-williamson This is the constraint that was removed

zac-williamson: In the new `from_field_pt_vector`, `key->contains_recursive_proof` is a circuit constant, it's not a witness (it's value is constructed out of a regular `bool` type)

The constraint is removed because there is no longer a `contains_recursive_proof` witness value at all - we effectively generate 2 different circuits depending on whether the input into `from_field_pt_vector` describes a recursive circuit or not.

cpp/src/barretenberg/dsl/acir_format/acir_format.cpp
```diff
@@ -182,6 +198,21 @@ Composer create_circuit(const acir_format& constraint_system,
...
+    // Add recursion constraints
+    for (size_t i = 0; i < constraint_system.recursion_constraints.size(); ++i) {
+        auto& constraint = constraint_system.recursion_constraints[i];
+        create_recursion_constraints<false>(composer, constraint);
+
+        // make sure the verification key records the public input indices of the final recursion output
+        // (N.B. up to the ACIR description to make sure that the final output aggregation object wires are public
+        // inputs!)
+        if (i == constraint_system.recursion_constraints.size() - 1) {
+            std::vector<uint32_t> proof_output_witness_indices(constraint.output_aggregation_object.begin(),
```
zac-williamson, line 210: I added a method into `aggregation_state` that unifies these two steps (populating witness indices and calling set_recursive_proof): `void aggregation_state::assign_object_to_proof_outputs`

Would you mind updating to use it? I think it would allows us to remove the `set_recursive_proof` method and remove the complexity of having to assign `proof_output_witness_indices`

cpp/src/barretenberg/dsl/acir_format/recursion_constraint.cpp
```diff
@@ -0,0 +1,356 @@
...
+            // (when verifying zk proofs this is acceptable as we make sure verification key points are not identical.
+            // And prover points should contain randomness for an honest Prover).
+            // This check can also trigger a runtime error due to causing 0 to be inverted.
+            // When creating dummy verification key points we must be mindful of the above and make sure that each
+            // transcript point is unique.
+            auto scalar = barretenberg::fr::random_element();
+            const auto element = barretenberg::g1::affine_element(barretenberg::g1::one * scalar);
+            const uint256_t x = element.x;
+            const uint256_t y = element.y;
+            const barretenberg::fr x_lo = x.slice(0, 136);
```
zac-williamson: Could we replace these magic numbers with derived values? These come from `proof_system::plonk::NUM_LIMB_BITS_IN_FIELD_SIMULATION` (68)

cpp/src/barretenberg/stdlib/recursion/verification_key/verification_key.hpp
vezenovm: Issue created here: https://github.com/AztecProtocol/barretenberg/issues/468

cpp/src/barretenberg/dsl/acir_format/acir_format.cpp
vezenovm: It does look like we may still have to assign proof_output_witness_indices as I have to construct an aggregation_state in order to use `assign_object_to_proof_outputs`. It also looks like constructing the aggregation state will require fetching all the values of the `output_aggregation_state` from the composer to construct P0 and P1. I can still of course push an update to show how it would look but it does seem that there is some extra complexity to using `assign_object_to_proof_outputs` even if we can get rid of `set_recursive_proof`

cpp/src/barretenberg/stdlib/recursion/verification_key/verification_key.hpp
zac-williamson: yes that's correct. Previously we did an `assert_equal` check in the circuit that effectively made this a circuit constant. This change removes that requirement to call `assert_equal`

cpp/src/barretenberg/stdlib/recursion/verification_key/verification_key.hpp
zac-williamson: Yes that's correct. We used the base key for very little so we figured it would be more efficient to remove entirely.

cpp/src/barretenberg/dsl/acir_format/acir_format.cpp
zac-williamson: Oh that is a good point. One alternative is to pass in a `bool add_to_public_inputs` parameter into `create_recursion_constraint`. If `add_to_public_inputs == true` then we call `assign_object_to_proof_outputs` on `aggregation_state_ct result` (line 116)

cpp/src/barretenberg/dsl/acir_format/recursion_constraint.cpp
vezenovm: Made a new function `export_g1_affine_element_as_fields` that uses constexprs to determine the limb slices

cpp/src/barretenberg/dsl/acir_format/acir_format.cpp
vezenovm: Unfortunately that alternative looks to also complicate how we currently construct our constraints. Calling `result.assign_object_proof_outputs` would mean we are now including the output aggregation object on the result aggregation state directly rather than it being a separate field on our RecursionConstraint. This change would require some larger changes in how we construct our recursion constraints and should be left as a separate optimization.

cpp/src/barretenberg/dsl/acir_format/recursion_constraint.cpp
```diff
@@ -0,0 +1,346 @@
...
+
+    // Assign the recursive proof outputs to `output_aggregation_object`
+    for (size_t i = 0; i < result.proof_witness_indices.size(); ++i) {
+        const auto lhs = field_ct::from_witness_index(&composer, result.proof_witness_indices[i]);
+        const auto rhs = field_ct::from_witness_index(&composer, input.output_aggregation_object[i]);
+        lhs.assert_equal(rhs);
+    }
+}
+
+template void create_recursion_constraints<false>(Composer&, const RecursionConstraint&);
```
kevaundray: This differs from how it is done for ecdsa_constraints -- we should add dummy_constraints in one way

cpp/src/barretenberg/dsl/acir_format/recursion_constraint.hpp
```diff
@@ -0,0 +1,108 @@
+#pragma once
+#include <vector>
+// #include "barretenberg/stdlib/types/types.hpp"
```
kevaundray: commented out code

cpp/src/barretenberg/dsl/acir_format/recursion_constraint.cpp
```diff
@@ -0,0 +1,346 @@
...
+#include "barretenberg/stdlib/recursion/aggregation_state/aggregation_state.hpp"
+#include "barretenberg/stdlib/recursion/verifier/verifier.hpp"
+#include "barretenberg/transcript/transcript_wrappers.hpp"
+
+namespace acir_format {
+
+using namespace proof_system::plonk;
+
+static constexpr uint64_t TWO_LIMBS_BITS_IN_FIELD_SIMULATION = NUM_LIMB_BITS_IN_FIELD_SIMULATION * 2;
+static constexpr uint64_t FOUR_LIMBS_BITS_IN_FIELD_SIMULATION = NUM_LIMB_BITS_IN_FIELD_SIMULATION * 4;
```
kevaundray, line 15: We should document what these constants are for

cpp/src/barretenberg/dsl/acir_proofs/acir_proofs.cpp
```diff
@@ -154,4 +189,194 @@ bool verify_proof(
...
+                              uint32_t proof_length,
+                              uint8_t const* vk_buf,
+                              uint32_t vk_length,
+                              uint32_t num_public_inputs,
+                              uint8_t const* input_aggregation_obj_buf,
+                              uint8_t** output_aggregation_obj_buf)
+{
+
+    bool inner_aggregation_all_zero = true;
+    std::vector<barretenberg::fr> aggregation_input(16);
```
kevaundray: We should document what this magic constant 16 is

cpp/src/barretenberg/dsl/acir_proofs/acir_proofs.cpp
```diff
@@ -154,4 +189,194 @@ bool verify_proof(
...
+    for (size_t i = 0; i < 16; i++) {
+        aggregation_input[i] = barretenberg::fr::serialize_from_buffer(&input_aggregation_obj_buf[i * 32]);
+        inner_aggregation_all_zero &= (aggregation_input[i].is_zero());
+    }
+
+    acir_format::aggregation_state_ct previous_aggregation;
+    if (!inner_aggregation_all_zero) {
+        std::array<acir_format::bn254::fq_ct, 4> aggregation_elements;
```
kevaundray: what does 4 represent here?

cpp/src/barretenberg/dsl/acir_format/recursion_constraint.cpp
vezenovm: I'll switch from template to a flag

cpp/src/barretenberg/plonk/composer/composer_base.hpp
```diff
@@ -186,6 +186,21 @@ class ComposerBase {
...
+        bool found = false;
+        uint32_t result = static_cast<uint32_t>(-1);
+        for (size_t i = 0; i < public_inputs.size(); ++i) {
+            if (real_variable_index[public_inputs[i]] == real_variable_index[witness_index]) {
+                found = true;
+                result = static_cast<uint32_t>(i);
+                break;
+            }
+        }
+        ASSERT(found == true);
```
kevaundray: small nit: you can get rid of `found` and just `ASSERT(result != -1)`

cpp/src/barretenberg/env/CMakeLists.txt
vezenovm: I have removed this so it is no longer an issue

cpp/src/barretenberg/dsl/acir_proofs/acir_proofs.cpp
vezenovm: I replaced its usage with `acir_format::RecursionConstraint::AGGREGATION_OUTPUT_STATE`

cpp/src/barretenberg/dsl/acir_proofs/acir_proofs.cpp
vezenovm: It represents the total number of field elements needed to represent an aggregation state. I'm going to see how I can switch this to use constexpr's and avoid the magic numbers

cpp/src/barretenberg/dsl/acir_format/recursion_constraint.hpp
vezenovm: Removed

cpp/src/barretenberg/dsl/acir_proofs/acir_proofs.cpp
vezenovm: I introduced a new `NUM_AGGREGATION_ELEMENTS` in `recursion_constraint.hpp`. Let me know what you think and if I missed anywhere that it can be used

cpp/src/barretenberg/plonk/composer/composer_base.hpp
vezenovm: fixed

cpp/src/barretenberg/dsl/acir_format/recursion_constraint.cpp
vezenovm: Documented let me know if the description makes sense

cpp/src/barretenberg/dsl/acir_format/recursion_constraint.cpp
vezenovm: Switched

cpp/src/barretenberg/dsl/acir_proofs/acir_proofs.cpp
```diff
@@ -154,4 +189,200 @@ bool verify_proof(
...
+    } else {
+        previous_aggregation.has_data = false;
+    }
+
+    acir_format::Composer composer;
+
+    std::vector<acir_format::field_ct> proof_fields(proof_length / 32);
+    std::vector<acir_format::field_ct> key_fields(vk_length / 32);
+    for (size_t i = 0; i < proof_length / 32; i++) {
+        // TODO(maxim): The stdlib pairing primitives fetch the context from the elements being used in the pairing
```
vezenovm, line 250: Issue for this TODO can be found here: https://github.com/AztecProtocol/barretenberg/issues/447

cpp/src/barretenberg/dsl/acir_format/acir_format.cpp
zac-williamson: sounds good. Might be a good idea to create a GH Issue to refactor the AggregationState interface to more easily handle this case + regular recursion w/o exposing internals like `proof_output_witness_indices`

cpp/src/barretenberg/dsl/acir_format/ecdsa_secp256k1.test.cpp
zac-williamson, line 1: How come TestECDSACompilesForVerifier has been removed?

cpp/src/barretenberg/dsl/acir_format/recursion_constraint.cpp
```diff
@@ -0,0 +1,347 @@
...
+        previous_aggregation.P0 = bn254::g1_ct(aggregation_elements[0], aggregation_elements[1]);
+        previous_aggregation.P1 = bn254::g1_ct(aggregation_elements[2], aggregation_elements[3]);
+        previous_aggregation.has_data = true;
+    } else {
+        previous_aggregation.has_data = false;
+    }
+
+    transcript::Manifest manifest = Composer::create_unrolled_manifest(input.public_inputs.size());
+
+    std::vector<field_ct> key_fields;
```
zac-williamson, line 106: magic number `1` here. Might be better to replace in this test w. hardcoded `const size_t num_public_inputs = 1`

cpp/src/barretenberg/dsl/acir_format/recursion_constraint.cpp
zac-williamson: lgtm 👍

cpp/src/barretenberg/stdlib/recursion/verification_key/verification_key.hpp
zac-williamson: yes. These values are Prover-dependent, but they are also inputs into the verification key hash; the circuit can validate they are correct against a known VK.

zac-williamson (approved):

Savio-Sou:
Superseded by https://github.com/AztecProtocol/barretenberg/pull/485.

PR #402, opened 2023-05-02
Pippenger over Grumpkin

codygunton:
# Description

This PR takes the existing scalar multiplication code, which was specific to the curve BN254, and generalizes it to work over the curve Grumpkin. Grumpkin also has a 254-bit scalar field--we do not generalize the Pippenger algorithm to work for SECP curves.

Major points:
 - To run the scalar multiplication test suite over Grumpkin, we need an CRS. I added grumpkin_srs_gen.cpp to accomplish this. For now, we generate transcript files on the fly ($2^{20}$) rather than to generate a full set of transcripts and pull them in from an S3 bucket.
 - The scalar multiplication code was in ecc/curves/bn254/scalar_multiplication. It has been moved to ecc/scalar_multiplication.
 - To reuse the scalar multiplication tests, I took the usual approach of turning `TEST`s into `TYPED_TEST`s. To accomplish this I added various structs containing only type aliases (e.g., we have a struct `curve::bn254::ScalarField` to provide `barretenberg::fr`).
 - To avoid lots of explicit instantiation and to allow for simpler type aliasing, the functions that were in  `barretenberg::io` are now static member of a class `srs::IO`. The same could/should be done for the functions in scalar_multiplication.hpp, but we use explicit instantiation for expediency.
 - An annoying wrinkle was the fact that some of the srs IO functions take an argument of type `barretenberg::g2` but there is no `grumpkin::g2`. Rather than to just make some dummy class and add garbage to the new SRS, I use template metaprogramming to generalize appropriately. In particular, I use the new C++20 SFINAE paradigm of a post-function `requires` clause to disable compiling certain class member functions when no `g2` type exists--the usual method of `std::enable_if_t` didn't compile, but using `requires(HasG2<Curve>)` works.

# Checklist:


codygunton:
^ rebase to 0e7d32f6460cc549d4e4b7e41a50ca6642a7bfa2

cpp/srs_db/grumpkin/monomial/README.md
```diff
@@ -0,0 +1,16 @@
+# Quick-and-dirty Grumpkin transcript
+
+The Grumpkin transcript currently departs in structure from the BN254
+transcript in that:
+ - It does not contain a checksum
+ - It does not contain any g2 points (indeed, there is no grumpkin::g2).
+ - The transcript generation binary only produces a single transcript file.
+   If more than 504000 points are desired at some point, it is likely we will 
+   need to a small refactor so that more files are created.
```
Rumata888, line 9: "to do"?

codygunton: I don't know if we ever actually need to address this, but yeah I'll make an issue nonetheless.

cpp/src/barretenberg/ecc/scalar_multiplication/pippenger.cpp
```diff
@@ -0,0 +1,51 @@
...
+    : num_points_(num_points)
+{
+    monomials_ = point_table_alloc<AffineElement>(num_points);
+
+    srs::IO<Curve>::read_transcript_g1(monomials_, num_points, path);
+    scalar_multiplication::generate_pippenger_point_table<Curve>(monomials_, monomials_, num_points);
+}
+
+template <typename Curve>
+typename Curve::Element Pippenger<Curve>::pippenger_unsafe(ScalarField* scalars, size_t from, size_t range)
```
Rumata888, line 36: This function and file in general could definitely use some comments. Especially around:
1) The expected scalar/point structure
2) What an "unsafe" pippenger means. That's the first question every newcomer asks when they start working with pippenger in our codebase

cpp/src/barretenberg/ecc/scalar_multiplication/runtime_states.cpp
```diff
@@ -45,13 +48,9 @@ pippenger_runtime_state::pippenger_runtime_state(const size_t num_initial_points
...
     for (size_t i = 0; i < num_threads; ++i) {
         const size_t thread_offset = i * points_per_thread;
-        memset((void*)(point_pairs_1 + thread_offset + (i * 16)),
-               0,
-               (points_per_thread + 16) * sizeof(g1::affine_element));
-        memset((void*)(point_pairs_2 + thread_offset + (i * 16)),
-               0,
-               (points_per_thread + 16) * sizeof(g1::affine_element));
-        memset((void*)(scratch_space + thread_offset), 0, (points_per_thread) * sizeof(fq));
+        memset((void*)(point_pairs_1 + thread_offset + (i * 16)), 0, (points_per_thread + 16) * sizeof(AffineElement));
```
Rumata888, line 51: Maybe we need to specify what these 16 and 64 numbers mean?

cpp/src/barretenberg/grumpkin_srs_gen/grumpkin_srs_gen.cpp
```diff
@@ -0,0 +1,37 @@
+#include <iostream>
+#include <fstream>
+
+#include "barretenberg/ecc/curves/grumpkin/grumpkin.hpp"
+#include "barretenberg/srs/io.hpp"
+
+/* Generates a monomial basis Grumpkin SRS for testing purposes.
+   We only provide functionality create a single transcript file.
+   The SRS has the form [1]_1, [x]_1, [x^2]_1, ... where x = 2. */
+int main(int argc, char** argv)
```
Rumata888, line 13: Maybe we should add a TODO to change this logic in the future, since we are not going to use this format in the future but rather sample hashed points?

cpp/src/barretenberg/ecc/scalar_multiplication/scalar_multiplication.hpp
```diff
@@ -0,0 +1,272 @@
...
+/**
+ * pointers that describe how to add points into buckets, for the pippenger algorithm.
+ * `wnaf_table` is an unrolled two-dimensional array, with each inner array being of size `n`,
+ * where `n` is the number of points being multiplied. The second dimension size is defined by
+ * the number of pippenger rounds (fixed for a given `n`, see `get_num_rounds`)
+ *
+ * An entry of `wnaf_table` contains the following three pieces of information:
+ * 1: the point index that we're working on. This is stored in the high 32 bits
+ * 2: the bucket index that we're adding the point into. This is stored in the low 31 bits
+ * 3: the sign of the point we're adding (i.e. do we actually need to subtract). This is stored in the 32nd bit.
```
Rumata888: Is the lowest bit the first bit or the 0th bit?

cpp/src/barretenberg/srs/io.hpp
```diff
@@ -17,28 +18,381 @@ struct Manifest {
...
+
+template <typename Curve> class IO {
+    using Fq = typename Curve::BaseField;
+    using Fr = typename Curve::ScalarField;
+    using Element = typename Curve::Element;
+    using AffineElement = typename Curve::AffineElement;
+
+    static constexpr size_t BLAKE2B_CHECKSUM_LENGTH = 64;
+
+    static size_t get_transcript_size(const Manifest& manifest)
```
Rumata888, line 70: Maybe we should formally specify the structure of the transcript somewhere in this file. It seems fairly simple, but it will save the person interested a minute or two.

cpp/srs_db/grumpkin/monomial/README.md
codygunton: #492

cpp/src/barretenberg/srs/io.hpp
```diff
@@ -17,28 +18,381 @@ struct Manifest {
...
+        }
+
+        return result;
+    }
+
+    static void read_manifest(std::string const& filename, Manifest& manifest)
+    {
+        std::ifstream file;
+        file.open(filename, std::ifstream::binary);
+        file.read((char*)&manifest, sizeof(Manifest));
```
Rumata888, line 89: You are not checking that enough data has been read here. If the file is small you might start accessing uninitialized memory

cpp/src/barretenberg/ecc/scalar_multiplication/pippenger.cpp
codygunton: Agree, but I can't do this now. You made an issue here #486

cpp/src/barretenberg/srs/io.hpp
```diff
@@ -17,28 +18,381 @@ struct Manifest {
...
+        size = amount ? amount : get_file_size(filename);
+
+        std::ifstream file;
+        file.open(filename, std::ifstream::binary);
+        file.seekg((int)offset);
+
+        // Read the desired size, but return the actual size read
+        file.read(buffer, (int)size);
+        if (!file) {
+            ptrdiff_t read = file.gcount();
```
Rumata888, line 134: I think this will fail, if initial amount is zero, but offset is not

cpp/src/barretenberg/ecc/scalar_multiplication/runtime_states.cpp
codygunton: The actual implementation of Pippenger is outside of the scope of this PR, which is simply a refactor that doesn't touch the core components (might need to manually construct some diffs to see this, unfortunately). Agree we need plenty of documentation here, and noted your comment in #486

cpp/src/barretenberg/grumpkin_srs_gen/grumpkin_srs_gen.cpp
codygunton: Mentioned in #492

Rumata888 (approved):
Everything seems fine. There are just 2 issues I'd mention:
1) When we get the amount of points just from the file size we might run into a situation where we fail, because we ignore the offset in computation
2) We need to specify somewhere in the srs generator that it is not supposed to be used in production (and since we are using IPA we don't need a structured srs, we can simply get hashed points)

PR #414, opened 2023-05-03
Zw/noir recursion 2

zac-williamson:
# Description

PR reduces cost of recursive proof composition in UltraPlonk to under 2^18 constraints.

The cost of "maximal" double recursion is 523,506; under the 2^19 threshold. "Maximal" means the circuit verifies 2 UltraPlonk proofs, where both inner proofs are themselves recursive and contain an aggregation object that must be accumulated by the outer circuit. (old cost pre zw/noir-recursion was 738,038)

Additionally, several security flaws have been noticed and fixed.

### Security Fixes

* pedersen_lookup::hash_single . Fixed bug where we were not validating the sum of the decomposed bit slices matched the input scalar.

* stdlib::recursion::verify_proof . When recovering an aggregation object from public inputs, we now apply range constraints to validate the bigfield limbs have the intended size

* stdlib::recursion::verify_proof . When extracting group elements from the proof, we were constructing biggroup elements directly from the raw proof data. These witnesses were not being linked to the group elements being hashed by stdlib::recursion::transcript . This is now fixed.

* stdlib::recursion::transcript . Previously when splitting field elements into 2 slices to better format the hash preimage buffer, we were not validating that the sum of the slices was less than the field modulus (over the integers). This meant the hash preimage was not unique per proof. This is now fixed.

### Changes 

* UltraComposer::decompose_into_default_range does not create extra range tables to handle "leftover" bits if the range table size is large (e.g. a 27-bit range check will not be decomposed into a 14-bit and a 13-bit check. Instead, 2 14-bit checks will be used in combination with an arithmetic gate to prevent overflow).

* pedersen_lookup::merkle_damgard_compress . The "size" parameter of the Merkle-Damgard construction has been moved to the start of the hash block (not the end). This removes one call to `hash_single` in the case where the IV is constant.

* UltraComposer evaluate_non_native_field_multiplication is no longer cached. evaluate_partial_non_native_field_multiplication is cached. Saves ~6,000+ gates

* pedersen_lookup::compress_with_relaxed_range_constraints . New method that does not apply extra range checks when calling `pedersen_lookup::hash_single`. Used when compressing the recursive transcript and the verification key. The extra range checks are required to ensure the decomposed bit slices that represent the input scalars sum to a value that is < the field modulus (over the integers). We know these values are 31 bytes from previous range constraints and therefore this cannot happen.

* pedersen_lookup::hash_single . Reduced constraints required when performing the above mentioned range check

* bigfield::operator- . Fixed bug where we were not using the UltraPlonk code path to save an addition gate. Fixed bug where UltraPlonk code path computed the wrong output.

* biggroup::validate_on_curve uses `mult_madd` to save on 1 field reduction 

* big group::create_group_element_rom_tables / read_group_element_rom_tables now tracks the maximum value of bigfield limbs added into / read from the table (tracking is worst-case. When reading a limb, its maximum value is the max of every limb in the table).

* biggroup::create_group_element_rom_tables will generate size-6 lookup tables if doing so reduces the total number of tables required 

* biggroup::add_sub new method, computes A + B and A - B. Used in biggroup::create_group_element_rom_tables. Prevents a redundant bigfield addition, subtraction and `assert_is_not_equal` call.

* biggroup::quadruple_and_add will perform 2 doublings and add a vector of points into the result. Used in `biggroup_batch_mul` to save 3 field reductions per iteration

* biggroup::multiple_mongtomery_ladder replaces several `montgomery_ladder` implementations (now deleted). Will perform repeated iterations of the Montgomery ladder (A + B + A). Each successive iteration saves on 1 field reduction at the cost of increased additions.

*  biggroup_nafs replaced calls to `field_t::create_range_constraint` with direct calls to the Composer's method. For UltraPlonk this will prevent an unnecessary addition gate being constructed for each bit.

* biggroup_tables : removed unnecessary `reduce()` calls whenever a negated point was written into the table

 * stdlib::recursion::aggregation_state no longer calls `normalize` on the aggregation object, instead calling `reduce` (the top-level verifier can validate the final outputs are in the correct field
 
 * stdlib::recursion::aggregation_state only calls `normalize` on an aggregation object being added into public inputs 
 
 * stdlib::recursion::verification_key::compress now optimally packs inputs into a minimal number of field elements prior to calling `pedersen::compress`. Saves >10,000 constraints (previously we were splitting each group element into 16 68-bit limbs and hashing them as individual field elements)
 
 
 # Checklist:


cpp/src/barretenberg/plonk/composer/ultra_composer.cpp
```diff
@@ -1138,7 +1138,29 @@ std::vector<uint32_t> UltraComposer::decompose_into_default_range(const uint32_t
...
+                create_add_gate({
+                    .a = limb_idx,
+                    .b = zero_idx,
+                    .c = diff_idx,
+                    .a_scaling = 1,
+                    .b_scaling = 0,
+                    .c_scaling = 1,
+                    .const_scaling = -barretenberg::fr(last_limb_range),
+                });
+                create_new_range_constraint(diff_idx, sublimb_mask);
```
Rumata888, line 1160: You need to range constrain the limb itself,too. Otherwise it is possible to underflow.

cpp/src/barretenberg/proof_system/circuit_constructors/ultra_circuit_constructor.cpp
```diff
@@ -1534,32 +1545,28 @@ std::array<uint32_t, 2> UltraCircuitConstructor::evaluate_partial_non_native_fie
...
+        .b = input.b,
+        .lo_0 = lo_0_idx,
+        .hi_0 = hi_0_idx,
+        .hi_1 = hi_1_idx,
+    };
+    cached_partial_non_native_field_multiplications.emplace_back(cache_entry);
     return std::array<uint32_t, 2>{ lo_0_idx, hi_1_idx };
 }
+/**
+ * @brief The following tests FAILED:
```
Rumata888: Are they still failing?

cpp/src/barretenberg/stdlib/recursion/verifier/verifier.hpp
```diff
@@ -380,33 +370,13 @@ aggregation_state<Curve> verify_proof(typename Curve::Composer* context,
     for (const auto& to_add : elements_to_add) {
         opening_result = opening_result + to_add;
     }
-    opening_result = opening_result.normalize();
 
     g1_ct rhs = g1_ct::template wnaf_batch_mul<128>(rhs_elements, rhs_scalars);
-    rhs = rhs + PI_Z;
-    rhs = (-rhs).normalize();
-
-    std::vector<uint32_t> proof_witness_indices{
```
codygunton, line 389: This is moved to a member function on the `aggregation_state`.

cpp/src/barretenberg/stdlib/recursion/aggregation_state/aggregation_state.hpp
```diff
@@ -30,7 +30,28 @@ template <typename Curve> struct aggregation_state {
 
     void add_proof_outputs_as_public_inputs()
     {
-        ASSERT(proof_witness_indices.size() > 0);
+        ASSERT(proof_witness_indices.size() == 0);
+
+        P0 = P0.reduce();
```
codygunton, line 50: Looks like these reductions were missing before :heavy_check_mark:

cpp/src/barretenberg/stdlib/recursion/verification_key/verification_key.hpp
```diff
@@ -20,10 +20,139 @@
...
+    using witness_pt = witness_t<Composer>;
+
+    Composer* context;
+
+    PedersenPreimageBuilder(Composer* ctx = nullptr)
+        : context(ctx){};
+
+    field_pt compress(const size_t hash_index)
+    {
+        // we can only use relaxed range checks in pedersen::compress iff bits_per_element < modulus bits
```
codygunton, line 51: Not verifying whether this condition is sharp in the current review.

cpp/src/barretenberg/stdlib/recursion/verification_key/verification_key.hpp
```diff
@@ -20,10 +20,139 @@
...
+        }
+        if constexpr (Composer::type == ComposerType::PLOOKUP) {
+            return pedersen_plookup_commitment<Composer>::compress_with_relaxed_range_constraints(preimage_data,
+                                                                                                  hash_index);
+        } else {
+            return pedersen_commitment<Composer>::compress(preimage_data, hash_index);
+        }
+    }
+
+    std::vector<field_pt> work_element;
```
codygunton: It would be helpful to explain what these data members are.

cpp/src/barretenberg/plonk/proof_system/verification_key/verification_key.cpp
```diff
@@ -49,47 +49,38 @@ barretenberg::fr compress_native_evaluation_domain(barretenberg::evaluation_doma
...
+        write(preimage_data, selector.y);
+        write(preimage_data, selector.x);
     }
 
+    write(preimage_data, eval_domain.root);
+
     barretenberg::fr compressed_key;
-    if (proof_system::ComposerType(composer_type) == proof_system::ComposerType::PLOOKUP) {
-        compressed_key = crypto::pedersen_commitment::lookup::compress_native(preimage_data, hash_index);
+    if (proof_system::ComposerType(composer_type) == ComposerType::PLOOKUP) {
```
suyash67, line 79: Why is `hash_index` removed? I think we had added it so as to not use the default generators (`hash_index = 0`) since we'd end up using overlapping generators for Merkle hashing and VK. Is that okay?

cpp/src/barretenberg/stdlib/recursion/verification_key/verification_key.hpp
```diff
@@ -20,10 +20,139 @@
...
+    size_t current_bit_counter = 0;
+
+    void add_element(const field_pt& element) { slice_element(element, 256); }
+
+    void add_element_with_existing_range_constraint(const field_pt& element, const size_t num_bits)
+    {
+        slice_element(element, num_bits);
+    }
+
+    void slice_element(const field_pt& element, const size_t num_bits)
```
codygunton, line 105: This is a refactor of the lambda function `split` that was defined in `apply_fiat_shamir`. That function used to split into bytes whereas this one splits into bits.

cpp/src/barretenberg/stdlib/recursion/transcript/transcript.hpp
```diff
@@ -133,196 +132,136 @@ template <typename Composer> class Transcript {
...
-                work_element = (lo * lo_shift);
-                work_element = work_element.normalize();
-            }
-        };
-
-        std::vector<field_pt> compression_buffer;
         field_pt working_element(context);
 
-        size_t byte_counter = 0;
+        PedersenPreimageBuilder<Composer, 31 * 8> preimage_buffer(context);
```
codygunton: Magic number 31 * 8 = 248? What's the significance of this?

cpp/src/barretenberg/stdlib/recursion/verification_key/verification_key.hpp
```diff
@@ -120,9 +211,15 @@ template <typename Curve> struct verification_key {
         key->num_public_inputs = witness_t<Composer>(ctx, input_key->num_public_inputs);
         key->domain = evaluation_domain<Composer>::from_witness(ctx, input_key->domain);
         key->contains_recursive_proof = witness_t<Composer>(ctx, input_key->contains_recursive_proof);
-
         for (const auto& [tag, value] : input_key->commitments) {
-            key->commitments.insert({ tag, Curve::g1_ct::from_witness(ctx, value) });
+            // We do not perform on_curve() circuit checks when constructing the Curve::g1_ct element.
```
codygunton, line 239: You're saying this hash check would be done in a circuit? Maybe I misunderstand, but it seems like even if the circuit creator (who constructed the `input_key`) was honest, we're still relying on witnesses being constructed in the default way. Anyway, you make it clear in your comment that there's ambiguity around the security model, something that we can address in the audit.

cpp/src/barretenberg/stdlib/recursion/verification_key/verification_key.hpp
```diff
@@ -189,71 +286,65 @@ template <typename Curve> struct verification_key {
...
-        preimage_data.push_back(Composer::type);
-        preimage_data.push_back(compressed_domain);
-        preimage_data.push_back(num_public_inputs);
+        PedersenPreimageBuilder<Composer> preimage_buffer(context);
+
+        field_t<Composer> composer_type = witness_t<Composer>::create_constant_witness(context, Composer::type);
+        domain.generator.create_range_constraint(16, "domain.generator");
+        domain.domain.create_range_constraint(32, "domain.generator");
+        num_public_inputs.create_range_constraint(32, "num_public_inputs");
+        preimage_buffer.add_element_with_existing_range_constraint(composer_type, 8);
```
codygunton, line 319: I like this `with_existing` interface.

cpp/src/barretenberg/stdlib/recursion/verifier/verifier.hpp
```diff
@@ -320,8 +307,7 @@ aggregation_state<Curve> verify_proof(typename Curve::Composer* context,
         opening_elements.push_back(previous_output.P0);
         opening_scalars.push_back(random_separator);
 
-        rhs_elements.push_back(
-            (-(previous_output.P1)).reduce()); // TODO: use .normalize() instead? (As per defi bridge project)
```
codygunton, line 324: `reduce` call is now in `add_proof_outputs_as_public_inputs`

cpp/src/barretenberg/stdlib/recursion/verifier/verifier.hpp
```diff
@@ -369,7 +359,7 @@ aggregation_state<Curve> verify_proof(typename Curve::Composer* context,
         opening_elements.push_back(g1_ct(x0, y0));
         opening_scalars.push_back(recursion_separator_challenge);
 
-        rhs_elements.push_back((-g1_ct(x1, y1)).normalize());
```
codygunton, line 372: Not immediately clear why this isn't needed, similarly for `opening_result` below.

codygunton (commented):
I looked over what's in stdlib/recursion. As discussed elsewhere, I'm not doing a thorough review here. This is is a preliminary PR introducing complex circuit optimizations, and, as such, recursion in Barretenberg should not be put into production anywhere before a deep audit is conducted (note this is _not_ the code that was put into production as part of Aztec Connect).

cpp/src/barretenberg/stdlib/primitives/biggroup/biggroup_impl.hpp
```diff
@@ -294,200 +327,220 @@ element<C, Fq, Fr, G> element<C, Fq, Fr, G>::montgomery_ladder(const chain_add_a
...
+        std::vector<Fq> lambda1_right;
+        std::vector<Fq> lambda1_add;
+
+        if (i == 0) {
+            lambda1_add.emplace_back(-y);
+        } else {
+            lambda1_left = previous_y.mul_left;
+            lambda1_right = previous_y.mul_right;
+            lambda1_add = previous_y.add;
+            // std::copy(previous_y.mul_left.begin(), previous_y.mul_left.end(), std::back_inserter(lambda1_left));
```
Rumata888: Do we need these commented out lines?

Rumata888 (changes requested):
At a glance everything seems ok, except for the soundness issue in range constraint

suyash67 (commented):
Pedersen stuff looks okay to me, just left one comment about hash index.

cpp/src/barretenberg/stdlib/recursion/verification_key/verification_key.hpp
zac-williamson: I think it's ok because the Prover does not have the ability to influence the witness. We can assume that some higher-level logic is going to check the correctness of the verification key hash (e.g. it comes from a valid leaf in a Merkle tree). If circuit creator points are on curve and the Prover-supplied witness is not, the vkey hash won't ma

cpp/src/barretenberg/stdlib/recursion/verifier/verifier.hpp
zac-williamson: The multiscalar multiplication algorithm can handle unreduced points. Similar to how stdlib methods that use `field_ct` do not need to call `normalize` constantly (...anymore)

cpp/src/barretenberg/plonk/proof_system/verification_key/verification_key.cpp
zac-williamson: the methods to compress uint8_t vectors did not take in a hash index. Will fix the proper way but adding this support

cpp/src/barretenberg/stdlib/recursion/transcript/transcript.hpp
zac-williamson: 31 bytes = largest number of bytes we can fit into a field element. Will update w. named constant

cpp/src/barretenberg/proof_system/circuit_constructors/ultra_circuit_constructor.cpp
zac-williamson: good catch. No. That was debugging info from when tests were failing. have removed.

Rumata888 (approved):
The bug is fixed

PR #433, opened 2023-05-05
feat(dsl)!: Arbitrary depth recursion

vezenovm:
# Description

This PR was pushed for testing the full CI before merging into the main `mv/noir-recursion` branch. This PR adds a `nested_aggregation_object` to a RecursionConstraint in `acir_format`. It also includes tests in `acir_format` and `acir_proofs` that show full composition of recursive proofs.

This PR does a couple other things to enable interop with Noir/DSLs. The `verify_recursive_proof` function which writes the output aggregation object to a buffer now accepts the number of public inputs of the inner proof while also handling the input aggregation object. This will enable DSLs to simulate the result of a recursion operation without running through the proving operation. 

The `init_verification_key` method also now writes the recursive proof indices of a given key. This is used by method looking to simulate the result of recursive verification. Our circuit require's that the recursive proof indices are circuits constants, thus we must fetch the information from the backend in some manner. I chose to keep this as the verification key and then whomever is interacting with the `acir_proofs` package can extract the indices from the verification key witnesses.

# Checklist:

- [X] I have [linked](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) this pull request to the issue(s) that it resolves.
- [] There are no circuit changes, OR specifications in `/markdown/specs` have been updated.
- [] There are no circuit changes, OR a cryptographer has been assigned for review.
- [X] New functions, classes, etc. have been documented according to the doxygen comment format. Classes and structs must have `@brief` describing the intended functionality.
- [X] If existing code has been modified, such documentation has been added or updated.


cpp/src/barretenberg/dsl/acir_format/recursion_constraint.cpp
```diff
@@ -113,12 +122,14 @@ void create_recursion_constraints(Composer& composer, const RecursionConstraint&
 
     // Assign the output aggregation object to the proof public inputs (16 field elements representing two
     // G1 points)
-    result.add_proof_outputs_as_public_inputs();
+    // result.add_proof_outputs_as_public_inputs();
```
zac-williamson:
```suggestion
```

cpp/src/barretenberg/dsl/acir_proofs/acir_proofs.cpp
```diff
@@ -296,19 +327,30 @@ size_t verify_recursive_proof(uint8_t const* proof_buf,
...
 
-    transcript::Manifest manifest = acir_format::Composer::create_unrolled_manifest(1);
+    transcript::Manifest manifest = acir_format::Composer::create_unrolled_manifest(num_public_inputs);
     // We currently only support RecursionConstraint where inner_proof_contains_recursive_proof = false.
     // We would either need a separate ACIR opcode where inner_proof_contains_recursive_proof = true,
     // or we need non-witness data to be provided as metadata in the ACIR opcode
-    std::shared_ptr<acir_format::verification_key_ct> vkey =
-        acir_format::verification_key_ct::template from_field_pt_vector<false>(&composer, key_fields);
+    // recursively verify the proof
+    std::array<uint32_t, acir_format::RecursionConstraint::AGGREGATION_OBJECT_SIZE> nested_aggregation_object = {};
```
zac-williamson, line 335: This method could be quite slow as we are building a circuit that technically we do not need to (if we made some tweaks).

One alternative to this would be to enable the native verification algorithm to output an aggregation object. i.e. the native verifier performs a pairing check in the last step. Instead of this it would return the G1 points it inputs into the pairing; these would be the new aggregation object. 

Possibly complex as it may require also making the native verifier use the recursion-friendly hash function (not keccak). Maybe create this as an issue that we can address later?

zac-williamson (approved):
Looks good to me. Flagged an issue we might want to create for a future optimisation.

cpp/src/barretenberg/dsl/acir_proofs/acir_proofs.cpp
vezenovm: It does get a little complex with determining which hash function to use, but I don't think it is too bad and may be worth it to not generate a circuit. Is it very inefficient if we are not generating a proof? Either way I can open up an issue for this.

cpp/src/barretenberg/dsl/acir_format/recursion_constraint.cpp
vezenovm: I forgot about this commented out code. Do you think I should just remove the comment above as well? It may be confusing to leave it otherwise?

vezenovm: I removed the entire comment for now but can add it back in the main PR if a further explanation is deemed necessary

cpp/src/barretenberg/dsl/acir_proofs/acir_proofs.cpp
vezenovm: Issue made here: https://github.com/AztecProtocol/barretenberg/issues/447

PR #434, opened 2023-05-06
WASI-20. Low memory prover (512k possible). Native pthread support (15s wasm time).

charlielye:
# barretenberg

Wins:
* Move to WASI 20.
* Latest clang etc.
* Proper static initialisation of globals.
* Can remove OpenMP (for now can switch to new threading with `-DOMP_MULTITHREADING=OFF`).
* Can remove asyncify (conditional variables + threads as alternative).
* Less memory usage and copying memory.
* Much faster performance, granular threading, (AC was very coarse).
* Same threading model as native code.
* Could get rid of worker queue (but don't yet, just disabled cbinds).
* Get rid of host environment threading (PooledPippenger PooledFft) in ts code.
* Unified bb.js browser/node code with very simple abstraction layer to deal with differences.
* Same `simple.rawtest.ts` works in node and when compiled to browser via webpack.
* No more node polyfills in bb.js.

How:
* wasi "_initialize" solves re-entry crash, enables move to wasi 20.
* wasi-libc bleeding edge fork for increased (not perfect) join stability.
* parallel_for replaces OpenMP. (several versions, using "parallel_for_mutex_pool" to avoid stability bug).
* SlabAllocator enables 512k gates within 4GB.
* Beyond that, PolynomialStoreCache and host engine swapping (unused currently as it's kinda janky TBH).
* Polynomial coeffs in shared_ptr + SlabAllocator.

# bb.js

We are moving towards standalone prover/verifier binaries to be maintained by bb team(s), and to be consumed by noir team(s) (and whoever else).
This is the first implementation of such a prover/verifier.

Wins:

* Completely architecture/os agnostic. As a wasm it runs everywhere, even the browser.
* Handles up to 512k gates, proving time 18s on M1 Mac (builds off prior work).
* New cleaned up AcirComposer interface that's simple and sane to use.


codygunton:
Charlie gave me a tour of changes relevant to the cryptography team. Otherwise we are foregoing a review due to the massive scope, the amount of new code, the fact that things are very much in flux, and the fact that this is not intended for production.

PR #417, opened 2023-05-03
feat!: add support for ROM and RAM ACVM opcodes

guipublic:
# Description

This PR modifies the acir_format structure to take into account memory constraints in UP.
They are defined as block constraints which contain the memory trace on each 'block of data' (i.e an array) and are translated into table lookups using the stdlib memory helpers.

# Checklist:

- [X] I have reviewed my diff in github, line by line.
- [X] Every change is related to the PR description.
- [ ] I have [linked](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) this pull request to the issue(s) that it resolves.
- [X] There are no unexpected formatting changes, superfluous debug logs, or commented-out code.
- [X] There are no circuit changes, OR specifications in `/markdown/specs` have been updated.
- [X] There are no circuit changes, OR a cryptographer has been assigned for review.
- [ ] I've updated any terraform that needs updating (e.g. environment variables) for deployment.
- [ ] The branch has been rebased against the head of its merge target.
- [X] I'm happy for the PR to be merged at the reviewer's next convenience.
- [ ] New functions, classes, etc. have been documented according to the doxygen comment format. Classes and structs must have `@brief` describing the intended functionality.
- [X] If existing code has been modified, such documentation has been added or updated.

PR #436, opened 2023-05-11
feat(dsl)!: add hash index to pedersen constraint

sirasistant:
# Description

Adds the hash_index value to the PedersenConstraint used by noir (previously it was always using the default, zero)
Also adds a new cbind to be able to call plookup commit with hash index.

# Checklist:


cpp/src/barretenberg/dsl/acir_format/pedersen.hpp
```diff
@@ -7,6 +7,8 @@ namespace acir_format {
 // P = xG + bH
 struct PedersenConstraint {
     std::vector<uint32_t> scalars;
+    uint32_t hash_index;
```
phated, line 10: can this be moved to the bottom of the struct and be defaulted so the change isn't breaking?

kevaundray: I'm not sure how this would work when `read` is called; I think you would need to do a conditional read, but imagine that this is in the middle of the buffer, then its not easy to know if the next byte is for the next constraint or if its the hash_index

sirasistant: Yes, I think there is no way of making this a non-breaking change with our current serialization

sirasistant:
I was going to change the noir ref to use my pedersen with domain separator branch but another barretenberg breaking change has landed already https://github.com/AztecProtocol/barretenberg/pull/417 which complicates more having a noir branch that works with both breaking changes. 
Anyway here is the noir CI run that runs noir tests with this BB branch and the updated aztec_backend and acvm https://github.com/noir-lang/noir/actions/runs/4978028027/jobs/8909135154

kevaundray (approved):

PR #438, opened 2023-05-11
DSL: Add valid verifier dummy data for ecdsa

kevaundray:
# Description

When the verifier creates a constraint system, they will use 0s as the witness values. This will cause panics since a lot of the ecdsa code adds assertions to ensure that certain data is not zero or valid. One of these being the r, s values need to be non-zero and the public key, signatures need to valid.

This PR creates a dummy signature for the verifier. 

Note: A problem with this approach and wherever else in the codebase this is applied. If the prover passes in a zero intentionally/accidentally, then their program will crash. 

Please provide a paragraph or two giving a summary of the change, including relevant motivation and context.

# Checklist:

### DIFF
--- a/cpp/src/barretenberg/dsl/acir_format/acir_format.cpp
+++ b/cpp/src/barretenberg/dsl/acir_format/acir_format.cpp
@@ -234,7 +234,7 @@ Composer create_circuit_with_witness(const acir_format& constraint_system,
 
     // Add ECDSA constraints
     for (const auto& constraint : constraint_system.ecdsa_constraints) {
-        create_ecdsa_verify_constraints(composer, constraint);
+        create_ecdsa_verify_constraints<true>(composer, constraint);
     }
 
     // Add blake2s constraints
@@ -320,7 +320,7 @@ Composer create_circuit_with_witness(const acir_format& constraint_system, std::
 
     // Add ECDSA constraints
     for (const auto& constraint : constraint_system.ecdsa_constraints) {
-        create_ecdsa_verify_constraints(composer, constraint);
+        create_ecdsa_verify_constraints<true>(composer, constraint);
     }
 
     // Add blake2s constraints
@@ -404,7 +404,7 @@ void create_circuit_with_witness(Composer& composer, const acir_format& constrai
 
     // Add ECDSA constraints
     for (const auto& constraint : constraint_system.ecdsa_constraints) {
-        create_ecdsa_verify_constraints(composer, constraint);
+        create_ecdsa_verify_constraints<true>(composer, constraint);
     }
 
     // Add blake2s constraints
--- a/cpp/src/barretenberg/dsl/acir_format/acir_format.cpp
+++ b/cpp/src/barretenberg/dsl/acir_format/acir_format.cpp
@@ -13,6 +13,11 @@ void read_witness(Composer& composer, std::vector<barretenberg::fr> witness)
 }
 using curve = proof_system::plonk::stdlib::secp256k1<Composer>;
 
+// Add dummy constraints for ECDSA because when the verifier creates the
+// constraint system, they usually use zeroes for witness values.
+//
+// This does not work for ECDSA as the signature, r, s and public key need
+// to be valid.
 void dummy_ecdsa_constraint(Composer& composer, EcdsaSecp256k1Constraint const& input)
 {
 
--- a/cpp/src/barretenberg/dsl/acir_format/acir_format.cpp
+++ b/cpp/src/barretenberg/dsl/acir_format/acir_format.cpp
@@ -1,5 +1,6 @@
 #include "acir_format.hpp"
 #include "barretenberg/common/log.hpp"
+#include "barretenberg/crypto/ecdsa/ecdsa.hpp"
 
 namespace acir_format {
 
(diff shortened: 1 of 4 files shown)


cpp/src/barretenberg/dsl/acir_format/ecdsa_secp256k1.test.cpp
```diff
@@ -40,6 +40,7 @@ size_t generate_ecdsa_constraint(acir_format::EcdsaSecp256k1Constraint& ecdsa_co
         const auto byte = static_cast<uint8_t>(hashed_message[i]);
         witness_values.emplace_back(byte);
     }
+    std::cout << message_in.size() << std::endl;
```
kevaundray:
```suggestion
```

cpp/src/barretenberg/dsl/acir_format/acir_format.cpp
```diff
@@ -145,6 +191,7 @@ Composer create_circuit(const acir_format& constraint_system,
 
     // Add ECDSA constraints
     for (const auto& constraint : constraint_system.ecdsa_constraints) {
+        dummy_ecdsa_constraint(composer, constraint);
```
vezenovm: Could we template `create_ecdsa_verify_constraints` to be `create_ecdsa_verify_constraints<has_valid_witness_assignment>`? Then we can encapsulate `dummy_ecdsa_constraint` inside of `create_ecdsa_verify_constraints` and move `dummy_ecdsa_constraint` out of `acir_format.cpp`.

cpp/src/barretenberg/dsl/acir_format/acir_format.cpp
```diff
@@ -10,6 +11,50 @@ void read_witness(Composer& composer, std::vector<barretenberg::fr> witness)
         composer.variables[i + 1] = witness[i];
     }
 }
+using curve = proof_system::plonk::stdlib::secp256k1<Composer>;
```
vezenovm: I believe we have `secp256k1_ct` in our dsl types

cpp/src/barretenberg/dsl/acir_format/ecdsa_secp256k1.cpp
```diff
@@ -5,6 +5,7 @@
 namespace acir_format {
 
 using namespace proof_system::plonk;
+using curve = proof_system::plonk::stdlib::secp256k1<acir_format::Composer>;
```
vezenovm: Some comment as earlier, why is this being done if we have `secp256k1_ct`?

cpp/src/barretenberg/dsl/acir_format/ecdsa_secp256k1.test.cpp
```diff
@@ -110,6 +110,31 @@ TEST(ECDSASecp256k1, TestECDSAConstraintSucceed)
...
+        .sha256_constraints = {},
+        .blake2s_constraints = {},
+        .keccak_constraints = {},
+        .hash_to_field_constraints = {},
+        .pedersen_constraints = {},
+        .compute_merkle_root_constraints = {},
+        .constraints = {},
+    };
+    auto crs_factory = std::make_unique<proof_system::ReferenceStringFactory>();
+    auto composer = create_circuit(constraint_system, std::move(crs_factory));
```
vezenovm, line 138: Maybe leave a comment that we are just testing that we can construct an ECDSA circuit without a witness and reference the `dummy_ecdsa_constraint`? As the test name is `TestECDSAConstraintSucceedSeparate` I think it would be confusing to someone looking at this the first time why there is not an `EXPECT_EQ(..., true)`

vezenovm (commented):
Some minor comments with regards to code organization

cpp/src/barretenberg/dsl/acir_format/acir_format.cpp
kevaundray: We had this before in the first commit, but I was not a fan of the templating

vezenovm: As per slack discussion: a boolean flag could work too, I just think `dummy_ecdsa_constraint` should be inside the ecdsa_constraint file not acir_format

kevaundray: Changed it to a boolean flag as it makes sense to move it into ecdsa_secp

cpp/src/barretenberg/dsl/acir_format/ecdsa_secp256k1.cpp
```diff
@@ -84,9 +84,15 @@ witness_ct ecdsa_index_to_witness(Composer& composer, uint32_t index)
     return { &composer, value };
 }
 
-void create_ecdsa_verify_constraints(Composer& composer, const EcdsaSecp256k1Constraint& input)
+void create_ecdsa_verify_constraints(Composer& composer,
+                                     const EcdsaSecp256k1Constraint& input,
+                                     bool has_valid_witness_assignments)
```
vezenovm, line 89: This is a `nit` but you could set a default to `false` and avoid having to write out the param for `create_ecdsa_verify_constraints` when there is not a valid witness assignment

vezenovm (approved):

cpp/src/barretenberg/dsl/acir_format/ecdsa_secp256k1.cpp
kevaundray: That's a good point, I think we want to make the default = true though so that it fails fast and not replace witness values by default.

PR #444, opened 2023-05-15
Multi-constraint Relations

### LINKED ISSUE #427: Resolve relation batching ambiguities in UltraHonk
In Standard Honk, relations were introduced to represent individual polynomial constraints, i.e. the number powers of alpha needed to construct the larger sumcheck relation is equal to the number of relations. This allows us to defer multiplication by alpha when batching relation contributions in a way that saves a ton of field muls. For Ultra Honk it's less clear that this is the right approach. In the "elliptic widget" for example, there are really two relations, both of which require some common computation. If we split them into two relations as we've been doing, the savings we get from deferring multiplication by alpha may be outweighed by having to repeat otherwise shareable computations. It would also be convenient just for the sake of readability and code complexity. The "gen perm widget" is conceptually simple but is made up of 4 separate constraints and thus requires 4 powers of alpha. It would be nice to have a single object that handles this rather than 4 separate relations.

The current approach is to simply multiply by powers of alpha within certain ultra relations but this is technically not valid since it will lead to use of repeated powers of alpha in the batching process. An approach might be to make the external loop responsible for batching over all relations aware of the number of powers used by each relation so it can update accordingly.

ledwards2225:
# Description

We originally designed Sumcheck so that each `Relation` corresponded to a single constraint. This was reasonable for standard Honk which only required 3 relations (1 arithmetic and 2 for the permutation argument) but for Ultra Honk this approach is awkward conceptually and potentially inefficient. This PR expands the notion of a `Relation` to a set of one or more constraints, generally with a coherent purpose. For example, this allows the two constraints that ensure correct initialization and construction of the permutation grand product (formerly two separate relations) to be combined into the single `PermutationRelation`.

The changes in this work are essentially all driven by the following changes to the containers that store the relation contributions for the Sumcheck prover and verifier:
1) Prover: Conversion from a tuple of Univariates to tuple of tuples of Univariates
2) Verifier Conversion array of field elements to a tuple of arrays of field elements

A tuple of tuples of Univariates is required because each relation now contributes its own tuple of heterogeneously sized Univariates. (Similarly, for the verifier each relation contributes a differing number of field elements). This change requires the addition of some utility functions in SumcheckRound to handle compile-time access of nestled tuples. 

# Checklist:


cpp/src/barretenberg/honk/sumcheck/relations/ultra_relation_consistency.test.cpp
```diff
@@ -108,34 +109,74 @@ class UltraRelationConsistency : public testing::Test {
...
-        // Compute the expression index-by-index
-        Univariate<FF, FULL_RELATION_LENGTH> expected_evals_index{ 0 };
-        for (size_t i = 0; i < FULL_RELATION_LENGTH; ++i) {
-            // Get an array of the same size as `extended_edges` with only the i-th element of each extended edge.
-            ClaimedEvaluations evals_i = transposed_univariate_array_at(extended_edges, i);
-            // Evaluate the relation
-            relation.add_full_relation_value_contribution(
-                expected_evals_index.value_at(i), evals_i, relation_parameters);
+        // First check that the verifier's computation on individual evaluations is correct.
+        // Note: it is sufficient to check at only the first index of the input edges.
```
maramihali: Can you explain why here please

cpp/src/barretenberg/honk/sumcheck/relations/auxiliary_relation.hpp
```diff
@@ -4,26 +4,37 @@
 
 #include "../polynomials/univariate.hpp"
 #include "barretenberg/numeric/uint256/uint256.hpp"
-#include "relation.hpp"
+#include "relation_parameters.hpp"
 
 namespace proof_system::honk::sumcheck {
 
 template <typename FF> class AuxiliaryRelation {
```
maramihali, line 11: I think it would be great to have, above each of this relation, a sentence that explain the general idea of what the relation is meant to do, it would make it easier to also understand what RelationUnivariates and RelationValues are. Especially this Relation which bundles together soo many things.

cpp/src/barretenberg/honk/flavor/standard.hpp
```diff
@@ -42,7 +41,7 @@ class Standard {
     using CommitmentHandle = G1::affine_element;
     using PCSParams = pcs::kzg::Params;
 
-    static constexpr size_t NUM_WIRES = CircuitConstructor::NUM_WIRES;
+    static constexpr size_t NUM_WIRES = CircuitConstructor::NUM_WIRES; // MERGETODO
```
maramihali: what's this `MERGETODO`?

cpp/src/barretenberg/honk/sumcheck/relations/arithmetic_relation.hpp
```diff
@@ -3,7 +3,7 @@
 #include <tuple>
 
 #include "../polynomials/univariate.hpp"
-#include "relation.hpp"
+#include "relation_parameters.hpp"
```
maramihali, line 6: I think this is a really good change, I was confused about the naming of this file beforehand :)

cpp/src/barretenberg/honk/sumcheck/relations/arithmetic_relation.hpp
```diff
@@ -44,10 +50,10 @@ template <typename FF> class ArithmeticRelation {
         tmp += q_o * w_o;
         tmp += q_c;
         tmp *= scaling_factor;
-        evals += tmp;
+        std::get<0>(evals) += tmp;
     };
 
-    void add_full_relation_value_contribution(FF& full_honk_relation_value,
+    void add_full_relation_value_contribution(RelationValues& full_honk_relation_value,
```
maramihali, line 63: I think it's clear what this function is doing but also it would benefit from a bit of documentation

cpp/src/barretenberg/honk/sumcheck/relations/auxiliary_relation.hpp
```diff
@@ -4,26 +4,37 @@
...
   public:
     // 1 + polynomial degree of this relation
-    static constexpr size_t RELATION_LENGTH = 6; // degree() = WORKTODO
+    static constexpr size_t RELATION_LENGTH = 6;
+
+    static constexpr size_t NUM_CONSTRAINTS = 6;
+    static constexpr std::array<size_t, NUM_CONSTRAINTS> CONSTRAINT_LENGTH = { 6, 6, 6, 6, 6, 6 }; // Each has degree 5
```
maramihali, line 17: do we expect some constraints to have different lengths than 6 or why do we need to have a an array for this?

cpp/src/barretenberg/honk/sumcheck/relations/ultra_arithmetic_relation.hpp
```diff
@@ -20,47 +26,70 @@ template <typename FF> class UltraArithmeticRelation {
      *        (q_l * w_l) + (q_r * w_r) + (q_o * w_o) + (q_4 * w_4) + q_c +
      *        (q_arith - 1)w_4_shift ]
      *
+     *    q_arith *
+     *      (q_arith - 2) * (q_arith - 1) * (w_l + w_4 - w_l_shift + q_m)
```
maramihali, line 66: Can you make it more clear that these two are seperate identities and why are they different? it's not clear unless very familiar with the code that the second one is for `big_add` stuff

cpp/src/barretenberg/honk/sumcheck/sumcheck_round.hpp
```diff
@@ -74,79 +75,44 @@ template <typename Flavor> class SumcheckRound {
...
-        auto& univariate = std::get<idx>(univariate_accumulators);
-        std::fill(univariate.evaluations.begin(), univariate.evaluations.end(), FF(0));
-
-        if constexpr (idx + 1 < NUM_RELATIONS) {
-            reset_accumulators<idx + 1>();
-        }
-    };
-    // IMPROVEMENT(Cody): This is kind of ugly. There should be a one-liner with folding
-    // or std::apply or something.
+        zero_univariates(univariate_accumulators);
```
maramihali, line 91: what happens here should still be documented

cpp/src/barretenberg/honk/sumcheck/sumcheck_round.hpp
```diff
@@ -320,12 +262,142 @@ template <typename Flavor> class SumcheckRound {
...
+    }
+
+    /**
+     * @brief Scale Univaraites by consecutive powers of the provided challenge
+     *
+     * @param tuple Tuple of tuples of Univariates
+     * @param challenge
+     * @param current_scalar power of the challenge
+     */
+    static void scale_univariates(auto& tuple, const FF& challenge, FF current_scalar)
```
maramihali, line 319: are we scaling or are we creating a random linear combination I'm slightly confused? Also, I think this has a different purpose than `scaling_factor` present in the `add_edge_contribution` functions from Relations, right?

cpp/src/barretenberg/honk/sumcheck/sumcheck.test.cpp
```diff
@@ -402,4 +408,304 @@ TEST(Sumcheck, ProverAndVerifierLonger)
...
+    std::optional verifier_output = sumcheck_verifier.execute_verifier(relation_parameters);
+
+    ASSERT_TRUE(verifier_output.has_value());
+}
+
+/**
+ * @brief Test the Ultra Sumcheck Prover and Verifier for a real circuit
+ *
+ */
+TEST(Sumcheck, RealCircuitUltra)
```
maramihali, line 497: Shouldn't there be one (or multiple) tests, or maybe just more logic in  this one, that check ROM Consistency, RAM Timestamp, non-native arithmetic? I know their correctness is checked in the relation folder  but I'm just wary of very subtle bugs.

cpp/src/barretenberg/honk/sumcheck/sumcheck.test.cpp
```diff
@@ -402,4 +408,304 @@ TEST(Sumcheck, ProverAndVerifierLonger)
...
+
+    // Add a sort gate (simply checks that consecutive inputs have a difference of < 4)
+    a_idx = composer.add_variable(FF(0));
+    b_idx = composer.add_variable(FF(1));
+    c_idx = composer.add_variable(FF(2));
+    d_idx = composer.add_variable(FF(3));
+    composer.create_sort_constraint({ a_idx, b_idx, c_idx, d_idx });
+
+    // Add an elliptic curve addition gate
+    grumpkin::g1::affine_element p1 = crypto::generators::get_generator_data({ 0, 0 }).generator;
```
maramihali, line 551: why are the group elements from grumpkin if the scalars are from bn254? is this gate also atesting non-native arithmetic under the hood?

cpp/src/barretenberg/honk/sumcheck/relations/ultra_relation_consistency.test.cpp
ledwards2225: Added a comment. The point is that the verifier evaluates the relations/identities at a single evaluation point. The extended edges give us MAX_RELATION_LENGTH-many points to choose from so we just arbitrarily choose the 0th. The full extended edges are only needed for the provers computations.

cpp/src/barretenberg/honk/flavor/standard.hpp
ledwards2225: Woops, just left over from the somewhat tricky merge with master after some of my other Sumcheck changes went in prior to this PR. Removed

cpp/src/barretenberg/honk/sumcheck/relations/auxiliary_relation.hpp
ledwards2225: It just so happens that all constraints for this relation have the same length. (I think it turned out this way because Zac was trying to squeeze as much functionality out of each one as possible while not increasing the degree of the overall Sumcheck relation). If you look at say the elliptic or permutation relations you'll see that they dont in general all have the same degree.

cpp/src/barretenberg/honk/sumcheck/sumcheck_round.hpp
ledwards2225: Thanks, added comments where the function is called. There is also documentation on the function itself

cpp/src/barretenberg/honk/sumcheck/sumcheck_round.hpp
ledwards2225: This function is simply scaling each univariate by consecutive powers of the provided challenge. No summing/batching is done here

cpp/src/barretenberg/honk/sumcheck/sumcheck.test.cpp
ledwards2225: The real tests for all of that functionality should live in what is currently called the UltraHonkComposerTests which construct circuits with the functionality of interest then do full proof construction/verification. This is just a supplemental test that I've added to bridge the gap between checking the relations individually and doing full proof verification. Here we're checking that we can construct and verify proofs only up to and including Sumcheck. The circuit is very non-trivial but certainly doesn't capture every type of gate supported by Ultra Honk.

cpp/src/barretenberg/honk/sumcheck/sumcheck.test.cpp
ledwards2225: This elliptic curve addition gate handles the addition of two points where one is scaled by the endomorphism constant and/or negated. My understanding of when this is useful is pretty vague (this example was copied directly from the elliptic add gate test in the composer tests). It's not clear to me why endomorphism constant is a grumpkin::fq (bberg::fr) -- this is firmly in @codygunton's wheelhouse, maybe he can shed some light

codygunton: Any curve of the form $Y^2 = X^3  + aX + b$ with $a=0$, has an 'endomorphism' given by $\psi:(x,y)\mapsto (\beta \cdot x, y)$, where  $\beta$ lives in the same field that contains $b$ and $\beta$ satisfies $\beta^3 = 1$. The reason we care about this is that, when you have such an endomorphism, you can turn one scalar multiplication $[n]P$ into a multiscalar multiplication $[n_1]P + [n_2]\psi(P)$ that is faster to compute ("GLV method"; see [this](https://www.iacr.org/archive/eurocrypt2009/54790519/54790519.pdf) or the [original paper](https://link.springer.com/content/pdf/10.1007/3-540-44647-8_11.pdf)).

[Grumpkin](https://github.com/AztecProtocol/aztec2-internal/blob/master/markdown/specs/AZTEC2-yellowpaper.md#2-grumpkin---a-curve-on-top-of-bn-254-for-snark-efficient-group-operations) is the curve $Y^2 = X^3 - 17$ defined over a field `grumpkin::fq`, meaning that a point of Grumpkin is a pair $(x,y)$ in `grumpkin::fq`$\times$`grumpkin::fq`. Such a $\beta\in$`grumpkin::fq` exists for Grumpkin. The fact that  `grumpkin::fq` is the same as `barretenberg::fr` (which should be called `bn254::fr`) is not relevant yet. Where it is relevant is the fact that we can efficiently represent elements of Grumpkin in circuits defined over BN254. By this I mean the selector and witness values of our circuits are all `barretenberg::fr`'s. So, this gate struct `ecc_add_gate gate{ x1, y1, x2, y2, x3, y3, beta_scalar, -1 };` points to a bunch of `barretenberg::fr`'s and this single, small bit of information, plus our special relation/widget, lets us record an elliptic curve addition over Grumpkin very efficiently. When we want to, say, try to describe scalar muls over SECP256k1 (which also has $a=0$), in a circuit over BN254, we can't do so as succinctly, since we need multiple `barretenberg::fr`'s to record a single coordinate of a point of secp256k1. Hope that helps, happy to clarify further as needed.

cpp/src/barretenberg/honk/sumcheck/relations/auxiliary_relation.hpp
ledwards2225: I've added some more documentation to this and a few other relations, mostly reproduced from their counterparts in Plonk. As we discussed earlier, I think going through and improving the docs in sumcheck/relations still warrants an entire Pr in and of itself

cpp/src/barretenberg/honk/sumcheck/relations/arithmetic_relation.hpp
ledwards2225: Good call - Added a Doxygen brief to this guy

cpp/src/barretenberg/honk/sumcheck/relations/ultra_arithmetic_relation.hpp
ledwards2225: Added some docs here

cpp/src/barretenberg/honk/sumcheck/sumcheck.test.cpp
maramihali: Thank you, it completely does but I overlooked this aspect.

maramihali (approved):
lgtm, we should maybe open an issue for additional documentation + renaming

PR #501, opened 2023-06-02
refactor!: Use circuit builders

### LINKED ISSUE #551: Reinstate ACIR format tests.

codygunton:
# Description

Finally finish the splitting work, up to some renaming for clarity to be done in a follow-up.

# Checklist:


cpp/src/barretenberg/grumpkin_srs_gen/grumpkin_srs_gen.cpp
```diff
@@ -7,6 +7,7 @@
 #include "barretenberg/common/net.hpp"
 
 const std::string protocol_name = "BARRETENBERG_GRUMPKIN_IPA_CRS";
+// WORKTODO: this is incorrect now right?
```
kevaundray: what is the difference between WORKTODO and a regular TODO?

cpp/src/barretenberg/stdlib/recursion/verifier/verifier.test.cpp
```diff
@@ -150,7 +153,7 @@ template <typename OuterComposer> class stdlib_verifier : public testing::Test {
             }
 
             info("Verifying the ultra (inner) proof natively...");
-            auto native_result = native_verifier.verify_proof(recursive_proof);
+            auto native_result = native_verifier.verify_proof(recursive_proof); // WORKTODO
```
kevaundray: what is the TODO referring to here

kevaundray (commented):
Still in draft so can ignore the nits.

Note: if `check_circuit` being false causes the prover to not create a proof, then this would be a breaking change since the observed behaviour from someone doing a recursive verification will be different when the circuit is invalid.

If only a flag is outputted or more generally, the observed behaviour does not change with or without `check_circuit` then it is non-breaking

cpp/src/barretenberg/grumpkin_srs_gen/grumpkin_srs_gen.cpp
codygunton: WORKTODO is something I search for and then resolve before putting the PR up for review.

cpp/src/barretenberg/stdlib/recursion/verifier/verifier.test.cpp
codygunton: `recursive_proof` is a bad name... misleading shorthand for `proof_to_be_recursively_verified`. Plan to rename to something easier to parse.

codygunton:
> Still in draft so can ignore the nits.
> 
> Note: if `check_circuit` being false causes the prover to not create a proof, then this would be a breaking change since the observed behaviour from someone doing a recursive verification will be different when the circuit is invalid.
> 
> If only a flag is outputted or more generally, the observed behaviour does not change with or without `check_circuit` then it is non-breaking

@kevaundray Did you see something that caused you to have a concern around this?

cpp/src/barretenberg/grumpkin_srs_gen/grumpkin_srs_gen.cpp
codygunton: Tbh maybe I should use something shorter like `HEY`

codygunton: Or perhaps `HMM` alla Zac

codygunton:
^ space needed :1st_place_medal:

kevaundray:
> > Still in draft so can ignore the nits.
> > Note: if `check_circuit` being false causes the prover to not create a proof, then this would be a breaking change since the observed behaviour from someone doing a recursive verification will be different when the circuit is invalid.
> > If only a flag is outputted or more generally, the observed behaviour does not change with or without `check_circuit` then it is non-breaking
> 
> @kevaundray Did you see something that caused you to have a concern around this?

Nope! I was modifying the PR titles and needed to check the contents to check if they were breaking or not

cpp/.clangd
```diff
@@ -59,8 +59,6 @@ Diagnostics:
       - readability-function-cognitive-complexity
       # It is often nicer to not be explicit
       - google-explicit-constructor
-    CheckOptions:
```
codygunton, line 62: I added this option before to avoid a warning we had, but now we don't need to avoid that warning and I added the option incorrectly anyway.

cpp/src/barretenberg/barretenberg.hpp
```diff
@@ -21,8 +21,8 @@
 #include "ecc/curves/grumpkin/grumpkin.hpp"
 #include "numeric/random/engine.hpp"
 #include "numeric/uint256/uint256.hpp"
-#include "plonk/composer/turbo_plonk_composer.hpp"
```
codygunton, line 24: The follow-up to this PR will include lots of renaming, eg "circuit constructor" becomes "circuit builder" in various places, "composer" becomes "circuit" or "builder" or "circuit_builder" or something in some places, and stays "composer" in other places.

cpp/src/barretenberg/benchmark/honk_bench/benchmark_utilities.hpp
```diff
@@ -41,11 +35,13 @@ struct BenchParams {
...
  * @param num_iterations
  */
-template <typename Composer> void generate_basic_arithmetic_circuit(Composer& composer, size_t num_gates)
+template <typename Builder> void generate_basic_arithmetic_circuit(Builder& builder, size_t num_gates)
 {
-    plonk::stdlib::field_t a(plonk::stdlib::witness_t(&composer, barretenberg::fr::random_element()));
-    plonk::stdlib::field_t b(plonk::stdlib::witness_t(&composer, barretenberg::fr::random_element()));
-    plonk::stdlib::field_t c(&composer);
+    proof_system::plonk::stdlib::field_t a(
+        proof_system::plonk::stdlib::witness_t(&builder, barretenberg::fr::random_element()));
```
codygunton, line 41: There was some `using namespace` pollution that let us get away with shortened namespace qualifiers.

cpp/src/barretenberg/barretenberg.hpp
codygunton: That said, there were certain places where I found it helpful/clarifying to do some renaming in order to get this refactor done, and I leave some of that work in place.

cpp/src/barretenberg/benchmark/honk_bench/benchmark_utilities.hpp
```diff
@@ -41,11 +35,13 @@ struct BenchParams {
  * @param composer
  * @param num_iterations
  */
-template <typename Composer> void generate_basic_arithmetic_circuit(Composer& composer, size_t num_gates)
+template <typename Builder> void generate_basic_arithmetic_circuit(Builder& builder, size_t num_gates)
```
codygunton, line 38: Is it better to be more verbose and use `CircuitBuilder`, at least for the type name?

cpp/src/barretenberg/benchmark/honk_bench/honk.bench.cpp
```diff
@@ -1,7 +1,6 @@
-#include "barretenberg/ecc/curves/bn254/fr.hpp"
 #include <benchmark/benchmark.h>
 #include <cstddef>
-#include "barretenberg/honk/composer/standard_honk_composer.hpp"
```
codygunton, line 4: Renaming and moving comes in the next PR.

cpp/src/barretenberg/benchmark/honk_bench/ultra_honk.bench.cpp
```diff
@@ -1,10 +1,16 @@
+#include <benchmark/benchmark.h>
+
+#include "barretenberg/proof_system/circuit_constructors/ultra_circuit_constructor.hpp"
```
codygunton, line 3: I believe this was necessary as part of the clean-up around `composers_fwd.hpp` etc.

cpp/src/barretenberg/common/log.hpp
```diff
@@ -1,34 +1,16 @@
...
 #include <algorithm>
 #include <vector>
 #include <string>
+#include "barretenberg/stdlib/primitives/circuit_builders/circuit_builders_fwd.hpp"
+#include "barretenberg/env/logstr.hpp"
 
 #define BENCHMARK_INFO_PREFIX "##BENCHMARK_INFO_PREFIX##"
 #define BENCHMARK_INFO_SEPARATOR "#"
 #define BENCHMARK_INFO_SUFFIX "##BENCHMARK_INFO_SUFFIX##"
-#define GET_COMPOSER_NAME_STRING(composer)                                                                             \
```
codygunton, line 11: This was confusing to reason about (for instance, this macro has been used incorrectly in the join-split tests for as long as we've had benchmarks there). The new model is: store a name string in each composer/circuit constructor. I'm hoping this can also replace the (much loathed by me) `ComposerType` enum.

cpp/src/barretenberg/common/log.hpp
```diff
@@ -38,7 +20,9 @@ template <typename... Args> std::string format(Args... args)
 
 template <typename T> void benchmark_format_chain(std::ostream& os, T const& first)
 {
-    // We will be saving these values to a CSV file, so we can't tolerate commas
+    // We will be saving these values to
```
codygunton: Idk why this was reformatted. First time this file has been touched in a while?

cpp/src/barretenberg/dsl/acir_format/acir_format.cpp
```diff
@@ -22,119 +22,109 @@ void create_circuit(Composer& composer, acir_format const& constraint_system)
...
 
-Composer create_circuit_with_witness(acir_format const& constraint_system,
-                                     WitnessVector const& witness,
-                                     std::shared_ptr<barretenberg::srs::factories::CrsFactory> const& crs_factory,
-                                     size_t size_hint)
+Builder create_circuit_with_witness(acir_format const& constraint_system,
+                                    WitnessVector const& witness,
+                                    size_t size_hint)
 {
-    Composer composer(crs_factory, size_hint);
```
codygunton, line 125: Before splitting, we could initialize a composer like this. Now the size hint is given to the circuit constructor and the crs factory is given to the composer (often by default, using the global object introduced by Charlie).

cpp/src/barretenberg/dsl/acir_format/acir_format.cpp
```diff
@@ -22,119 +22,109 @@ void create_circuit(Composer& composer, acir_format const& constraint_system)
...
-    create_circuit_with_witness(composer, constraint_system, witness);
-    return composer;
+    Builder builder(size_hint);
+    create_circuit_with_witness(builder, constraint_system, witness);
+    return builder;
 }
 
-Composer create_circuit_with_witness(const acir_format& constraint_system, WitnessVector const& witness)
```
codygunton, line 130: This function goes away because, when I removed the unneeded `crs_factory` argument in the previous function, it lead to a function signature class with this one. Turns out I could just delete this function.

cpp/src/barretenberg/dsl/acir_format/acir_format.cpp
```diff
@@ -22,119 +22,109 @@ void create_circuit(Composer& composer, acir_format const& constraint_system)
...
-    create_circuit(composer, constraint_system);
-    return composer;
+    Builder builder(size_hint);
+    create_circuit(builder, constraint_system);
+    return builder;
 }
 
-Composer create_circuit_with_witness(acir_format const& constraint_system,
-                                     WitnessVector const& witness,
-                                     std::shared_ptr<barretenberg::srs::factories::CrsFactory> const& crs_factory,
```
codygunton, line 122: CRS factory not needed by this function, as you could guess by the name of the function. I find these simplifications satisfying and a good sign overall.

codygunton: We see the `crs_factory`  going away in other places as well.

cpp/src/barretenberg/examples/simple/simple.hpp
```diff
@@ -7,12 +7,18 @@ namespace examples::simple {
 using namespace proof_system::plonk;
 using namespace stdlib::types;
 
-Composer* create_composer(std::shared_ptr<barretenberg::srs::factories::CrsFactory> const& crs_factory);
+struct BuilderComposerPtrs {
```
codygunton, line 10: I'm taking suggestions for making this less ugly--I'm lacking context I just did a quick job here.

cpp/src/barretenberg/join_split_example/proofs/compute_circuit_data.hpp
```diff
@@ -46,15 +48,18 @@ circuit_data get_circuit_data(std::string const& name,
                               bool pk,
                               bool vk,
                               bool padding,
-                              bool mock,
+                              bool mock, // TODO(#541)
```
codygunton, line 51: Refactoring this test with guidance might be a good first task for someone new #541

cpp/src/barretenberg/join_split_example/proofs/compute_circuit_data.hpp
```diff
@@ -68,29 +73,23 @@ circuit_data get_circuit_data(std::string const& name,
         (compute && !load)) {
         info(name, ": Building circuit...");
         Timer timer;
-        build_circuit(composer);
+        build_circuit(builder);
 
-        benchmark_collator.benchmark_info_deferred(GET_COMPOSER_NAME_STRING(proof_system::ComposerType),
```
codygunton, line 73: The macro was not being used correctly here, and was logging the results under the name `NullPlonk`.

cpp/src/barretenberg/join_split_example/proofs/join_split/join_split.test.cpp
```diff
@@ -15,108 +15,6 @@ namespace join_split_example::proofs::join_split {
 
 using namespace proof_system::plonk::stdlib::merkle_tree;
 
-template <typename Composer> class join_split : public testing::Test {
```
codygunton, line 18: I'm deleting this because it was never properly implemented (my fault)--functions like `join_split_circuit` would need to be properly templated, but they were not, they were just run repeatedly with the type specified in join_split_example/types.hpp.

cpp/src/barretenberg/join_split_example/proofs/join_split/join_split.test.cpp
```diff
@@ -1540,7 +1440,8 @@ TEST_F(join_split_tests,
     /**
      * tx represents:
      *   - 100 in (virtual)
-     *   - 110 in (virtual, different asset_id = virtual_asset_id + 1) <-- fails, since input note 2 value should equal
+     *   - 110 in (virtual, different asset_id = virtual_asset_id + 1) <-- fails, since input note 2 value should
```
codygunton: I don't know why the formatting changed here.

cpp/src/barretenberg/join_split_example/proofs/verify.hpp
```diff
@@ -63,86 +63,12 @@ auto verify_logic_internal(Composer& composer, Tx& tx, CircuitData const& cd, ch
...
-    result.public_inputs = composer.get_public_inputs();
+    result.public_inputs = builder.get_public_inputs();
     result.logic_verified = true;
-    result.number_of_gates = composer.get_num_gates();
+    result.number_of_gates = builder.get_num_gates();
 
     return result;
 }
 
-template <typename Composer, typename Tx, typename CircuitData, typename F>
```
codygunton, line 73: This function is not used anywhere.

cpp/src/barretenberg/plonk/composer/composer_helper/ultra_plonk_composer_helper.hpp
```diff
@@ -101,117 +103,7 @@ class UltraPlonkComposerHelper {
      */
     static transcript::Manifest create_manifest(const size_t num_public_inputs)
     {
-        // add public inputs....
```
codygunton, line 104: Moving this definition in the flavor ultimately helped me to avoid using composers in the Aztec circuits.

cpp/src/barretenberg/solidity_helpers/circuits/recursive_circuit.hpp
```diff
@@ -104,14 +106,14 @@ template <typename OuterComposer> class RecursiveCircuit {
     };
 
   public:
-    static OuterComposer generate(std::string srs_path, uint256_t inputs[])
+    static OuterComposer generate(uint256_t inputs[])
     {
-        InnerComposer inner_composer = InnerComposer(srs_path);
-        OuterComposer outer_composer = OuterComposer(srs_path);
+        InnerBuilder inner_circuit;
```
codygunton, line 111: I'm kind of in favor of calling circuits constructors `CircuitBuilder`s and then, depending on context, calling an instance `builder` or `circuit`. Seeking opinions.

cpp/src/barretenberg/stdlib/commitment/pedersen/pedersen.cpp
```diff
@@ -36,7 +36,8 @@ point<C> pedersen_commitment<C>::commit(const std::vector<field_t>& inputs,
         throw_or_abort("Vector size mismatch.");
     }
 
-    if constexpr (C::type == ComposerType::PLOOKUP && C::commitment_type == pedersen::CommitmentType::LOOKUP_PEDERSEN) {
+    if constexpr (C::type == proof_system::ComposerType::PLOOKUP &&
```
codygunton, line 39: IOU: get rid of `ComposerType` and introduce a concept like `UsePlookup` or something like this.

cpp/src/barretenberg/stdlib/hash/sha256/sha256.test.cpp
```diff
@@ -113,52 +114,24 @@ std::array<uint64_t, 8> inner_block(std::array<uint64_t, 64>& w)
     return output;
 }
 
-TEST(stdlib_sha256, test_duplicate_proving_key)
```
codygunton, line 116: After rewriting this (stdlib) test to not use a composer, it was clear that it's not a test of circuit logic, but instead of composer/proof system logic, so I deleted it.

cpp/src/barretenberg/stdlib/primitives/composers/composers.hpp
```diff
@@ -1,73 +0,0 @@
-/**
```
codygunton, line 1: (moved and pared down to one of the above files, similarly with composers_fwd.hpp)

cpp/src/barretenberg/stdlib/recursion/verifier/verifier.hpp
```diff
@@ -415,6 +413,23 @@ aggregation_state<Curve> verify_proof_(typename Curve::Composer* context,
     return result;
 }
 
+template <typename Flavor>
```
codygunton, line 416: I added this so that I could avoid needing to construct a manifest (a proof-system detail) in the the ACL. @maramihali in future recursion work, we should expose a nice and simple interface.

cpp/src/barretenberg/stdlib/recursion/verifier/verifier.test.cpp
```diff
@@ -1,70 +1,101 @@
 #include "verifier.hpp"
```
codygunton, line 1: It took a while to properly refactor this. Since the tests were already hard to read IMO, I went ahead and did a deeper refactor, aiming to save the next person some time.

codygunton (commented):
Leaving some explanatory notes.

cpp/src/barretenberg/benchmark/honk_bench/benchmark_utilities.hpp
ledwards2225: I think I prefer `CircuitBuilder` but I dont feel strongly. As long as we're not using `B`..

cpp/src/barretenberg/benchmark/honk_bench/benchmark_utilities.hpp
```diff
@@ -115,22 +111,23 @@ template <typename Composer> void generate_ecdsa_verification_test_circuit(Compo
 
         bool first_result =
             crypto::ecdsa::verify_signature<Sha256Hasher, fq, fr, g1>(message_string, account.public_key, signature);
+        static_cast<void>(first_result); // TODO(Cody): This is not used anywhere.
```
ledwards2225, line 114: I've started using `[[ maybe_unused ]] first_result = ...` instead, but I think in this case the entire `crypto::ecdsa::verify_signature` can be deleted. I think this was debugging code that I forgot to delete.

cpp/src/barretenberg/common/log.hpp
ledwards2225: Strange. Line breaks way too short

cpp/src/barretenberg/common/log.hpp
ledwards2225: lovely

cpp/src/barretenberg/dsl/acir_format/acir_format.cpp
ledwards2225: nice

cpp/src/barretenberg/examples/simple/simple.cpp
```diff
@@ -11,58 +11,61 @@ using namespace stdlib::types;
...
     info("proof validity: ", valid);
     return valid;
 }
 
-void delete_composer(Composer* composer)
+void delete_builder_and_composer(BuilderComposerPtrs pair)
 {
-    delete composer;
+    delete pair.builder;
+    delete pair.composer;
```
ledwards2225, line 68: This is silly. We define unique pointers, extract the raw pointer from them, then call delete. I dont think delete makes sense here anyway because theres's no dynamically allocated memory. If we just stick with the unique pointers the mem management is automatic. Anyway, this PR is not the place.

cpp/src/barretenberg/solidity_helpers/circuits/recursive_circuit.hpp
ledwards2225: I suppose the difficulty is that the `CircuitBuilder` contains both the methods that do the building and the data that defines the circuit. This isnt very appealing and I'm not suggesting it but logically just `Circuit` actually makes the most sense to me. Presumably this is what led you to wanting to call the instance `inner_circuit.` So perhaps your blend of `InnerBuilder inner_circuit` is just the thing..

cpp/src/barretenberg/stdlib/recursion/verifier/verifier.hpp
ledwards2225: Automatically nicer for Honk since no manifest..

cpp/src/barretenberg/stdlib/recursion/verifier/verifier.test.cpp
```diff
@@ -272,170 +267,166 @@ template <typename OuterComposer> class stdlib_verifier : public testing::Test {
...
-    static bool check_recursive_proof_public_inputs(OuterComposer& composer,
+    // /**
+    //  * @brief Check the correctness of the recursive proof public inputs
+    //  *
+    //  * @details Circuit constructors have no notion of SRS and any proof-related stuff except for the existence of
+    //  * recursive proof-specific public inputs, so we can't check the recursive proof fully in check_circuit. So we
+    //  * use this additional function to check that the recursive proof points work.
+    //  *
+    //  * @return boolean result
+    //  */
```
ledwards2225: uncommment this description?

cpp/src/barretenberg/stdlib/recursion/verifier/verifier.test.cpp
```diff
@@ -449,267 +440,163 @@ template <typename OuterComposer> class stdlib_verifier : public testing::Test {
...
-                      outer_composer.composer_helper.crs_factory_->get_verifier_crs()->get_precomputed_g2_lines()),
-                  true);
+        check_pairing(circuit_output);
+        check_recursive_verification_circuit(outer_circuit, true);
     }
 
-    // verifies a proof of a circuit that verifies one of two proofs. Test 'a' uses a proof over the first of the two
-    // variable circuits
+    // verifies a proof of a circuit that verifies one of two proofs. Test 'a' uses a proof over the first of the
+    // two variable circuits
```
ledwards2225, line 457: I find that this happens comment code is further commented out then formatted

cpp/src/barretenberg/stdlib/recursion/verifier/verifier_turbo.test.cpp
```diff
@@ -1,516 +1,517 @@
...
     TestFixture::test_recursive_proof_composition();
 };
 
 HEAVY_TYPED_TEST(stdlib_verifier_turbo, double_verification)
 {
-    TestFixture::test_double_verification();
+    if constexpr (std::same_as<TypeParam, TurboPlonkComposerHelper>) {
+        TestFixture::test_double_verification();
+    } else {
+        // Test doesn't compile.
```
ledwards2225, line 512: Is this expected? Could be worth elaborating here

ledwards2225 (approved):
Looks great. Big step forward. No major comments

cpp/src/barretenberg/stdlib/recursion/verifier/verifier.test.cpp
codygunton: Ah, good call.

cpp/src/barretenberg/stdlib/recursion/verifier/verifier_turbo.test.cpp
codygunton: Yeah, the comment I rewrote, which said  that the test was too heavy to run in CI, was misleading. It doesn't run because in the `else` branch there's a type clash due to how the test setup functions are written.

PR #512, opened 2023-06-07
fix:  Ecdsa Malleability Bug

suyash67:
# Description

ECDSA has an inherent malleability because of the way $s$ in an ECDSA signature is computed. Read more about this [here](https://www.derpturkey.com/inherent-malleability-of-ecdsa-signatures/amp/). This PR adds the constraint $s < |Fr|/2$ in ECDSA verification to fix the malleability. Additionally, in the signature construction, we make sure that $s < |Fr|/2$.

Acknowledgements: @Rumata888 corrected the constraint on $s$ in verification.

# Checklist:

### DIFF
--- a/cpp/src/barretenberg/stdlib/primitives/bigfield/bigfield.hpp
+++ b/cpp/src/barretenberg/stdlib/primitives/bigfield/bigfield.hpp
@@ -248,7 +248,7 @@ template <typename Composer, typename T> class bigfield {
     bigfield conditional_select(const bigfield& other, const bool_t<Composer>& predicate) const;
 
     void assert_is_in_field() const;
-    void assert_less_than(const uint256_t& upper_limit) const;
+    void assert_less_than(const uint256_t upper_limit) const;
     void assert_equal(const bigfield& other) const;
     void assert_is_not_equal(const bigfield& other) const;
 
--- a/cpp/src/barretenberg/stdlib/primitives/bigfield/bigfield.hpp
+++ b/cpp/src/barretenberg/stdlib/primitives/bigfield/bigfield.hpp
@@ -248,6 +248,7 @@ template <typename Composer, typename T> class bigfield {
     bigfield conditional_select(const bigfield& other, const bool_t<Composer>& predicate) const;
 
     void assert_is_in_field() const;
+    void assert_less_than(const uint256_t& upper_limit) const;
     void assert_equal(const bigfield& other) const;
     void assert_is_not_equal(const bigfield& other) const;
 
--- a/cpp/src/barretenberg/stdlib/primitives/bigfield/bigfield.test.cpp
+++ b/cpp/src/barretenberg/stdlib/primitives/bigfield/bigfield.test.cpp
@@ -708,6 +708,42 @@ template <typename Composer> class stdlib_bigfield : public testing::Test {
         EXPECT_EQ(result, true);
     }
 
+    static void test_assert_less_than_success()
+    {
+        auto composer = Composer();
+        size_t num_repetitions = 10;
+        constexpr size_t num_bits = 200;
+        constexpr uint256_t bit_mask = (uint256_t(1) << num_bits) - 1;
+        for (size_t i = 0; i < num_repetitions; ++i) {
+
+            fq inputs[4]{ uint256_t(fq::random_element()) && bit_mask,
+                          uint256_t(fq::random_element()) && bit_mask,
+                          uint256_t(fq::random_element()) && bit_mask,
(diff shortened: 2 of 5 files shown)


cpp/src/barretenberg/stdlib/primitives/bigfield/bigfield_impl.hpp
```diff
@@ -1664,6 +1664,67 @@ template <typename C, typename T> void bigfield<C, T>::assert_is_in_field() cons
     }
 }
 
+template <typename C, typename T> void bigfield<C, T>::assert_less_than(const uint256_t& upper_limit) const
```
Rumata888: Please create a TODO to merge it with assert_is_in_field in the future

cpp/src/barretenberg/stdlib/primitives/bigfield/bigfield.test.cpp
```diff
@@ -712,6 +712,42 @@ template <typename Composer> class stdlib_bigfield : public testing::Test {
         EXPECT_EQ(result, true);
     }
 
+    static void test_assert_less_than_success()
```
Rumata888, line 715: Please create a test that checks specifically for the edge case a<a. I think that the assert_less_than function will allow it, since we do a-a and check that it can be range constrained.

cpp/src/barretenberg/stdlib/primitives/bigfield/bigfield.hpp
```diff
@@ -248,6 +248,7 @@ template <typename Composer, typename T> class bigfield {
     bigfield conditional_select(const bigfield& other, const bool_t<Composer>& predicate) const;
 
     void assert_is_in_field() const;
+    void assert_less_than(const uint256_t& upper_limit) const;
```
Rumata888: Why should it be a reference?

Rumata888 (approved):
I've made two changes:
1) The function was actually less or equal.
2) Once I changed it to be less_than, the logic for constraining s became incorrect. You were checking that s < (Fr::modulus / 2). It is different from 2*s < Fr::modulus, because Fr::modulus is odd. Think  mod 3. If you check that something is less than 3/2, then the only option is zero, while 1 should also be a legitimate value.

cpp/src/barretenberg/stdlib/primitives/bigfield/bigfield.test.cpp
```diff
@@ -712,6 +712,53 @@ template <typename Composer> class stdlib_bigfield : public testing::Test {
...
+            // fq_ct c = a + a + a + a - b - b - b - b;
+            c.assert_less_than(bit_mask + 1);
+            uint256_t result = (c.get_value().lo);
+            EXPECT_EQ(result, uint256_t(expected));
+            EXPECT_EQ(c.get_value().get_msb() < num_bits, true);
+        }
+        bool result = composer.check_circuit();
+        EXPECT_EQ(result, true);
+        // Checking edge conditions
+        fq random_input = fq::random_element();
```
kevaundray, line 746: nit, but not a blocker: unit tests should ideally not contain random elements as they then become unreproducable

kevaundray (approved):
Looks good to me

cpp/src/barretenberg/stdlib/primitives/bigfield/bigfield.test.cpp
Rumata888: You are partially right. I'd add something like "random_element_for_test" into the field, so that it would automatically print it. But it is actually better to have random elements than static, because if there is a bug there is a chance that it triggers. And then you can stress the test to find it (that was the case with construct_addition_chains)

kevaundray:
Since this modifies the stdlib by adding extra constraints, I'm marking this a breaking change

kevaundray:
Update: It was decided that changes to the stdlib will not be marked as breaking changes as this would include most commits

suyash67:
> I've made two changes:
> 
> 1. The function was actually less or equal.
> 2. Once I changed it to be less_than, the logic for constraining s became incorrect. You were checking that s < (Fr::modulus / 2). It is different from 2*s < Fr::modulus, because Fr::modulus is odd. Think  mod 3. If you check that something is less than 3/2, then the only option is zero, while 1 should also be a legitimate value.

Nice catch, thanks Kesha!

cpp/src/barretenberg/stdlib/primitives/bigfield/bigfield.hpp
suyash67: It doesn't need to be, you're right. Removing it.

PR #514, opened 2023-06-07
feat(honk): Shared relation arithmetic

### LINKED ISSUE #481: Share Relation arithmetic between Prover/Verifier
The identity arithmetic in the relations is currently duplicated for the Prover and Verifier. This was done consciously to allow for the two to be optimized independently but it has become clear that the benefits of removing the code duplication outweigh the costs.

ledwards2225:
# Description

Introduces a framework that allows the relation arithmetic to be shared between the prover and verifier.

The sumcheck prover and verifier accumulate the contributions from each relation (really, each sub-relation) into, respectively, Univariates and individual field elements. When performing relation arithmetic on Univariates, we introduce UnivariateViews to reduce full length Univariates to the minimum required length and to avoid unnecessary copies. 

To share the relation arithmetic. we introduce simple structs that specify two types: `Accumulators` and `AccumulatorViews`. For the prover, who accumulates Univariates, these are respectively `std::tuple<Univariate>` and `std::tuple<UnivariateView>`. For the verifier, who accumulates `FF`s, both types are simply aliases for `std::array<FF>` (since no "view" type is necessary). The containers `std::tuple` and `std::array` are needed to accommodate multiple sub-relations within each relation, where, for efficiency, each sub-relation has its own specified degree.

# Checklist:

### DIFF
--- a/cpp/src/barretenberg/proof_system/flavor/flavor.hpp
+++ b/cpp/src/barretenberg/proof_system/flavor/flavor.hpp
@@ -269,23 +269,6 @@ template <class FF, size_t ExtendedLength, std::size_t Length = 2> static conste
     }
 }
 
-/**
- * @brief Recursive helper function to construct BarycentricData to extend each Relation in a tuple
- *
- */
-template <class FF, typename Tuple, size_t ExtendedLength, std::size_t Index = 0>
-static constexpr auto create_barycentric_utils()
-{
-    if constexpr (Index >= std::tuple_size<Tuple>::value) {
-        return std::tuple<>{}; // Return empty when reach end of the tuple
-    } else {
-        constexpr size_t relation_length = std::tuple_element_t<Index, Tuple>::RELATION_LENGTH;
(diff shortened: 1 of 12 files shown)


cpp/src/barretenberg/proof_system/flavor/flavor.hpp
```diff
@@ -275,23 +275,6 @@ template <class FF, size_t ExtendedLength, std::size_t Length = 2> static conste
     }
 }
 
-/**
```
ledwards2225, line 272: Note: this was an earlier iteration of the method `instantiate_barycentric_utils`, defined just above, which is the one actually being used. This was never meant to go in and this is the first I'm noticing it so I'm deleting it in this unrelated PR.

kevaundray:
Since this only modifies honk related code -- marking this feature as non-breaking while the honk codepath not known to be used externally

zac-williamson (approved):

PR #524, opened 2023-06-11
feat: Initial native version of bb binary.

charlielye:
# Description

Just as bb.js provides an architecture agnostic cli prover/verifier, we probably want a native build that follows the same cli args for when we can target the native arch. This overcomes the 512k gate limit and is probably about 5x faster.

This is an initial implementation of the native `bb` executable.

# Todo:

* We currently just assume a max circuit size of 512k. We can be a bit smarter and determine the number of gates beforehand and adjust accordingly. We do this in bb.js and cache the result in the json file (as it takes a few seconds). That's non trivial in C++ due to no direct JSON support, so I'll probably make it cache the gate count in a file. bb.js could do the same if we wanted "compatibility" between them. But it's also fine perhaps for backends to just create an e.g. `.bb.js` and `.bb` folder where they store their backend specific info.
* We currently assume we're running in a unix like environment with `awk` `gunzip` `base64` and `curl`. This should serve linux and mac fine, and probably WSL.
* We don't print any useful `usage` info.
* Maxim added the "proof_as_fields" and "vk_as_fields" commands to help with recursion. I've reimplemented these here, but we should probably just make proof and vk generation write these out at the same time? Or something better?


cpp/src/barretenberg/bb/get_bytecode.hpp
```diff
@@ -0,0 +1,13 @@
...
+#include "exec_pipe.hpp"
+
+/**
+ * We can assume for now we're running on a unix like system and use the following to extract the bytecode.
+ * Maybe we should consider bytecode being output into its own independent file alongside the JSON?
+ */
+inline std::vector<uint8_t> get_bytecode(const std::string& jsonPath)
+{
+    std::string command =
+        "awk -F'\"bytecode\":' '{print $2}' " + jsonPath + " | awk -F'\"' '{print $2}' | base64 -d | gunzip";
```
kobyhallx, line 11: This essentially stops us executing on Windows ...

cpp/src/barretenberg/bb/get_crs.hpp
```diff
@@ -0,0 +1,84 @@
...
+#include <ios>
+#include <barretenberg/ecc/curves/bn254/g1.hpp>
+#include <barretenberg/srs/io.hpp>
+
+inline std::vector<uint8_t> download_g1_data(size_t num_points)
+{
+    size_t g1_start = 28;
+    size_t g1_end = g1_start + num_points * 64 - 1;
+    std::string command = "curl -s -H \"Range: bytes=" + std::to_string(g1_start) + "-" + std::to_string(g1_end) +
+                          "\" 'https://aztec-ignition.s3.amazonaws.com/MAIN%20IGNITION/monomial/transcript00.dat'";
```
kobyhallx, line 16: Would we need to consider transcript01.dat..transcript02.dat as well ?

cpp/src/barretenberg/bb/main.cpp
```diff
@@ -0,0 +1,183 @@
...
+    } else if (command == "proof_as_fields") {
+        std::string output_path = getOption(args, "-o", proof_path + "_fields.json");
+        proofAsFields(proof_path, vk_path, output_path);
+    } else if (command == "vk_as_fields") {
+        std::string output_path = getOption(args, "-o", vk_path + "_fields.json");
+        vkAsFields(vk_path, output_path);
+    } else {
+        std::cerr << "Unknown command: " << command << "\n";
+        return -1;
+    }
```
kobyhallx, line 182: these are essentially public API's wonder if they could get abstracted away so they would be callable from wasm ...

cpp/src/barretenberg/bb/get_bytecode.hpp
charlielye: Not on WSL? Current thinking is on windows we just target WSL, but yes I suppose if we want full native builds we need a mingw cross-compile and to rethink the above. Not sure if worth it if most devs have WSL installed.

cpp/src/barretenberg/bb/get_crs.hpp
charlielye: Good call, this won't work beyond 2^19 for now, but we've set max circuit size for now even for this binary as 2^19. Just a first pass at getting something in.

cpp/src/barretenberg/bb/main.cpp
charlielye: Not sure what you mean by callable from wasm. This binary is not meant to be used in/from any wasm context?

PR #531, opened 2023-06-14
feat(ultrahonk): Added a simple filler table to minimize the amount of entries used to make UltraHonk polynomials non-zero

### LINKED ISSUE #423: Avoid zero-polynomials in Ultra Honk
The approach taken in Standard Honk to avoid zero polynomials was to add a single valid arithmetic gate by default in the constructor of the standard circuit constructor.

The increased complexity of the relations in Ultra Honk makes this approach less straightforward. The biggest issue is that a lookup must be performed to populate various lookup related polys with non-zero values. This automatically increases the circuit size by at least the size of the corresponding table. Currently the approach (encapsulated in `add_gates_to_ensure_all_polys_are_non_zero()`) increases the circuit size by 16k and is obviously untenable. 

One solution would be to add a dummy table with one row and perform a lookup on that. We should also investigate whether there is a way to simply allow zero polynomials and handle the resulting commitments accordingly.

Relatedly, for Ultra Honk we have the additional requirement that all polynomials for which we require a shift must have constant coefficient equal to zero. The current approach to doing this is to add a valid gate with all four wires equal to zero. This only works if there are no public inputs, however, since PI are always placed at the start of the witness polys. A universal solution may be to simple add 0 as a PI by default.

Rumata888:
# Description
In the past we used UINT32 XOR and ADD lookup tables to ensure that UltraHonk! polynomials are not zero. This made the minimum size of the circuit 2^14. This PR adds a dummy table which can be used for the same purpose but only has 2 basic tables each consisting of just 4 entries. In addition I got rid of s_randomness in UltraHonkComposerHelper, which produced bugs, since we don't randomise the polynomials in the same way in Honk!
# Checklist:


cpp/src/barretenberg/honk/composer/composer_helper/ultra_honk_composer_helper.cpp
```diff
@@ -230,7 +220,7 @@ std::shared_ptr<typename Flavor::ProvingKey> UltraHonkComposerHelper_<Flavor>::c
     polynomial poly_q_table_column_3(subgroup_size);
     polynomial poly_q_table_column_4(subgroup_size);
 
-    size_t offset = subgroup_size - tables_size - s_randomness - 1;
+    size_t offset = subgroup_size - tables_size;
```
maramihali, line 223: why was the -1 for previously?

cpp/src/barretenberg/proof_system/circuit_constructors/ultra_circuit_constructor.cpp
```diff
@@ -100,23 +100,17 @@ void UltraCircuitConstructor::add_gates_to_ensure_all_polys_are_non_zero()
     // to get a non-zero value in table_4. I assume this index is arbitrary and could
     // start from 1 instead of 0?
     uint32_t left_value = 3;
-    uint32_t right_value = 5;
+    uint32_t right_value = 3;
```
maramihali, line 103: can you explain what this value is because the change looks arbitrary - maybe add comment as well

cpp/src/barretenberg/proof_system/plookup_tables/dummy.hpp
```diff
@@ -0,0 +1,95 @@
...
+ *
+ */
+
+#include "types.hpp"
+
+namespace plookup {
+namespace dummy_tables {
+
+/**
+ * @brief Lookup the value corresponding to a sepcific key
```
maramihali: typo in specific

cpp/src/barretenberg/proof_system/plookup_tables/dummy.hpp
```diff
@@ -0,0 +1,95 @@
...
+ * @tparam table_id The id of the table this function is instantiated for
+ * @param id Table id that is the same for all circuits
+ * @param table_index The index for this table that is used in this circuit. 0, 1, ...
+ * @return A table of values
+ */
+template <uint64_t table_id>
+inline BasicTable generate_honk_dummy_table(const BasicTableId id, const size_t table_index)
+{
+
+    ASSERT(table_id == static_cast<uint64_t>(id));
```
maramihali, line 48: why do this two have to be equal - i understand that `id` is static while `table_id` can change depending on function (?)

cpp/src/barretenberg/honk/composer/composer_helper/ultra_honk_composer_helper.cpp
Rumata888: We inserted a 1 at the end so that the polynomial would not be zero. Don't know why since we add randomness anyway

cpp/src/barretenberg/proof_system/circuit_constructors/ultra_circuit_constructor.cpp
Rumata888: The change is because the multitable only works on 2-bit values, so 5 is too big

cpp/src/barretenberg/proof_system/plookup_tables/dummy.hpp
Rumata888: Thanks

cpp/src/barretenberg/proof_system/plookup_tables/dummy.hpp
Rumata888: We are using the same function in 2 places by templating it, but the API already includes the id in arguments. This is just a simple bug check to ensure that we don't use the wrong instantiation

PR #534, opened 2023-06-15
feat: Make the circuit constructors field agnostic so we can check circuits on grumpkin

maramihali:
This PR continues the work of getting Honk over Grumpkin by making the circuit constructors and gates field agnostic and ensuring the `StandardCircuitConstructor` is able to verify basic circuits over Grumpkin.

# Checklist:


cpp/src/barretenberg/honk/flavor/standard.hpp
```diff
@@ -32,7 +33,7 @@ namespace proof_system::honk::flavor {
  */
 class Standard {
   public:
-    using CircuitConstructor = StandardCircuitConstructor;
+    using CircuitConstructor = StandardCiruitConstructor;
```
kevaundray:
```suggestion
    using CircuitConstructor = StandardCircuitConstructor;
```

maramihali:
I reduced scope to only having standard circuits on Grumpkin because ultra circuits require field-agnostic plookup components and I want to prioritise getting Standard Honk on Grumpkin before moving to Ultra (which should be overall similar but would likely require more components in bberg to be templated)

cpp/src/barretenberg/ecc/curves/bn254/bn254.hpp
```diff
@@ -15,5 +16,6 @@ class BN254 {
     using AffineElement = typename Group::affine_element;
     using G2AffineElement = typename barretenberg::g2::affine_element;
     using G2BaseField = typename barretenberg::fq2;
+    using Fq12 = barretenberg::fq12;
```
codygunton: Could you give this a more generic name like `PairingTargetField` or something like this? That way, if we have end up implementing another pairing-friendly curve with possibly a different pairing extension degree (not 12), we will not have to change aliases. It's also more readable for someone looking at the library for the first time. It might also more future-proof to not refer to the name of the prime in the alias.

cpp/src/barretenberg/proof_system/arithmetization/gate_data.hpp
```diff
@@ -4,63 +4,70 @@
...
+    FF q_o;
+    FF q_c;
 
-    friend bool operator==(poly_triple const& lhs, poly_triple const& rhs) = default;
+    friend bool operator==(poly_triple_<FF> const& lhs, poly_triple_<FF> const& rhs) = default;
 };
 
+using poly_triple = poly_triple_<barretenberg::fr>;
+
+// TODO: figure out what to do with this...
```
codygunton: Unresolved TODO. If you conclude this does not need to be handled now, please make a github issue and update the comment to read `// TODO(#<issue number>): ...`.

cpp/src/barretenberg/proof_system/arithmetization/gate_data.hpp
```diff
@@ -73,7 +80,7 @@ template <typename B> inline void read(B& buf, poly_triple& constraint)
     read(buf, constraint.q_o);
     read(buf, constraint.q_c);
 }
-
+// TODO: and this..
```
codygunton: Ditto

cpp/src/barretenberg/proof_system/arithmetization/gate_data.hpp
```diff
@@ -4,63 +4,70 @@
...
-    barretenberg::fr a_scaling;
-    barretenberg::fr b_scaling;
-    barretenberg::fr c_scaling;
-    barretenberg::fr const_scaling;
+    FF a_scaling;
+    FF b_scaling;
+    FF c_scaling;
+    FF const_scaling;
 };
+using add_triple = add_triple_<barretenberg::fr>;
```
codygunton, line 19: Doing this aliaising is ok for now, but probably we'll want an alternative method later so that we don't have `add_triple_grumpkin` etc. Can discuss more when appropriate. Is this just a convenience to avoid having to add an explicit template parameter elsewhere?

codygunton: FYI we can probably avoid a lot of boilerplate by turning lines like
`ctx->create_foo_gate(foo_triple{x,...,y,...})`
into lines like
`ctx->create_foo_gate({x,...,y,...})`
IMO explicitly initializing the input struct doesn't add clarity, whereas I do think using the explicit syntax
`ctx->create_foo_gate({.w_1=x,...,.q_m=y,...})`
does help (but I'm not asking for that!).

cpp/src/barretenberg/proof_system/circuit_constructors/turbo_circuit_constructor.test.cpp
```diff
@@ -187,8 +187,8 @@ TEST(turbo_circuit_constructor, small_scalar_multipliers)
     bool skew = false;
     barretenberg::wnaf::fixed_wnaf<num_wnaf_bits, 1, 2>(&scalar_multiplier_base.data[0], &wnaf_entries[0], skew, 0);
 
-    fr accumulator_offset = (fr::one() + fr::one()).pow(static_cast<uint64_t>(initial_exponent)).invert();
-    fr origin_accumulators[2]{ fr::one(), accumulator_offset + fr::one() };
+    fr accumulator_ofrset = (fr::one() + fr::one()).pow(static_cast<uint64_t>(initial_exponent)).invert();
```
codygunton: Find and replate pollution: `ofrset` ~> `offset` in this file

cpp/src/barretenberg/proof_system/circuit_constructors/ultra_circuit_constructor.test.cpp
```diff
@@ -15,7 +15,7 @@ TEST(ultra_circuit_constructor, create_gates_from_plookup_accumulators)
 {
     UltraCircuitConstructor circuit_constructor = UltraCircuitConstructor();
 
-    barretenberg::fr input_value = fr::random_element();
+    fr input_value = fr::random_element();
```
codygunton: Doesn't this make the file a little less clear? We'll want to find and replace `barretenberg::fr` to `bn254::fr` at some point in the future, and so I'd rather leave this as it was.

cpp/src/barretenberg/proof_system/circuit_constructors/ultra_circuit_constructor.cpp
```diff
@@ -30,14 +30,15 @@ void UltraCircuitConstructor::finalize_circuit()
      *  1. ComposerBase::n => refers to the size of the witness of a given program,
      *  2. proving_key::n => the next power of two ≥ total witness size.
      *
-     * In this case, we have composer.num_gates = n_computation before we execute the following two functions.
+     * In this case, we have composer.this->num_gates = n_computation before we execute the following two functions.
```
codygunton: Find-and-replace polluting comments.

cpp/src/barretenberg/proof_system/circuit_constructors/ultra_circuit_constructor.cpp
```diff
@@ -284,11 +286,11 @@ void UltraCircuitConstructor::create_balanced_add_gate(const add_quad& in)
     q_lookup_type.emplace_back(0);
     q_elliptic.emplace_back(0);
     q_aux.emplace_back(0);
-    ++num_gates;
+    ++this->num_gates;
     // Why 3? TODO: return to this
     // The purpose of this gate is to do enable lazy 32-bit addition.
     // Consider a + b = c mod 2^32
-    // We want the 4th wire to represent the quotient:
+    // We want the this->4th wire this->to represent this->the quotient:
```
codygunton: Ditto here and elsewhere in this file.

codygunton (changes requested):
Nice work, thanks. I'm only asking for cleanup and conversations.

cpp/src/barretenberg/ecc/curves/bn254/bn254.hpp
maramihali: Dyt `TargetField` is enough? You won't have a target field unless the curve supports pairings.

cpp/src/barretenberg/proof_system/circuit_constructors/ultra_circuit_constructor.test.cpp
maramihali: Makes sense, this change was trigger by the fact that there's fr almost everywhere and then four places with barretenberg::fr and we are `using` the namespace. Should I change all places to `barretenberg::fr `?

cpp/src/barretenberg/proof_system/circuit_constructors/ultra_circuit_constructor.cpp
maramihali: wops vscode did something funny here, thanks for spotting

cpp/src/barretenberg/proof_system/arithmetization/gate_data.hpp
maramihali: I cleaned some boilerplatecode as you suggested, thanks for that - my idea is that this is sort of okay to get to MVP with Grumpkin on StandardHonk without much hassle in the rest of the codebase but then should be cleaned up (opened issue 557 for this file); the nicest thing to do imo is have no alias and just make the type for each gate explicit when we refer to them in the codebase. They don't appear in that many places after i did a bit of cleanup.

cpp/src/barretenberg/proof_system/arithmetization/gate_data.hpp
maramihali: I cleaned some boilerplatecode as you suggested, thanks for that - my idea is that this is sort of okay to get to MVP with Grumpkin on StandardHonk without much hassle in the rest of the codebase but then should be cleaned up (opened issue , including the read part); the nicest thing to do imo is not have an alias and just make the type for each gates explicit when we refer to them in the codebase. They don't appear in that many places

maramihali:
Opens issue #557 for addressing TODOs with templating gates.

cpp/src/barretenberg/proof_system/arithmetization/gate_data.hpp
codygunton: #557

codygunton (approved):

PR #540, opened 2023-06-16
fix: Revert "fix: add NUM_RESERVED_GATES before fetching subgroup size in composer"

vezenovm:
Reverts AztecProtocol/barretenberg#539


ludamad:
For posterity: This was approved verbally

PR #547, opened 2023-06-20
chore:  Random polynomial is now applied on a per-relation term by the Prover

zac-williamson:
# Description

The random polynomial is now applied on a per-relation term by the Prover (instead of being applied to all relations by the Verifier).

Required if we want relations that sum over the hypercube (i.e. we check the sum of the relation terms at each row are zero, instead of checking the relation term is zero at each row).

This is needed to evaluate logarithmic derivatives.

# Checklist:

- [x] The branch has been merged with/rebased against the head of its merge target.
- [x] No superfluous `include` directives have been added.
- [x] I have [linked](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) to any issue(s) it resolves.


cpp/src/barretenberg/honk/flavor/standard.hpp
```diff
@@ -56,6 +56,7 @@ class Standard {
     using Relations = std::tuple<sumcheck::ArithmeticRelation<FF>, sumcheck::PermutationRelation<FF>>;
 
     static constexpr size_t MAX_RELATION_LENGTH = get_max_relation_length<Relations>();
+    static constexpr size_t MAX_RANDOM_RELATION_LENGTH = MAX_RELATION_LENGTH + 1;
     static constexpr size_t NUM_RELATIONS = std::tuple_size<Relations>::value;
 
     // Instantiate the BarycentricData needed to extend each Relation Univariate
```
ledwards2225, line 65: I think this should now use `MAX_RANDOM_RELATION_LENGTH` instead of `MAX_RELATION_LENGTH` to ensure the correct static arrays in barycentric data are computed at compile time

cpp/src/barretenberg/honk/flavor/standard.hpp
```diff
@@ -56,6 +56,7 @@ class Standard {
     using Relations = std::tuple<sumcheck::ArithmeticRelation<FF>, sumcheck::PermutationRelation<FF>>;
 
     static constexpr size_t MAX_RELATION_LENGTH = get_max_relation_length<Relations>();
+    static constexpr size_t MAX_RANDOM_RELATION_LENGTH = MAX_RELATION_LENGTH + 1;
```
ledwards2225, line 62: A comment explaining that this is length after multiplying by the pow zeta univariate would be helpful here since `MAX_RANDOM_RELATION_LENGTH` does not immediately make it clear what this is

cpp/src/barretenberg/honk/sumcheck/sumcheck_round.hpp
```diff
@@ -285,13 +289,31 @@ template <typename Flavor> class SumcheckRound {
...
-    static void extend_and_batch_univariates(auto& tuple, Univariate<FF, extended_size>& result)
+    static void extend_and_batch_univariates(auto& tuple,
+                                             const PowUnivariate<FF>& pow_univariate,
+                                             Univariate<FF, extended_size>& result)
     {
-        auto extend_and_sum = [&](auto& element) {
-            using Element = std::remove_reference_t<decltype(element)>;
+        // Random poly R(X) = (1-X) + X.zeta_pow
+        auto random_poly_edge = Univariate<FF, 2>({ 1, pow_univariate.zeta_pow });
+        BarycentricData<FF, 2, extended_size> barycentric_2_to_max_2 = BarycentricData<FF, 2, extended_size>();
```
ledwards2225: Minor nit: At first I thought `barycentric_2_to_max_2` was a typo. I see why you did it now but I think something generic like `edge_extender` or `pow_extender` would be fine. The original `barycentric_2_to_max` is also not great since there are two max floating around now

ledwards2225 (approved):
Looks good to me (once the build errors are fixed, of course). Presumably you're thinking it will be necessary to have subrelations within a single relation that have different requirements re: linear independence? Maybe this will be the case when employing the log deriv if we maintain the same check for "shiftability". (Seems like there's got to be a better way to do that..). Anyway, would make things a bit nicer if the linear indep label could be applied at the Relation level but if this isn't the case then so be it

cpp/src/barretenberg/honk/sumcheck/sumcheck_round.hpp
zac-williamson: fixed, renamed to something more descriptive

cpp/src/barretenberg/honk/flavor/standard.hpp
zac-williamson: added comment

cpp/src/barretenberg/honk/flavor/standard.hpp
zac-williamson: instantiate_barycentric_utils now uses MAX_RANDOM_RELATION_LENGTH

PR #550, opened 2023-06-21
feat: Parallelised folding in Gemini

Rumata888:
# Description

The folding operation in Gemini now used parallelism. I also split gemini.hpp into gemini.cpp and gemini.hpp for faster compilation.

# Checklist:

### DIFF
--- a/cpp/src/barretenberg/honk/pcs/gemini/gemini.cpp
+++ b/cpp/src/barretenberg/honk/pcs/gemini/gemini.cpp
@@ -1,0 +1,409 @@
+
+#include "gemini.hpp"
+#include "barretenberg/common/thread.hpp"
+
+#include <memory>
+#include <vector>
+#include <bit>
+
+/**
+ * @brief Protocol for opening several multi-linear polynomials at the same point.
+ *
+ *
+ * m = number of variables
+ * n = 2ᵐ
+ * u = (u₀,...,uₘ₋₁)
+ * f₀, …, fₖ₋₁ = multilinear polynomials,
+ * g₀, …, gₕ₋₁ = shifted multilinear polynomial,
+ *  Each gⱼ is the left-shift of some f↺ᵢ, and gⱼ points to the same memory location as fᵢ.
+ * v₀, …, vₖ₋₁, v↺₀, …, v↺ₕ₋₁ = multilinear evalutions s.t. fⱼ(u) = vⱼ, and gⱼ(u) = f↺ⱼ(u) = v↺ⱼ
+ *
+ * We use a challenge ρ to create a random linear combination of all fⱼ,
+ * and actually define A₀ = F + G↺, where
+ *   F  = ∑ⱼ ρʲ fⱼ
+ *   G  = ∑ⱼ ρᵏ⁺ʲ gⱼ,
+ *   G↺ = is the shift of G
+ * where fⱼ is normal, and gⱼ is shifted.
+ * The evaluations are also batched, and
+ *   v  = ∑ ρʲ⋅vⱼ + ∑ ρᵏ⁺ʲ⋅v↺ⱼ = F(u) + G↺(u)
+ *
+ * The prover then creates the folded polynomials A₀, ..., Aₘ₋₁,
+ * and opens them at different points, as univariates.
+ *
+ * We open A₀ as univariate at r and -r.
+ * Since A₀ = F + G↺, but the verifier only has commitments to the gⱼs,
+ * we need to partially evaluate A₀ at both evaluation points.
+ * As univariate, we have
+ *  A₀(X) = F(X) + G↺(X) = F(X) + G(X)/X
+ * So we define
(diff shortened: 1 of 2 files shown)


ledwards2225:
I'm still digging into the details but I have to admit I'm finding the code quite hard to read. I think you said that you originally tried simply parallelizing each round of folding but that the gain was not significant. It seems like your current approach is to split portions of the d-many folding rounds across individual threads. Is this a significant improvement over the first approach? I ask because as far as I can tell, the first approach would literally be a 1 line change. If the difference is not significant, perhaps the added complexity is not worth it..

cpp/src/barretenberg/honk/pcs/gemini/gemini.cpp
```diff
@@ -0,0 +1,365 @@
...
+                                      Polynomial&& batched_to_be_shifted)
+{
+
+    using Fr = typename Params::Fr;
+    using Polynomial = barretenberg::Polynomial<Fr>;
+
+    const size_t num_variables = mle_opening_point.size(); // m
+
+    const size_t num_threads = get_num_cpus_pow2();
+    constexpr size_t efficient_operations_per_thread = 64; // A guess of the number of operation for which there
```
ledwards2225, line 68: We can always play with this later but I'm just curious whether this is motivated by some experimentation or if its really just a guess

cpp/src/barretenberg/honk/pcs/gemini/gemini.cpp
```diff
@@ -0,0 +1,365 @@
...
+        size_t chunk_size = n_l / num_used_threads;
+        size_t last_chunk_size = (n_l % chunk_size) ? (n_l % num_used_threads) : chunk_size;
+
+        // Openning point is the same for all
+        const Fr u_l = mle_opening_point[l];
+
+        // A_l_fold = Aₗ₊₁(X) = (1-uₗ)⋅even(Aₗ)(X) + uₗ⋅odd(Aₗ)(X)
+        auto A_l_fold = fold_polynomials[l + offset_to_folded].data();
+
+        parallel_for(num_used_threads, [&](size_t i) {
```
ledwards2225, line 117: As far as I can tell, `parallel_for` will determine the number of threads to use via openMP and essentially just ignores the input `num_used_threads`. Maybe this is fine but I wonder if we should update `parallel_for` to call `omp_set_num_threads()` so the num threads that is input is actually enforced.

cpp/src/barretenberg/honk/pcs/gemini/gemini.cpp
```diff
@@ -0,0 +1,365 @@
...
+    for (size_t l = 0; l < num_variables - 1; ++l) {
+        // size of the previous polynomial/2
+        const size_t n_l = 1 << (num_variables - l - 1);
+
+        // Use as many threads as it is useful so that 1 thread doesn't process 1 element, but make sure that there is
+        // at least 1
+        size_t num_used_threads = std::min(n_l / efficient_operations_per_thread, num_threads);
+        num_used_threads = num_used_threads ? num_used_threads : 1;
+        size_t chunk_size = n_l / num_used_threads;
+        size_t last_chunk_size = (n_l % chunk_size) ? (n_l % num_used_threads) : chunk_size;
```
ledwards2225, line 109: I think this robustness is nice to have but I wonder if in practice we can always assume (otherwise, enforce) that num_threads is a power of 2 and thus divides `n_l`.

ledwards2225 (approved):
Very nice. Most of my comments are just points for general discussion around multithreading

cpp/src/barretenberg/honk/pcs/gemini/gemini.cpp
ledwards2225: I suppose since `parallel_for` defines a loop based on the num threads it gets as input, `num_used_threads` will indeed be the max num threads used for multithreading, there is just no guarantee that it wont be fewer.

cpp/src/barretenberg/honk/pcs/gemini/gemini.cpp
Rumata888: It's a guess. I think it would be helpful to actually check what parallelize_for is useful for. I'll add an issue to concretely evaluate the size of computation for which it is efficient.

Rumata888: Issue: https://github.com/AztecProtocol/barretenberg/issues/553

cpp/src/barretenberg/honk/pcs/gemini/gemini.cpp
Rumata888: Yes, that's why I get the value of the number of available threads first and then compute num_used_threads based on that

cpp/src/barretenberg/honk/pcs/gemini/gemini.cpp
Rumata888: Lots of 6-core laptops out there)

PR #559, opened 2023-06-27
chore: Shared Permutation+Lookup relation arithmetic

zac-williamson:
# Description

For relations that use a grand product, the algebra that defines the relation is contained exclusively within the respective Relation class.

This aligns with how we treat non-grand-product relations and prevents the duplication of complex polynomial algebra (e.g. previously the permutation+grand product algebra was duplicated between the relation class and `prover_library.cpp`

# Checklist:


cpp/src/barretenberg/honk/proof_system/permutation_library.hpp
```diff
@@ -0,0 +1,165 @@
...
+    using FF = typename Flavor::FF;
+
+    constexpr size_t NUM_RELATIONS = std::tuple_size<GrandProductRelations>{};
+    barretenberg::constexpr_for<0, NUM_RELATIONS, 1>([&]<size_t i>() {
+        using PermutationRelation = typename std::tuple_element<i, GrandProductRelations>::type;
+
+        // Assign the grand product polynomial to the relevant std::span member of `full_polynomials` (and its shift)
+        // For example, for UltraPermutationRelation, this will be `full_polynomials.z_perm`
+        // For example, for LookupRelation, this will be `full_polynomials.z_lookup`
+        std::span<FF>& full_polynomial = PermutationRelation::get_grand_product_polynomial(full_polynomials);
```
ledwards2225: Took me a second to work out what was going on here. The fact that it needed to be done this way seems like an indication that something in our full_polynomials/proving_key model could be improved but nothing really to do with this PR

cpp/src/barretenberg/honk/sumcheck/relations/permutation_relation.hpp
```diff
@@ -40,32 +79,21 @@ template <typename FF> class PermutationRelationBase {
...
             auto lagrange_last = View(input.lagrange_last);
 
             // Contribution (1)
             std::get<0>(accumulator) +=
-                (((z_perm + lagrange_first) * (w_1 + id_1 * beta + gamma) * (w_2 + id_2 * beta + gamma) *
-                  (w_3 + id_3 * beta + gamma)) -
-                 ((z_perm_shift + lagrange_last * public_input_delta) * (w_1 + sigma_1 * beta + gamma) *
-                  (w_2 + sigma_2 * beta + gamma) * (w_3 + sigma_3 * beta + gamma))) *
+                (((z_perm + lagrange_first) *
+                  compute_permutation_numerator<AccumulatorTypes>(input, relation_parameters, 0)) -
```
ledwards2225: The index = 0 concerned me at first until I tracked down the different implementations of `get_view`. It's minor but would be nice to find a way to improve the interface here. Happy to leave as a possible TODO

ledwards2225 (approved):
Looks good to me. One minor change that I think is worth making: I'd like to remove the keyword "permutation" from objects that refer to arbitrary grand products more generally. For example, I would prefer the function `compute_permutation_grand_products` (which for Ultra computes both the perm and lookup grand products) be called simply `compute_grand_products` or perhaps `compute_grand_product_polynomials`. There are several similar instances.

PR #581, opened 2023-07-05
feat: Support public inputs in Ultra Honk

### LINKED ISSUE #579: Allow UltraHonk to handle public inputs
Honk requires all polynomials for which we need a shift to have constant coefficient equal to zero. For standard Honk, the only polynomial that is shifted is the grand product, for which we enforce z[0] = 0. For UltraHonk we shift many more polynomials, including the wires. Since PI are placed at the beginning of the coefficient arrays, the constant coefficients will in general not be zero. A hack solution is to simply add 0 as the first PI in all circuits. A better solution may be to simply place the PI at the second position instead of the first, effectively leaving a row of zeros in the execution trace. More generally, we can define a `public_inputs_offset` specifying the shift the placement of the PI. For goblin, this value will be equal to the number of ecc op "gates" since these will live at the start of the execution trace.

ledwards2225:
# Description

The primary goal of this PR is to support public inputs in Ultra Honk. This was previously not supported because adding a non-zero public input to an UH circuit would result in wires with a non-zero first entry, making them not shiftable. The solution is to add a "zero row" at the start of the execution trace for all UH circuits. The ordering of the execution trace is thus: zero row, public inputs, conventional gates.

A portion of the work in this PR is simply cleanup that was necessary to limit the number of places in the code where we need to specify the "zero row offset". For example, this offset has to be taken into account in the computation of the dyadic circuit size, which was previously computed independently in 3 separate locations within the Honk composers.

Note about implementation: In general there are many "zero rows" in every circuit, i.e. the padding between the "filled" rows of the execution trace and the dyadic (power of 2) circuit size. For this reason, 0 is added as a variable in the circuit builders by default. To add a zero row at the start of the execution trace, I simply add the first index of each wire to the copy cycle associated with the variable 0. Whether or not this zero row is utilized is specified in the Flavor, since, for example, it is not needed for plonk variants or even Standard Honk.

# Checklist:


cpp/src/barretenberg/honk/transcript/transcript.test.cpp
```diff
@@ -87,7 +89,7 @@ template <typename Flavor> class TranscriptTests : public testing::Test {
         round++;
         // TODO(Mara): Make testing more flavor agnostic so we can test this with all flavors
         if constexpr (IsGrumpkinFlavor<Flavor>) {
-            manifest_expected.add_entry(round, "IPA:poly_degree", circuit_size);
+            manifest_expected.add_entry(round, "IPA:poly_degree", size_uint64);
```
ledwards2225, line 92: This was a bug in the test that was revealed by the work in this PR but is otherwise unrelated.

cpp/src/barretenberg/plonk/composer/composer_lib.hpp
```diff
@@ -9,6 +9,57 @@ struct SelectorProperties {
     bool requires_lagrange_base_polynomial = false;
 };
 
+/**
```
ledwards2225, line 12: These two functions were formerly in the shared composer lib but they are used only for plonk so they've been moved to the plonk specific composer lib

cpp/src/barretenberg/proof_system/composer/composer_helper_lib.hpp
```diff
@@ -1,179 +0,0 @@
-#pragma once
```
ledwards2225, line 1: This file was an exact copy of composer_lib and was not used at all so I've deleted it entirely.

cpp/src/barretenberg/honk/composer/ultra_composer.cpp
```diff
@@ -251,33 +249,6 @@ std::shared_ptr<typename Flavor::ProvingKey> UltraComposer_<Flavor>::compute_pro
 
     // Polynomial memory is zeroed out when constructed with size hint, so we don't have to initialize trailing space
 
-    // // In the case of using UltraPlonkComposer for a circuit which does _not_ make use of any lookup tables, all
```
ledwards2225, line 254: This is the Plonk strategy for ensuring non-zero polys. We wont ever use this for Honk.

cpp/src/barretenberg/honk/composer/ultra_composer.cpp
```diff
@@ -189,38 +205,20 @@ std::shared_ptr<typename Flavor::ProvingKey> UltraComposer_<Flavor>::compute_pro
...
-    const size_t num_randomized_gates = NUM_RESERVED_GATES;
-    // Initialize proving_key
-    // TODO(#392)(Kesha): replace composer types.
-    proving_key = initialize_proving_key<Flavor>(
-        circuit_constructor, crs_factory_.get(), minimum_circuit_size, num_randomized_gates);
+    proving_key = std::make_shared<ProvingKey>(dyadic_circuit_size, num_public_inputs);
 
     construct_selector_polynomials<Flavor>(circuit_constructor, proving_key.get());
 
-    // TODO(#217)(luke): Naively enforcing non-zero selectors for Honk will result in some relations not being
```
ledwards2225, line 208: This is the Plonk strategy for ensuring non-zero polys. We wont ever use this for Honk.

cpp/src/barretenberg/honk/composer/ultra_composer.hpp
```diff
@@ -36,6 +38,11 @@ template <UltraFlavor Flavor> class UltraComposer_ {
     std::vector<uint32_t> recursive_proof_public_input_indices;
     bool contains_recursive_proof = false;
     bool computed_witness = false;
+    size_t total_num_gates = 0; // num_gates + num_pub_inputs + tables + zero_row_offset (used to compute dyadic size)
+    size_t dyadic_circuit_size = 0; // final dyadic circuit size
```
Rumata888: I'd recommend adding (next power of two) in the comments. As a variable name dyadic is ok, but for me it wasn't clear from the start if this is supposed to be the logarithm or the power of 2

cpp/src/barretenberg/honk/flavor/standard.hpp
```diff
@@ -66,6 +66,9 @@ class Standard {
     using RelationUnivariates = decltype(create_relation_univariates_container<FF, Relations>());
     using RelationValues = decltype(create_relation_values_container<FF, Relations>());
 
+    // Whether or not the first row of the execution trace is reserved for 0s to enable shifts
+    static constexpr bool has_zero_row = false;
```
Rumata888, line 70: Don't we need it for all honk flavors?

cpp/src/barretenberg/honk/flavor/standard_grumpkin.hpp
```diff
@@ -57,8 +57,8 @@ class StandardGrumpkin {
     using RelationUnivariates = decltype(create_relation_univariates_container<FF, Relations>());
     using RelationValues = decltype(create_relation_values_container<FF, Relations>());
 
-    // define utilities to extend univarates from RELATION_LENGTH to MAX_RELATION_LENGTH for each Relation
-    // using BarycentricUtils = decltype(create_barycentric_utils<FF, Relations, MAX_RELATION_LENGTH>());
+    // Whether or not the first row of the execution trace is reserved for 0s to enable shifts
+    static constexpr bool has_zero_row = false;
```
Rumata888, line 61: Don't we need it for all honk flavors?

Rumata888 (approved):

cpp/src/barretenberg/honk/flavor/standard_grumpkin.hpp
ledwards2225: No need for the zero row for Standard flavors since in this case the only shifted poly is z_perm which is constructed explicitly with a constant coeff equal to 0. The problem arises for UltraHonk since we need to shift the wire polys themselves

cpp/src/barretenberg/honk/composer/ultra_composer.hpp
ledwards2225: good call - done

PR #648, opened 2023-08-01
feat: Minimal circleci

ludamad:
Trying to alleviate spot instance issues

Later we can remove the commented out code, but thought it was good to have for reference for now in case people end up redoing any of the parts


ludamad:
Merging through aztec-packages

PR #702, opened 2023-08-22
feat: add bbjs testing workflow

jonybur:
This PR is used to test out the workflow that is implemented here: https://github.com/AztecProtocol/aztec-packages/pull/1733

Please close once merged.


kevaundray:
Should we close this in favour of https://github.com/AztecProtocol/aztec-packages/pull/1855 ?

PR #1353, opened 2025-04-17
Update Doxyfile

shystrui1199:
fix to a correct link

you are welcome


Maddiaa0:
thanks, but this has no outcome, there is a redirect on the existing link

